PR #16, opened 2021-01-16
Initial support for LoadBalancer

### LINKED ISSUE #15: Expose the StatefulSet for use external to Kubernetes
A client needs to access each specific broker, so a LoadBalancer is needed for each broker.

The broker needs to advertise the address of the LoadBalancer, which is a little more tricky. I think it would be worth investigating https://github.com/kubernetes-sigs/external-dns - this should allow a predictable DNS entry for each LoadBalancer  associated with each Redpanda broker, and generate the `advertised_kafka_api` with helm. Some docs here: https://github.com/kubernetes-sigs/external-dns/blob/master/docs/tutorials/gke.md

Something like:

service.loadbalancer.yaml
```yaml
{{- $lb := .Values.loadBalancer }}
{{- if $lb.enabled }}
{{- range untilStep 0 (.Values.statefulset.replicas|int) 1 }}
apiVersion: v1
kind: Service
metadata:
  name: redpanda-{{ . }}
annotations:
  external-dns.alpha.kubernetes.io/hostname: redpanda-{{ . }}.{{.Values.loadBalancer.parentZone}}
spec:
  externalTrafficPolicy: Local
  ports:
    - name: kafka
      protocol: TCP
      port: 9092
      targetPort: 9092
  selector:
    statefulset.kubernetes.io/pod-name: redpanda-{{ . }}
  type: LoadBalancer
---
{{- end }}
{{- end }}
```

statefulset.yaml
```yaml
rpk --config $CONFIG config set redpanda.advertised_kafka_api.address redpanda-$NODE_ID.{{.Values.loadBalancer.parentZone}}
```

Values.yaml
```yaml
loadBalancer:
  enabled: true
  parentZone: my-project.gcp.domain.com.
```

Currently Redpanda can currently only [listen on one address](https://github.com/vectorizedio/redpanda/issues/186), which means traffic inside the K8s cluster would have to go via the LoadBalancer.

BenPope:
## Initial support for exposing statefulset externally via a LoadBalancer

Fixes #15 

E.g.:
`helm template--create-namespace --namespace redpanda redpanda . --set loadBalancer.enabled=true --set loadBalancer.parentZone=my-project.namespace.example.com.`

Includes integration with: https://github.com/kubernetes-sigs/external-dns

### DIFF
--- a/redpanda/templates/service.loadbalancer.yaml
+++ b/redpanda/templates/service.loadbalancer.yaml
@@ -1,0 +1,54 @@
+{{/*
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
+(the "License"); you may not use this file except in compliance with
+the License.  You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/}}
+{{- if .Values.loadBalancer.enabled }}
+{{- range untilStep 0 (.Values.statefulset.replicas|int) 1 }}
+# This service is to expose a pod in the statefulset with a loadbalancer.
+# Includes integration with https://github.com/kubernetes-sigs/external-dns
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ template "redpanda.fullname" $ }}-{{ . }}
+  namespace: {{ $.Release.Namespace | quote }}
+  labels:
+    helm.sh/chart: {{ template "redpanda.chart" $ }}
+    app.kubernetes.io/name: {{ template "redpanda.name" $ }}
+    app.kubernetes.io/instance: {{ $.Release.Name | quote }}
+    app.kubernetes.io/managed-by: {{ $.Release.Service | quote }}
+    app.kubernetes.io/component: {{ template "redpanda.name" $ }}
+  {{- with $.Values.labels }}
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  annotations:
+    external-dns.alpha.kubernetes.io/hostname: {{ template "redpanda.fullname" $ }}-{{ . }}.{{$.Values.loadBalancer.parentZone}}
+    {{- range $key, $value := $.Values.loadBalancer.annotations}}
+    {{$key}}: {{$value}}
+    {{- end }}
+spec:
+  externalTrafficPolicy: Local
+  ports:
+    - name: kafka
+      protocol: TCP
+      port: {{ $.Values.loadBalancer.port }}
+      targetPort: {{ template "redpanda.kafka.external.listen.port" $ }}
+  selector:
+    app.kubernetes.io/name: {{ template "redpanda.name" $ }}
+    app.kubernetes.io/instance: {{ $.Release.Name | quote }}
+    statefulset.kubernetes.io/pod-name: {{ template "redpanda.fullname" $ }}-{{ . }}
+  type: LoadBalancer
+---
+{{- end }}
+{{- end }}
--- a/redpanda/templates/statefulset.yaml
+++ b/redpanda/templates/statefulset.yaml
@@ -104,6 +104,10 @@ spec:
               --kafka-addr=internal://{{ template "redpanda.kafka.internal.listen.address" . }}:{{ template "redpanda.kafka.internal.listen.port" . }}
               --advertise-rpc-addr={{ template "redpanda.rpc.advertise.address" . }}:{{ template "redpanda.rpc.advertise.port" . }}
               --rpc-addr={{ template "redpanda.rpc.listen.address" . }}:{{ template "redpanda.rpc.listen.port" . }}
+              {{- if .Values.loadBalancer.enabled }}
+              --advertise-kafka-addr=external://{{ template "redpanda.kafka.external.advertise.address" . }}:{{ template "redpanda.kafka.external.advertise.port" . }}
+              --kafka-addr=external://{{ template "redpanda.kafka.external.listen.address" . }}:{{ template "redpanda.kafka.external.listen.port" . }}
+              {{- end }}
               --
               --default-log-level={{ .Values.config.seastar.default_log_level }}
           ports:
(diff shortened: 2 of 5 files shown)


redpanda/values.yaml
```diff
@@ -209,3 +209,10 @@ storage:
     labels: {}
     # Additional annotations to apply to the created PersistentVolumeClaims.
     annotations: {}
+
```
rkruze, line 212: Can you please add comments?

redpanda/templates/statefulset.yaml
```diff
@@ -69,7 +69,11 @@ spec:
               fi;
               rpk --config $CONFIG config set redpanda.advertised_rpc_api.address $SERVICE_NAME;
               rpk --config $CONFIG config set redpanda.advertised_rpc_api.port {{ .Values.config.redpanda.rpc_server.port }};
+              {{- if .Values.loadBalancer.enabled }}
+              rpk --config $CONFIG config set redpanda.advertised_kafka_api.address {{ template "redpanda.name" . }}-$NODE_ID.{{.Values.loadBalancer.parentZone}}
```
rkruze, line 73: I'll be adding a fix for this shortly, but I believe that this should be `redpanda.fullname` in case "redpanda" is not used for the `Release.Name`.

rkruze (commented):
The NOTES.txt should also be updated to print out the address to be used for external clients.

BenPope:
This now requires and supports: https://github.com/vectorizedio/redpanda/pull/451

redpanda/templates/statefulset.yaml
```diff
@@ -70,11 +70,12 @@ spec:
               rpk --config $CONFIG config set redpanda.advertised_rpc_api.address $SERVICE_NAME;
               rpk --config $CONFIG config set redpanda.advertised_rpc_api.port {{ .Values.config.redpanda.rpc_server.port }};
               {{- if .Values.loadBalancer.enabled }}
-              rpk --config $CONFIG config set redpanda.advertised_kafka_api.address {{ template "redpanda.name" . }}-$NODE_ID.{{.Values.loadBalancer.parentZone}}
+              ADVERTISED_KAFKA_API='[{"address": "'"$SERVICE_NAME"'", "port": {{.Values.config.redpanda.kafka_api.port}}, "name":"internal"},{"address": "{{ template "redpanda.fullname" . }}-'"$NODE_ID"'.{{.Values.loadBalancer.parentZone}}", "port": {{.Values.loadBalancer.port}}, "name":"external"}]';
```
BenPope, line 73: For some reason when running in the cluster, rpk only renders:
```yaml
  advertised_kafka_api:
  - address: redpanda-0.redpanda.redpanda.svc.cluster.local
```
But if I run the command locally, I get:
```yaml
  advertised_kafka_api:
  - address: redpanda-0.redpanda.redpanda.svc.cluster.local
    name: internal
    port: 9092
  - address: redpanda-0.my-project.gcp.domain.com.
    name: external
    port: 9092
```

senior7515: @dotnwat  ^^

BenPope: Ignore this, it's now fine. rpk start doesn't understand it, though.

BenPope:
Useful for Mac: https://www.thehumblelab.com/kind-and-metallb-on-mac/

BenPope:
https://github.com/vectorizedio/redpanda/pull/556 released - this is now ready for review (depends on #25 )

RafalKorepta (approved):
LGTM

redpanda/templates/service.loadbalancer.yaml
```diff
@@ -0,0 +1,54 @@
...
+    helm.sh/chart: {{ template "redpanda.chart" $ }}
+    app.kubernetes.io/name: {{ template "redpanda.name" $ }}
+    app.kubernetes.io/instance: {{ $.Release.Name | quote }}
+    app.kubernetes.io/managed-by: {{ $.Release.Service | quote }}
+    app.kubernetes.io/component: {{ template "redpanda.name" $ }}
+  {{- with $.Values.labels }}
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  annotations:
+    external-dns.alpha.kubernetes.io/hostname: {{ template "redpanda.fullname" $ }}-{{ . }}.{{$.Values.loadBalancer.parentZone}}
```
RafalKorepta, line 36: There could be problem when I create 2 releases with the same name and 2 different namespaces there will be clash of the hostnames.

RafalKorepta (changes requested):
Sorry one comment

redpanda/templates/service.loadbalancer.yaml
BenPope: I think that's expected, though. The answer is don't do that.

```
helm install  --set loadBalancer.parentZone=unique.example.com
```

BenPope: Maybe the default should include the namespace to make that clearer, and then a comment saying reaffirming the unique requirement?

RafalKorepta: Agree. Adding namespace it will help to create unique hostnames in on cluster.

redpanda/templates/statefulset.yaml
```diff
@@ -81,21 +84,42 @@ spec:
...
+              --smp={{ .Values.statefulset.resources.limits.cpu }}
+              --memory={{ template "redpanda.parseMemory" . }}
+              --reserve-memory=0M
+              --advertise-kafka-addr=internal://{{ template "redpanda.kafka.internal.advertise.address" . }}:{{ template "redpanda.kafka.internal.advertise.port" . }}
+              --kafka-addr=internal://{{ template "redpanda.kafka.internal.listen.address" . }}:{{ template "redpanda.kafka.internal.listen.port" . }}
+              --advertise-rpc-addr={{ template "redpanda.rpc.advertise.address" . }}:{{ template "redpanda.rpc.advertise.port" . }}
+              --rpc-addr={{ template "redpanda.rpc.listen.address" . }}:{{ template "redpanda.rpc.listen.port" . }}
+              {{- if .Values.loadBalancer.enabled }}
+              --advertise-kafka-addr=external://{{ template "redpanda.kafka.external.advertise.address" . }}:{{ template "redpanda.kafka.external.advertise.port" . }}
+              --kafka-addr=external://{{ template "redpanda.kafka.external.listen.address" . }}:{{ template "redpanda.kafka.external.listen.port" . }}
```
RafalKorepta, line 109: NIT: I would use different port to differentiate internal/external port. 

Current setup routes LB -> port 9092 -> reroute 9093 -> hit the container.

RafalKorepta (changes requested):
Only the one issue to address. I'm ok with the comment in the docs that release name + parent zone can not duplicated across one dns zone.

PR #17, opened 2021-01-16
Initial github actions testing

BenPope:
Introduce some testing for the helm chart.

* Cleanup whitespace
* Fix maintainer - should be a github user, not an arbitrary string
* Introduce testing
* Add test states to readme

Signed-off-by: BenPope <benpope81@gmail.com>


BenPope:
Replaced by #19

PR #18, opened 2021-01-19
Support release names that are not "redpanda"

rkruze:
If you specified a release name like "my-redpanda" the cluster would not
form as the dns name of the instances was not "redpanda-0" any longer
but was "my-redpanda-0".

PR #19, opened 2021-01-20
Initial GitHub actions testing

RafalKorepta:
Introduce some testing for the helm chart.

- Cleanup whitespace
- Fix maintainer - should be a github user, not an arbitrary string
- Introduce testing
- Add test states to readme

It's based on #17 and #18

### DIFF
--- a/.github/workflows/lint-test.yml
+++ b/.github/workflows/lint-test.yml
@@ -1,0 +1,53 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Lint and Test Charts
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - *
+jobs:
+  lint-test:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - name: Set up Helm
+        uses: azure/setup-helm@v1
+        with:
+          version: v3.4.1
+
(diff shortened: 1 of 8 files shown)


RafalKorepta:
The pipeline passed here https://github.com/RafalKorepta/helm-charts/runs/1736284671?check_suite_focus=true

rkruze (changes requested):
Should we move cluster.yaml and ct.yaml to under the .github/workflow directory?

BenPope:
> Should we move cluster.yaml and ct.yaml to under the .github/workflow directory?

`redpanda/ci` seems good.

cluster.yaml
```diff
@@ -0,0 +1,11 @@
+kind: Cluster
```
BenPope, line 1: Is this used in github actions?

I guess it's based on what I did for local cluster testing?

cluster.yaml
```diff
@@ -0,0 +1,11 @@
...
+apiVersion: kind.x-k8s.io/v1alpha4
+containerdConfigPatches:
+  - |-
+    [plugins."io.containerd.grpc.v1.cri".registry.mirrors."localhost:5000"]
```
senior7515, line 5: wow this string is unreal. never seen a `"foo".bar.baz."postfix"` in my life. crazy. is this a bug?

BenPope: https://kind.sigs.k8s.io/docs/user/local-registry/

cluster.yaml
RafalKorepta: Yes, I will remove this in the next push.

cluster.yaml
RafalKorepta: This file will be removed in the next push.

RafalKorepta:
I address the comments. `ct.yaml` is moved to `redpanda/ci`. I reorganize the commits. The unused file where removed.

.github/workflows/lint-test.yml
```diff
@@ -0,0 +1,42 @@
...
+# Use of this software is governed by the Business Source License
+# included in the file licenses/BSL.md
+#
+# As of the Change Date specified in that file, in accordance with
+# the Business Source License, use of this software will be governed
+# by the Apache License, Version 2.0
+
+name: Lint and Test Charts
+
+on: [push, pull_request]
```
0xdiba, line 12: Let's limit `push` runs to `main` branch only.

RafalKorepta: Ok. Done

.github/workflows/lint-test.yml
```diff
@@ -0,0 +1,47 @@
+# Copyright 2020 Vectorized, Inc.
+#
+# Use of this software is governed by the Business Source License
```
rkruze, line 3: This should be Apache 2 license

RafalKorepta: No problem. Changed.

BenPope (approved):

rkruze (approved):

PR #20, opened 2021-01-20
On publishing a new release, update the chart repo

rkruze:
This will use the tag name to update the chart version when publishing
to the gh-pages branch.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,35 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -0,0 +1,21 @@
...
+      run: echo ::set-output name=VERSION::${GITHUB_REF/refs\/tags\//}
+    
+    - uses: actions/checkout@v2
+    - name: Publish Helm charts
+      uses: stefanprodan/helm-gh-pages@master
+      with:
+        charts_dir: '.'
+        charts_url: https://charts.vectorized.io/
+        token: ${{ secrets.GITHUB_TOKEN }}
+        chart_version: ${{ steps.get_version.outputs.VERSION }}
```
0xdiba, line 21: You should be able to get this with `github.event.release.tag_name`, which also makes the previous step obsolete.

0xdiba (commented):
Looks good.
Only comment on using the github event API to get rid of extrapolating the tag.

RafalKorepta (approved):
LGTM. 

Should we have automated release github action?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,21 @@
+name: release
```
RafalKorepta, line 1: NIT: Add license header?

# Copyright 2020 Vectorized, Inc.
#
# Use of this software is governed by the Business Source License
# included in the file licenses/BSL.md
#
# As of the Change Date specified in that file, in accordance with
# the Business Source License, use of this software will be governed
# by the Apache License, Version 2.0

rkruze: Thanks, since this is not redpanda itself it will just be the apache license.

.github/workflows/release.yaml
rkruze: Thanks! Fixed.

PR #22, opened 2021-01-27
Initial support for kind

BenPope:
Currently based on #16 as that's a dependency.

The script currently only works with Linux and only tested with Ubuntu.

Useful for Mac: https://www.thehumblelab.com/kind-and-metallb-on-mac/

### DIFF
--- a/kind/README.md
+++ b/kind/README.md
@@ -1,0 +1,46 @@
+# Redpanda on KinD
+
+## TLDR
+
+```sh
+./make.sh
+```
+
+## Details
+
+### Local Registry
+
+There is a [local registry](https://kind.sigs.k8s.io/docs/user/local-registry/).
+
+The registry can be used like this.
+
+1. First we'll pull an image `docker pull vectorized/redpanda:latest`
+2. Then we'll tag the image to use the local registry `docker tag vectorized/redpanda:latest localhost:5000/redpanda:latest`
+3. Then we'll push it to the registry `docker push localhost:5000/redpanda:latest`
+4. And now we can use the image `kubectl create deployment hello-server --image=localhost:5000/redpanda:latest`
+
+If you build your own image and tag it like `localhost:5000/redpanda:latest` and then use it in kubernetes as `localhost:5000/redpanda:latest`.
+
+### LoadBalancer
+
+[metallb](https://metallb.universe.tf/) is used.
+
+This works on Ubuntu - YMMV.
+
+Useful info for Mac: https://www.thehumblelab.com/kind-and-metallb-on-mac/
+
+I don't know of a non-intrusive way to get DNS lookups from outside the cluster.
+
+For now modifying `/etc/hosts` is suggested, installation will describe how.
+
+### Redpanda
+
+3-node cluster.
+
+Each node has a PVC.
+
+Testing a rolling restart:
+
+```sh
+kubectl -n redpanda rollout restart statefulset redpanda
+```
--- a/kind/make.sh
+++ b/kind/make.sh
@@ -1,0 +1,69 @@
+#!/bin/bash
+set -euo pipefail
+
+# Create registry container unless it already exists
+reg_name='kind-registry'
+reg_port='5000'
+running="$(docker inspect -f '{{.State.Running}}' "${reg_name}" 2>/dev/null || true)"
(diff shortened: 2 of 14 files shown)


kind/make.sh
```diff
@@ -0,0 +1,67 @@
...
+  --set configInline.address-pools[0].addresses[0]="$ADDRESSES"
+
+#Install redpanda
+helm install \
+ --create-namespace \
+ --namespace redpanda \
+ redpanda ../redpanda \
+ --set loadBalancer.enabled=true \
+ --set image.repository=localhost:5000/redpanda \
+ --set image.tag=b2783df
```
BenPope, line 51: oops

kind/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+Testing a rolling restart:
+
+```sh
+kubectl -n redpanda rollout restart statefulset redpanda
+```
+
+### Accessing Kubernetes
+
+I'm not a fan of [kubernetes-dashboard](https://github.com/kubernetes/dashboard#kubernetes-dashboard) and it's a pain to set up.
```
RafalKorepta, line 48: Maybe project [octant](https://octant.dev/) can be listed hear too.

kind/make.sh
```diff
@@ -0,0 +1,66 @@
...
+echo ""
+kubectl -n redpanda rollout status -w statefulset/redpanda
+echo "
+Remove cluster with:
+
+  $ kind delete cluster --name redpanda
+
+Add these to /etc/hosts:
+"
+kubectl -n redpanda get svc | grep redpanda- | awk '{printf "%-15s %s %s.redpanda.redpanda.svc.cluster.local\n", $4, $1, $1}' || true
```
RafalKorepta, line 62:
```suggestion
kubectl -n redpanda get svc redpanda-bootstrap | awk '{printf "%-15s %s %s.redpanda.redpanda.svc.cluster.local\n", $4, $1, $1}' || true
```

kind/make.sh
```diff
@@ -0,0 +1,66 @@
...
+  --namespace metallb-system \
+  metallb bitnami/metallb \
+  -f metallb-values.yaml \
+  --set configInline.address-pools[0].addresses[0]="$ADDRESSES"
+
+#Install redpanda
+helm install \
+ --create-namespace \
+ --namespace redpanda \
+ redpanda ../redpanda \
```
RafalKorepta, line 48: What do you think about extract `redpanda` to local variable?
All places that uses the release name could use mentioned local variable.

kind/make.sh
```diff
@@ -0,0 +1,66 @@
...
+docker pull vectorized/redpanda:latest
+docker tag vectorized/redpanda:latest localhost:5000/redpanda:latest
+docker push localhost:5000/redpanda:latest
+
+NODES=$(kubectl get nodes -o json | jq -r '.items[].status.addresses | select(.[].address | startswith("redpanda-worker")) | .[] | select(.type == "InternalIP").address')
+SUBNET=$(echo "$NODES" | head -n1 | cut -d. -f 1,2).255
+ADDRESSES="$SUBNET.1-$SUBNET.254"
+
+# Install metallb
+helm repo add bitnami https://charts.bitnami.com/bitnami
```
RafalKorepta, line 36: What do you think to create empty helm chart that has `requirments.yaml` pointing to `metallb` and `redpanda`?

BenPope: Sounds neat.

kind/make.sh
BenPope: This script certainly isn't production ready!

kind/make.sh
BenPope: Ahh. This kind of relies on #16 - so it's supposed to grab the LoadBalance services. But yeah, the grep is wrong these days. I can harden this.

kind/README.md
BenPope: These types of opinions probably should be in this repo. I'll remove it.

kind/make.sh
RafalKorepta: I see. Thanks for explanation

BenPope: Rebased on #16 and hardened it with some jq magic.

PR #24, opened 2021-02-03
Initial Pandaproxy Deployment

BenPope:
Note: This only supports replicas: 1

Based on #22 for dependencies

TODO:
* Switch from `LoadBalancer` to `Ingress`
* Expose route based on `advertised_address`
* Maybe it should be a `StatefulSet`?

### DIFF
--- a/kind/README.md
+++ b/kind/README.md
@@ -1,0 +1,46 @@
+# Redpanda on KinD
+
+## TLDR
+
+```sh
+./make.sh
+```
+
+## Details
+
+### Local Registry
+
+There is a [local registry](https://kind.sigs.k8s.io/docs/user/local-registry/).
+
+The registry can be used like this.
+
+1. First we'll pull an image `docker pull vectorized/redpanda:latest`
+2. Then we'll tag the image to use the local registry `docker tag vectorized/redpanda:latest localhost:5000/redpanda:latest`
+3. Then we'll push it to the registry `docker push localhost:5000/redpanda:latest`
+4. And now we can use the image `kubectl create deployment hello-server --image=localhost:5000/redpanda:latest`
+
+If you build your own image and tag it like `localhost:5000/redpanda:latest` and then use it in kubernetes as `localhost:5000/redpanda:latest`.
+
+### LoadBalancer
+
+[metallb](https://metallb.universe.tf/) is used.
+
+This works on Ubuntu - YMMV.
+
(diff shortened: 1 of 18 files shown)


senior7515:
i don't think this makes sense since we'll add the proxy to the redpanda binary img.

dotnwat:
> i don't think this makes sense since we'll add the proxy to the redpanda binary img.

can you say a little more? it seems like deployment scenarios are distinct from packaging, so maybe you're implying something about a unified helm chart or unified operator for both modes of operation?

senior7515:
> > i don't think this makes sense since we'll add the proxy to the redpanda binary img.
> 
> can you say a little more? it seems like deployment scenarios are distinct from packaging, so maybe you're implying something about a unified helm chart or unified operator for both modes of operation?

from the demo today, basically our kubernetes operator is a stateful set. so ben wanted to have a stateful set for the proxy. once the redpanda binary has an option `config.enable_http_proxy` then it gets the same deployment model in k8s without additional sidecars.

BenPope:
> from the demo today, basically our kubernetes operator is a stateful set. so ben wanted to have a stateful set for the proxy. 

I think the proxy should be a `StatefulSet` because the instances have "identity", that is, it matters which one a client connects to.

I think there are trade-offs in running it standalone or embedded in Redpanda.

I wrote the code for the demo, for local testing, and possibly for other integration test scenarios, such as running it against a Kafka cluster. I left this PR as draft because it's not finished work, it's just what I had and wanted to publish it.

dotnwat:
> > > i don't think this makes sense since we'll add the proxy to the redpanda binary img.
> > 
> > 
> > can you say a little more? it seems like deployment scenarios are distinct from packaging, so maybe you're implying something about a unified helm chart or unified operator for both modes of operation?
> 
> from the demo today, basically our kubernetes operator is a stateful set. so ben wanted to have a stateful set for the proxy. once the redpanda binary has an option `config.enable_http_proxy` then it gets the same deployment model in k8s without additional sidecars.

This what I thought you had in mind. I think it makes sense; I'll be curious to see if there are conflicts arise: things that an operator may do for a redpanda statefulset may not be relevant for a proxy layer, especially when it comes to upgrade procedure and day 2 ops.

BenPope:
Not required due to https://github.com/vectorizedio/redpanda/pull/682

PR #25, opened 2021-02-05
Update helm chart

BenPope:
Depends on https://github.com/vectorizedio/redpanda/pull/556

* Template listener addresses and ports
* Remove `node-id` from `seeds`
* Use named address for `kafka-api`
* Use `rpk redpanda start` instead of `rpk start`
* Move configuration from `rpk config set` to `rpk redpanda start`

### DIFF
--- a/redpanda/templates/NOTES.txt
+++ b/redpanda/templates/NOTES.txt
@@ -25,10 +25,10 @@ Try some sample commands, like creating a topic called topic1:
 
   kubectl -n {{ .Release.Namespace }} run -ti --rm --restart=Never \
     --image {{ .Values.image.repository }}:{{ .Values.image.tag }} \
-    rpk -- --brokers={{ include "redpanda.fullname" . }}-bootstrap:{{ .Values.config.redpanda.kafka_api.port }} api topic create topic1
+    rpk -- --brokers={{ include "redpanda.fullname" . }}-bootstrap:{{ template "redpanda.kafka.internal.advertise.port" $ }} api topic create topic1
 
 To get the api status:
  
   kubectl -n {{ .Release.Namespace }} run -ti --rm --restart=Never \
     --image {{ .Values.image.repository }}:{{ .Values.image.tag }} \
-    rpk -- --brokers={{ include "redpanda.fullname" . }}-bootstrap:{{ .Values.config.redpanda.kafka_api.port }} api status
+    rpk -- --brokers={{ include "redpanda.fullname" . }}-bootstrap:{{ template "redpanda.kafka.internal.advertise.port" $ }} api status
--- a/redpanda/templates/statefulset.yaml
+++ b/redpanda/templates/statefulset.yaml
@@ -53,24 +53,27 @@ spec:
     spec:
       securityContext:
         {{- toYaml .Values.podSecurityContext | nindent 8 }}
+      # TODO:
+      # * Figure out what to do about node_id / seeds here - the operator will fix this separately
+      # * Once that's done, this initContainer can be removed
       initContainers:
         - name: {{ template "redpanda.name" . }}-configurator
           image: {{ .Values.image.repository }}:{{ .Values.image.tag }}
           command: ["/bin/sh", "-c"]
+          env:
+            - name: SERVICE_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.name
           args:
             - >
               CONFIG=/etc/redpanda/redpanda.yaml;
-              NODE_ID=${HOSTNAME##*-};
-              SERVICE_NAME={{ template "redpanda.fullname" . }}-$NODE_ID.{{ template "redpanda.fullname" . }}.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }};
-              cp /tmp/base-config/redpanda.yaml $CONFIG;
-              rpk --config $CONFIG config set redpanda.node_id $NODE_ID;
+              NODE_ID=${SERVICE_NAME##*-};
+              cp /tmp/base-config/redpanda.yaml "$CONFIG";
+              rpk --config "$CONFIG" config set redpanda.node_id $NODE_ID;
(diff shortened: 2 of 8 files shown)


redpanda/templates/NOTES.txt
```diff
@@ -25,10 +25,10 @@ Try some sample commands, like creating a topic called topic1:
 
   kubectl -n {{ .Release.Namespace }} run -ti --rm --restart=Never \
     --image {{ .Values.image.repository }}:{{ .Values.image.tag }} \
-    rpk -- --brokers={{ include "redpanda.fullname" . }}-bootstrap:{{ .Values.config.redpanda.kafka_api.port }} api topic create topic1
+    rpk -- --brokers={{ include "redpanda.fullname" . }}-bootstrap:{{ template "redpanda.kafka.internal.advertise.port" $ }} api topic create topic1
```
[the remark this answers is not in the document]
BenPope: This is slightly confusing - the pod will be named `rpk` and the arguments go after `--`, so the final command is the entrypoint + args: `["rpk"]` + `["--brokers=<generated>", "api", "topic", "create", "topic1"]`

redpanda/values.yaml
```diff
@@ -50,8 +50,10 @@ config:
       # Metrics are served off of this port at /metrics.
       port: 9644
     kafka_api:
```
RafalKorepta, line 52: Suggestion: Maybe comment for the user that only first port number will be taken into forming kubernetes service. WDYT?

RafalKorepta (approved):
LGTM. I'm not sure if the documentation should live in the `values.yaml` or in different place.

redpanda/values.yaml
BenPope: Maybe I should search for internal, and document that as special, but first entry is just as clear and easier to implement!

BenPope: In hindsight, maybe it would be even better to take the list here, and generate listeners for them all. (For example, #16) - but it's tricky to tie everything together such as what service type, without making this incompatible with the actual config file format. I think I'll just do your suggestion, and then add extensible listeners support at a later date, I feel it might be a bit YAGNI for now.

redpanda/templates/statefulset.yaml
```diff
@@ -81,21 +84,34 @@ spec:
...
-              --check=false
-              --smp {{ .Values.statefulset.resources.limits.cpu }}
-              --memory {{ template "redpanda.parseMemory" . }}
+              redpanda
               start
+              --smp={{ .Values.statefulset.resources.limits.cpu }}
+              --memory={{ template "redpanda.parseMemory" . }}
+              --reserve-memory=0M
+              --advertise-kafka-addr=internal://{{ template "redpanda.kafka.internal.advertise.address" . }}:{{ template "redpanda.kafka.internal.advertise.port" . }}
+              --kafka-addr=internal://{{ template "redpanda.kafka.internal.listen.address" . }}:{{ template "redpanda.kafka.internal.listen.port" . }}
```
dimitriscruz, line 104: What is `internal`?

BenPope: It's an arbitrary name that refers to the listener for access internal to K8s, #16 adds `external` for the `LoadBalancer`.

redpanda/values.yaml
BenPope: For posterity:
```diff
diff --git a/redpanda/templates/statefulset.yaml b/redpanda/templates/statefulset.yaml
index d54a203..5f8d6cd 100644
--- a/redpanda/templates/statefulset.yaml
+++ b/redpanda/templates/statefulset.yaml
@@ -101,7 +101,9 @@ spec:
               --memory={{ template "redpanda.parseMemory" . }}
               --reserve-memory=0M
               --advertise-kafka-addr=internal://{{ template "redpanda.kafka.internal.advertise.address" . }}:{{ template "redpanda.kafka.internal.advertise.port" . }}
-              --kafka-addr=internal://{{ template "redpanda.kafka.internal.listen.address" . }}:{{ template "redpanda.kafka.internal.listen.port" . }}
+            {{- range .Values.config.redpanda.kafka_api }}
+              --kafka-addr={{ .name }}://{{ .address }}:{{ .port }}
+            {{- end }}
               --advertise-rpc-addr={{ template "redpanda.rpc.advertise.address" . }}:{{ template "redpanda.rpc.advertise.port" . }}
               --rpc-addr={{ template "redpanda.rpc.listen.address" . }}:{{ template "redpanda.rpc.listen.port" . }}
               --
@@ -109,8 +111,10 @@ spec:
           ports:
             - containerPort: {{ .Values.config.redpanda.admin.port }}
               name: admin
-            - containerPort: {{ template "redpanda.kafka.internal.listen.port" . }}
-              name: kafka
+            {{- range .Values.config.redpanda.kafka_api }}
+            - containerPort: {{ .port }}
+              name: {{ .name }}
+            {{- end }}
             - containerPort: {{ template "redpanda.rpc.listen.port" $ }}
               name: rpc
           volumeMounts:
```
This gets difficult, since advertised has to be `addressed_kafka_api` for any of this to work, and we don't have an address to use at this point in the chart, and I don't really want to arbitrarily override the address.

redpanda/templates/statefulset.yaml
dimitriscruz: It corresponds to the name in rpk's `NamedSocketAddress`?

BenPope: This is the format for passing namedSocketAddress on the command line to `rpk redpanda start`, and I beleive also the format Kafka uses.

dimitriscruz: Thanks, not a Helm expert here, but changes look good. We just need to wait for the updated redpanda image.

BenPope:
https://github.com/vectorizedio/redpanda/pull/556 released - this is now ready for review.

dimitriscruz (approved):

PR #28, opened 2021-02-24
Add overprovisioned flag to arguments

RafalKorepta:
In the containerised environment redpanda process shares
hardware with other processes so it can not appropriate
CPU and memory resource to itself.


RafalKorepta:
The `--overprovisioned` flag should be used only when there is not natural number of lest than 1 of resources request cpu in kubernetes manifests. There shouldn't be a need for introducing this flag without properly calculating the values. 

Reference https://github.com/vectorizedio/helm-charts/blob/main/redpanda/values.yaml#L115-L133

BenPope:
See: https://github.com/vectorizedio/redpanda/pull/687#issuecomment-785066105

PR #43, opened 2022-04-28
Override default cert duration

jrkinley:


redpanda/values_add_tls.yaml
```diff
@@ -76,6 +76,7 @@ config:
 
 certIssuers:
   - name: kafka
+    duration: 8760h
```
rkruze, line 79: Do we need to have this in the config? Could we just use the default?

rkruze (changes requested):

redpanda/values_add_tls.yaml
jrkinley: We should use the default. I didn't remove the override after testing.

PR #47, opened 2022-05-09
Triggering cluster reload on config change

burandobata:
Making sure a config change will be applied on the cluster.

### DIFF
--- a/redpanda/templates/statefulset.yaml
+++ b/redpanda/templates/statefulset.yaml
@@ -46,9 +46,11 @@ spec:
       {{- with .Values.labels }}
         {{- toYaml . | nindent 8 }}
       {{- end }}
-    {{- with .Values.statefulset.annotations }}
-      annotations: {{- toYaml . | nindent 8 }}
-    {{- end }}
+      annotations:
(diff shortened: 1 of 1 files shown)


vuldin:
Thanks for this and your other PRs @burandobata ! In the meantime I see the CLAassistant says you haven't signed the contributor agreement, could you do that? I'll take a look at the PRs and provide any feedback soon.

vuldin:
Hi @burandobata . I was just added to the repo and am now able to review PRs. I'm trying to ensure this PR works as expected and unable to get the config updated. It's possible I'm using commands that are different from what you tested with. I'm running `helm upgrade redpanda redpanda` after modifying the redpanda section of `values.yaml` hoping to see the new config be applied to the redpanda cluster, but that doesn't happen.

I modify `enable_usage_stats` to `false`, and then use the following command to check config values in the first node of the cluster:
```
kubectl exec -it redpanda-0 -- rpk redpanda admin config print --host redpanda:9644
```

vuldin (approved):
I was able to verify this works. I modified `enable_coproc` (added to redpanda section of `values.yaml`) and ran the helm upgrade command. The pods restarted and the property value was updated. I think the issue with my previous test could have been that the property wasn't within the redpanda section... it was in the rpk section.

PR #50, opened 2022-05-10
Add dedicated parameters for memory flags

RafalKorepta:
The calculation for memory based on float number in limits memory is too complicated.
Reserved memory flag does not play nice with float number for memory declaration e.g. 0.4G

### DIFF
--- a/redpanda/ci/01-one-node-cluster-values.yaml
+++ b/redpanda/ci/01-one-node-cluster-values.yaml
@@ -17,3 +17,6 @@ statefulset:
   resources:
     limits:
       memory: 1Gi
+  redpanda:
+    memory: 700M
+    reservedMemory: 200M
--- a/redpanda/templates/statefulset.yaml
+++ b/redpanda/templates/statefulset.yaml
@@ -101,8 +101,8 @@ spec:
               redpanda
               start
               --smp={{ .Values.statefulset.resources.limits.cpu }}
-              --memory={{ template "redpanda.parseMemory" . }}
-              --reserve-memory=0M
+              --memory={{ .Values.statefulset.redpanda.memory }}
+              --reserve-memory={{ .Values.statefulset.redpanda.reservedMemory }}
               {{- if (first .Values.config.redpanda.kafka_api).external.enabled }}
               --advertise-kafka-addr=kafka://{{ template "redpanda.kafka.internal.advertise.address" . }}:{{ template "redpanda.kafka.internal.advertise.port" . }},external://{{ template "redpanda.kafka.external.advertise.address" . }}:{{ template "redpanda.kafka.external.advertise.nodeport.port" . }},
               --kafka-addr=kafka://{{ template "redpanda.kafka.internal.listen.address" . }}:{{ template "redpanda.kafka.internal.listen.port" . }},external://{{ template "redpanda.kafka.external.listen.address" . }}:{{ template "redpanda.kafka.external.listen.port" . }},
--- a/redpanda/values.yaml
+++ b/redpanda/values.yaml
@@ -160,7 +160,19 @@ statefulset:
   resources:
     limits:
       cpu: 1
-      memory: 2Gi
+      memory: 2.5Gi
+
+  # Redpanda memory parameter can not exceed POD resource memory limit
+  # Example calculation:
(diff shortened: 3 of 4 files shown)


redpanda/ci/01-one-node-cluster-values.yaml
```diff
@@ -17,3 +17,6 @@ statefulset:
   resources:
     limits:
       memory: 1Gi
```
vuldin, line 19: I believe this could be removed due to the associated lines being removed in `_helpers.tpl`

redpanda/values.yaml
```diff
@@ -160,7 +160,19 @@ statefulset:
   resources:
     limits:
       cpu: 1
-      memory: 2Gi
```
vuldin, line 163: I believe this could be removed due to the associated lines being removed in `_helpers.tpl`

vuldin (commented):
Non-blocking change request to possibly remove props that are no longer used.

redpanda/templates/statefulset.yaml
```diff
@@ -101,8 +101,8 @@ spec:
               redpanda
               start
               --smp={{ .Values.statefulset.resources.limits.cpu }}
```
vuldin, line 103: Ahh... this would also need to be changed or removed if this comment is approved: https://github.com/redpanda-data/helm-charts/pull/50/files#r869441863

redpanda/values.yaml
RafalKorepta: When I revert this change I see the following error as cgroup has less memory than combining:
```
--memory:2G
--reserve-memory=400k
```

an error:
```
exec /usr/bin/rpk redpanda start --smp=1 --memory=2G --reserve-memory=400k --advertise-kafka-addr=kafka://redpanda-0.redpanda.redpanda.svc.cluster.local.:9092,external://redpanda-0.streaming.rockdata.io:32005, --kafka-addr=kafka://10.244.0.10:9092,external://10.244.0.10:9093, --advertise-rpc-addr=redpanda-0.redpanda.redpanda.svc.cluster.local.:33145 --rpc-addr=10.244.0.10:33145 --schema-registry-addr=internal://10.244.0.10:8081,external://0.0.0.0:18081 --advertise-pandaproxy-addr=internal://redpanda-0.redpanda.redpanda.svc.cluster.local.:8082,external://172.18.0.2:8083 --pandaproxy-addr=internal://10.244.0.10:8082,external://10.244.0.10:8083 --default-log-level=info
...
Could not initialize seastar: std::runtime_error (insufficient physical memory: needed 2147483648 available 2147074048)
```

redpanda/ci/01-one-node-cluster-values.yaml
RafalKorepta: Nope. The kind cluster in Github action has less resources than we would like to have. We can run into issues that Redpanda will not start because of the insufficient memory.

redpanda/templates/statefulset.yaml
RafalKorepta: The smp is tight to cpu. It is recommended to have 2G per one smp (core), but with the `overprovisioned` flag set to `true` the strict restriction is removed. 

ref: https://github.com/redpanda-data/helm-charts/blob/74d3d106987b9aefd087f2654f39473bd24019b6/redpanda/values.yaml#L107

vuldin:
Hi @RafalKorepta , the following PR has recently been added which conflicts with this change: https://github.com/redpanda-data/helm-charts/pull/63

Do you have any thoughts on which one would be best?

PR #50 takes a less complicated approach to setting `memory` and `reserveMemory`, but leaving it up to the user may result in more cases where the user chooses invalid or problematic values. More would be left up to the user of the helm chart to know memory details or Redpanda, seastar, etc.

PR #63 attempts to manage the memory for the user, and better ensures the helm chart would "do the right thing" depending on how much memory the user gives the app. The downside is that it adds a little more complexity to the helm chart, and we may likely need to revisit this logic to ensure it will continue to properly set memory values.

Another option may be to merge both PRs, so that if optional memory values are set by the user then they are used. Otherwise we determine the best values. I think I like this third option the best.

RafalKorepta:
I agree with your third option. If my proposed Redpanda memory values are not set, then try to compute from resource limits

PR #51, opened 2022-05-11
Adding custom cert support

burandobata:
This way we can use custom certificates and e.g. configure different certs for internal and external traffic.


vuldin:
Hi @burandobata , I'm trying to ensure this PR works as expected but not sure how to get it working. I've tried both of the following commands to get the customer certificate section added to the generated statefulset config but it doesn't seem to be there with either:
```
helm -n helm-test install redpanda redpanda -f redpanda/values_add_custom_certs.yaml -f redpanda/values_add_tls.yaml
helm -n helm-test install redpanda redpanda -f redpanda/values_add_custom_certs.yaml
```

Do you have either steps to get to a working state, or a test to verify this works?

vuldin (approved):
Was able to verify these changes with the following commands:
```
minikube start
ark install cert-manager
k create ns helm-test
git checkout adding-custom-cert-support
helm -n helm-test install redpanda redpanda -f redpanda/values_add_custom_certs.yaml -f redpanda/values_add_tls.yaml
```
The new volume `tls-internal` is added to the `volumes` section with values from the `values_add_custom_certs.yaml` file.

PR #53, opened 2022-05-12
Adding more control over external port

burandobata:
Adding more control over external port, leaving defaults to what is was before.


vuldin (approved):
Verified changes

Issue #49, opened 2022-05-10
Setting --memory=(pod request memory) is too aggressive 


travisdowns:
Currently we set the `--memory` argument of redpanda to use 100.00% of the pod request memory, with no byte left over.

However, this `--memory` parameter doesn't set an absolute cap on the rp process memory use (as seen by cgroups): it is only for the primary memory map for the seastar allocator, but the redpanda process will use a bit than that: for file pages and page tables, at least: these are added the anonymous map size which will eventually grow to almost exactly the `--memory` value.

In practice this causes the rp process to be OOM killed by the kernel as it exceeds its cgroup budget.

So in addition to the larger buffer we need outside the container (for kernel memory, sidecards, etc) we need some buffer _inside_ the container. The page tables are only about 0.2% of the total memory. The file pages seem more variable, but let's say they won't exceed 200 MB.

So we can we use a `--memory` argument which is less than the request amount `R` by `0.002 * R + 200 Mi` ?

vuldin:
From slack:

```
So just to summarize, there are two different "reservations" (or set asides that RP won't use) we are talking about (would be easier on a whiteboard!):
1. The outer one, between the host RAM and the k8s pod limit. E.g., in this demonware case they have 392 GB on the host but the pod limit is 90 GB so they have spare for days (there are other things running there too but nothing very big). This outer buffer has to be fairly large b/c it encompasses all the socket buffers and other kernel memory would could run into the GBs. The 10% value is probably reasonable here (though we should probably try to quantify this more at some point and get to a formula which might look like 1 GB + max_open_connections * sockbuf_size.
2. The inner one, between the container limit (enforced by cgroup) and the redpanda --memory value. We don't need as much buffer here but setting it to 100.00% of the cgroup value doesn't work well either, per above.
```

```
So I think to be comfortable I'd say 200 MB + 0.2%, or if it has to be an absolute number encompassing also very high mem configs at least 1 GB (which is obvs way too big for small containers), since you could have a 400 GB pod without stretching it much and then 0.2% is 800 MB.
```

```
for --memory=200MB, 2MB makes sense… something like that up to 1G of memory. then you do 200MB for 2G of memory, then you do 200MB * .2% for the rest
```

```
I think we could say 200MB + 0.2% * mem since that's almost exactly the same as "200MB + 0.2% * (mem above 2GB)", they differ by only 4 MB and the former is simpler.
```

```
we can use a formula like 200 MB + 0.2% for the in-container reserve.
[...]

PR #63, opened 2022-05-19
reserve memory auto configuration added #49

redp01:
Reserve memory can be auto configured. It utilises the formula supplied by @travisdowns . There is potentially scope to override these values directly in the config without the automatic configuration though hopefully we can discuss on this PR.

It is related to issue #49 - note the available sprig math functions were not particularly helpful, hence the basic implementation.

### DIFF
--- a/redpanda/ci/01-one-node-cluster-values.yaml
+++ b/redpanda/ci/01-one-node-cluster-values.yaml
@@ -16,4 +16,4 @@ statefulset:
   replicas: 1
   resources:
     limits:
-      memory: 1Gi
+      memory: 2Gi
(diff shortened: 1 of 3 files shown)


travisdowns:
cc @BenPope

vuldin (changes requested):
Thanks for the PR @redp01 ! PR #50 was merged and introduced conflicts. After speaking with @RafalKorepta we agreed that merging both PRs would be the best approach: https://github.com/redpanda-data/helm-charts/pull/50#issuecomment-1134726911 . This will mean a minor change to this PR though to ensure that if specific memory values aren't given in the `values.yaml` then it defaults to the values this PR would provide. Let me know if you have any questions or concerns.

vuldin:
Hi @redp01, could this PR instead be pointing to v2 branch? We can discuss this at some point, but you likely already know the reason (that is the target for all next version changes).

vuldin:
Closing this since the changes were included and expanded upon in https://github.com/redpanda-data/helm-charts/pull/127

PR #116, opened 2022-07-14
Improve scripts and instructions for using them

vuldin:
CONTRIBUTING.md includes details around what has helped me while debugging and making improvements to the helm chart. This PR makes a few minor updates to the scripts outline, so that now they can be ran without any other prior setup and needing as few arguments as possible..

I also improved the instructions for how to use these scripts along with git to easily diff yaml file and Redpanda config.


vuldin:
I realized that I initially hard-coded the namespace and only pulled config from a single node. I pushed an update that makes namespace dynamic based on argument and also pulls the config for each broker in the statefulset.

r-vasquez (approved):

Issue #110, opened 2022-06-29
Most TLS tests expect the first listener in a group to have TLS enabled


vuldin:
One such test is `test-pandaproxy-internal-tls-status.yaml`.

Issue #109, opened 2022-06-29
Most tests fail if SASL is enabled


vuldin:
Ensure proper tests run depending on SASL status, and expand tests for when SASL is enabled. This is a bug because some tests that fail don't check whether SASL is enabled.

Issue #54, opened 2022-05-12
Handle centralized config


vuldin:
With `redpanda@22.1` comes centralized config. This removes all cluster-wide props from `redpanda.yaml` and requires using either `rpk config edit` or `rpk config import`. The `edit` command is seen as the primary method for setting cluster-wide props, but this doesn't work well with infrastructure-as-code scenarios.

The helm chart currently writes all values to `redpanda.yaml` and redpanda complains about this during startup. Interestingly enough, it seems to respect the configuration regardless. But we should separate cluster-level and node-level config into separate sections within `values.yaml`, and then only write the node-level config to `redpanda.yaml`. Cluster-wide config could be passed in once the cluster has started with `rpk config import` or included in the start command with conditional arguments (`redpanda start --set cloud_storage_enabled=true ...`).

vuldin:
Cluster values in `values.yaml` do get ignored in `v22.1.3` in testing. We need to:
1. detect Redpanda version, either "21.x and lower" or "22.x and higher"
2. continue as currently implemented if the former
3. if the latter, then write cluster-level props to `.bootstrap.yaml`. This method only works for setting props at startup, so this could cause confusion if the cluster props were updated, `helm update` was ran, and then these updated props didn't apply to the cluster. More testing and possibly more details on this process are needed.
3b) write cluster-level props to a temp `config/yaml`, and then use `rpk cluster config import` after the cluster is running. The benefit is this goes through an API, but the downside is certain cluster-level props require a restart (unsure which ones)

PR #117, opened 2022-07-15
Fixup central config

redp01:
@vuldin - moving the commits that were on your branch for PR85 onto v2.

Essentially this adds:

- upgrade to 22.1.4 and with it the need to use central config. 
- post-upgrade hook is also added to ensure that changes to the cluster values can be updated correctly by rpk with a helm update.
- sasl configuration has changed between versions, so amended this.
- The amendment to sasl configuration also highlighted that the helm tests were broken since a previous PR adjusted the TLS configuration. These have been fixed.
- Ordered Ready deployment used to facilitate lower down time upgrades

@vuldin - would be great if you could please review and assist in test cases for upgrades.
Questions: 
the values files between versions differ - i.e. sub headings for cluster, node, tuneables. Do we favour one format or handle both?
Updating the config triggers redeployment though one of the advantages of central config updates is often no restart necessary? Do we want to handle this in a helm update?


redpanda/templates/configmap.yaml
```diff
@@ -30,6 +30,21 @@ metadata:
...
+      superusers: ['{{ (first .).name }}']
+  {{- end }}
+{{- end }}
+{{- end }}
+ {{- with .Values.config.cluster }}
+    {{- toYaml . | nindent 6 }}
+  {{- end }}
+  {{- with .Values.config.tunable }}
+    {{- toYaml . | nindent 6 }}
+  {{- end }}
```
vuldin, line 47: It seems that `bootstrap.yaml` will be configured here with values from `values.yaml` regardless of the Redpanda version. I would expect to be able to use Redpanda `21.x` by modifying the version, and then the helm chart add the config to `redpanda.yaml` instead `bootstrap.yaml`.

redpanda/templates/post-upgrade.yaml
```diff
@@ -0,0 +1,85 @@
...
+
+{{- else if eq (include "tls-enabled" .) "true" }}
+            rpk --api-urls {{ template "redpanda.name" . }}-0.{{ include "redpanda.internal.domain" . }}:{{ template "redpanda.admin.port" $ }} cluster config import -f /tmp/base-config/bootstrap.yaml --admin-api-tls-enabled --admin-api-tls-truststore /etc/tls/certs/admin/ca.crt --tls-enabled --tls-truststore /etc/tls/certs/admin/ca.crt;
+{{- else if eq (include "tls-enabled" .) "true" }}
+            rpk --api-urls {{ template "redpanda.name" . }}-0.{{ include "redpanda.internal.domain" . }}:{{ template "redpanda.admin.port" $ }} cluster config import -f /tmp/base-config/bootstrap.yaml --user admin --password test --sasl-mechanism SCRAM-SHA-256;
+{{- else }}
+            rpk --api-urls {{ template "redpanda.name" . }}-0.{{ include "redpanda.internal.domain" . }}:{{ template "redpanda.admin.port" $ }} cluster config import -f /tmp/base-config/bootstrap.yaml;
+{{- end }}
```
vuldin, line 50: If I understand, this reconfigures the admin URLs to make use of TLS and/or SASL after the cluster has been started. This allows these auth features to be enabled and then applied after a cluster has already been created.

If that's the case, then what about the other listeners which may also make use of TLS/SASL? Would they also require modification at this point (and if not they would be in a broken state)?

redpanda/values.yaml
```diff
@@ -316,94 +316,104 @@ serviceAccount:
 # These values do not pertain to the kubernetes objects created with helm.
 # Instead these parameters get passed directly to the Redpanda binary at startup.
 # See https://docs.redpanda.com/docs/cluster-administration/configuration/
+# There are three subsections, cluster, tunable and node. The cluster and tunable sections
+# are written to the cluster .bootstrap.yaml and read into central config upon startup.
```
vuldin, line 320: Maybe this text should be taken out since it doesn't make sense for all Redpanda versions (older version wouldn't have `bootstrap.yaml` and would expect config to be written to `redpanda.yaml`).

vuldin (changes requested):
I mainly had a question around how to handle `values.yaml` config with previous versions. The PR is looking good, thanks!

redpanda/templates/configmap.yaml
redp01: @vuldin - I literally just added that question to the PR comments. It should be possible to alter the config map so depending on version either updates the configmap for redpanda.yaml or not. I will look into it now.

redpanda/templates/post-upgrade.yaml
redp01: @vuldin - so what this does is apply the cluster config and tuneables transitively from the changes to the values.yaml. It switches based on the tls / sasl mode in use. So its not changing the auth features - its just trying to apply the new config via the admin config import method which may be configured for tls, sasl or not.

vuldin:
It looks like this PR may fix #109 , fix #110 (let me know if you think this isn't the case).

redp01:
> It looks like this PR may fix #109 and #110 (let me know if you think this isn't the case).

Yes. #109 and generally #110 though we need to test with the individual service overrides - for example we could enable overarching TLS then set enabled: false for one service. Though this is certainly better as I would imagine its unlikely to enable TLS for one service and not all.

rockdata01:
> Also closes #54

Yes

redpanda/templates/configmap.yaml
rockdata01: Commit [b6e8ff8](https://github.com/redpanda-data/helm-charts/pull/117/commits/b6e8ff8e5ded707999865a629155e5ae8394e2c3) should now allow the old style values.yaml to be compatible with the older versions.

redpanda/values.yaml
rockdata01: @vuldin - the additional and potentially confusing lines are now removed in [24e9ee1](https://github.com/redpanda-data/helm-charts/pull/117/commits/24e9ee1aa87bcf44d9c39e677403c3e3c626ab21)

redpanda/templates/configmap.yaml
```diff
@@ -30,6 +30,21 @@ metadata:
...
+      superusers: ['{{ (first .).name }}']
+  {{- end }}
+{{- end }}
+ {{- with .Values.config.cluster }}
+    {{- toYaml . | nindent 6 }}
+  {{- end }}
+  {{- with .Values.config.tunable }}
+    {{- toYaml . | nindent 6 }}
+  {{- end }}
+{{- end }}
```
vuldin, line 47: I think there should be an `else` here, since we would want to take a different path for lower semver values (Redpanda versions without central config support). In this new `else` branch, `.Values.config.cluster` and `.Values.config.tunable` would be written to `redpanda.yaml` in the appropriate place.

redpanda/templates/configmap.yaml
```diff
@@ -30,6 +30,21 @@ metadata:
   {{- toYaml . | nindent 4 }}
 {{- end }}
 data:
+{{- if (include "redpanda.semver" . | semverCompare ">=22.1.4") }}
```
vuldin, line 33: I believe the correct version would be `22.1.1`: https://github.com/redpanda-data/redpanda/releases/tag/v22.1.1

redpanda/templates/configmap.yaml
```diff
@@ -39,30 +54,35 @@ data:
...
-{{- if .Values.auth.sasl.enabled }}
-  {{- with .Values.auth.sasl.users }}
+{{- if not (include "redpanda.semver" . | semverCompare ">=22.1.4") }}
+  {{- if .Values.auth.sasl.enabled }}
+    {{- with .Values.auth.sasl.users }}
       superusers:
     {{ toYaml . | indent 8 }}
+    {{- end }}
+  {{- end }}
+  {{- if not (eq (toYaml .Values.config) "null") }}
```
vuldin, line 64: This isn't a request to change, but is this the same as the following line?
```
{{- if not (hasKey .Values.listeners.rpc "tls") }}
```

redpanda/templates/post-upgrade.yaml
```diff
@@ -0,0 +1,85 @@
+{{- if (include "redpanda.semver" . | semverCompare ">=22.1.4") }}
```
vuldin, line 1: Same comment about semver version as before, I believe this should be `22.1.1`.

redpanda/templates/post-upgrade.yaml
```diff
@@ -0,0 +1,85 @@
...
+        image: {{ .Values.image.repository }}:{{ .Values.image.tag }}
+        command: ["/bin/sh", "-c"]
+        args:
+          - >
+{{- if and (eq (include "tls-enabled" .) "true")  (eq (include "sasl-enabled" .) "true") }}
+            rpk --api-urls {{ template "redpanda.name" . }}-0.{{ include "redpanda.internal.domain" . }}:{{ template "redpanda.admin.port" $ }} cluster config import -f /tmp/base-config/bootstrap.yaml --admin-api-tls-enabled --admin-api-tls-truststore /etc/tls/certs/admin/ca.crt --tls-enabled --tls-truststore /etc/tls/certs/admin/ca.crt --user admin --password test --sasl-mechanism SCRAM-SHA-256;
+
+{{- else if eq (include "tls-enabled" .) "true" }}
+            rpk --api-urls {{ template "redpanda.name" . }}-0.{{ include "redpanda.internal.domain" . }}:{{ template "redpanda.admin.port" $ }} cluster config import -f /tmp/base-config/bootstrap.yaml --admin-api-tls-enabled --admin-api-tls-truststore /etc/tls/certs/admin/ca.crt --tls-enabled --tls-truststore /etc/tls/certs/admin/ca.crt;
+{{- else if eq (include "tls-enabled" .) "true" }}
```
vuldin, line 46: Aren't lines 44 and 46 identical, which would mean line 44 would potentially pass to line 45 and then line 46 would never pass to line 47?

redpanda/templates/post-upgrade.yaml
```diff
@@ -0,0 +1,85 @@
...
+    spec:
+      restartPolicy: Never
+      containers:
+      - name: {{ template "redpanda.name" . }}-post-upgrade
+        image: {{ .Values.image.repository }}:{{ .Values.image.tag }}
+        command: ["/bin/sh", "-c"]
+        args:
+          - >
+{{- if and (eq (include "tls-enabled" .) "true")  (eq (include "sasl-enabled" .) "true") }}
+            rpk --api-urls {{ template "redpanda.name" . }}-0.{{ include "redpanda.internal.domain" . }}:{{ template "redpanda.admin.port" $ }} cluster config import -f /tmp/base-config/bootstrap.yaml --admin-api-tls-enabled --admin-api-tls-truststore /etc/tls/certs/admin/ca.crt --tls-enabled --tls-truststore /etc/tls/certs/admin/ca.crt --user admin --password test --sasl-mechanism SCRAM-SHA-256;
```
vuldin, line 42: I see this job run `rpk cluster config import /tmp/base-config/bootstrap.yaml`. I don't remember why this needs to be ran though, do you have details? I will look more into it tomorrow also but wanted to make a comment to remember.

redpanda/templates/statefulset.yaml
```diff
@@ -70,6 +70,9 @@ spec:
               CONFIG=/etc/redpanda/redpanda.yaml;
               NODE_ID=${SERVICE_NAME##*-};
               cp /tmp/base-config/redpanda.yaml "$CONFIG";
+              {{- if (include "redpanda.semver" . | semverCompare ">=22.1.4") }}
```
vuldin, line 73: Same as previous comments: semver should be `22.1.1`

vuldin (changes requested):
Thanks for the update to this PR, it's starting to look good. I need to do more testing locally running the code, but I went ahead and added my questions and comments so you would have them at the start of your day. Let me know if you have any questions/comments.

redpanda/templates/configmap.yaml
vuldin: Discussion continued [here](https://github.com/redpanda-data/helm-charts/pull/117#discussion_r929542045)

redpanda/templates/post-upgrade.yaml
vuldin: I found the answer to this question: `.bootstrap.yaml` applies config on first start only, while `rpk cluster config import` can be set config for a running cluster (and does so through the admin API).

redpanda/templates/statefulset.yaml
redp01: Fixed in [519e4f1](https://github.com/redpanda-data/helm-charts/pull/117/commits/519e4f193b60c4e2410f2b1042c3eb2d958f5ea7)

redpanda/templates/post-upgrade.yaml
redp01: [519e4f1](https://github.com/redpanda-data/helm-charts/pull/117/commits/519e4f193b60c4e2410f2b1042c3eb2d958f5ea7)

redpanda/templates/post-upgrade.yaml
redp01: I also reuse the bootstrap.yaml file (through the mapped config) to push changes to the API

redpanda/templates/configmap.yaml
redp01: Its pulling the old config format through lines 64-66

redp01: What this means is that if you have an older version of the values.yaml format, it is implied that you have transformed it to the cluster/tunables config format before you upgrade. If its lower than 22.1.1 then the old format should work.

redpanda/templates/configmap.yaml
redp01: Fixed in [519e4f1](https://github.com/redpanda-data/helm-charts/pull/117/commits/519e4f193b60c4e2410f2b1042c3eb2d958f5ea7)

redpanda/templates/configmap.yaml
redp01: We need to think about documenting the upgrade paths .......

redpanda/templates/post-upgrade.yaml
redp01: @vuldin - looks like that should be "sasl-enabled" only. Will update -  though logically would it be realistic for users to use this without tls?

redpanda/templates/configmap.yaml
vuldin: We discussed this internally giving two options: 1) one `values.yaml` that works across all Redpanda versions and 2) requiring the user to specify an older Redpanda version as well as properly format the `config` section for that version.

We decided to go with option 2 since this will be less code in the helm chart (and potential for issues down the road), and it would be a more deliberate path for the user to choose an older version (considering we would rather support new Redpanda versions). We will provide documentation on using older Redpanda versions with the latest helm chart, which should help users with this.

redpanda/templates/configmap.yaml
vuldin: After testing I see that these sections appear as `key: null` in the configmap output (rather than not existing). I think this is also related to a bug we discussed internally with using an older Redpanda version, where the cluster won't start with this branch.

redpanda/templates/post-upgrade.yaml
vuldin: We discussed internally and it seems we decided to continue allowing SASL without TSL.

redpanda/Chart.yaml
```diff
@@ -21,7 +21,7 @@ maintainers:
     url: https://github.com/orgs/redpanda-data/people
 type: application
 version: 2.0.0-alpha.0
-appVersion: 21.11.16
+appVersion: 22.1.4
```
vuldin, line 24:
```suggestion
appVersion: 22.1.5
```
This is a moving target, but as of now `22.1.5` is the latest.

redpanda/values.yaml
```diff
@@ -34,7 +34,7 @@ commonLabels: {}
 image:
   repository: vectorized/redpanda
   # Redpanda version. This determines the installed version (not Chart.appVersion)
-  tag: v21.11.16
+  tag: v22.1.4
```
vuldin, line 37:
```suggestion
  tag: v22.1.5
```

redpanda/values.yaml
```diff
@@ -228,7 +228,7 @@ statefulset:
   replicas: 3
   updateStrategy:
     type: RollingUpdate
-  podManagementPolicy: Parallel
+  podManagementPolicy: OrderedReady
```
vuldin, line 231: Did you notice any problems that led you to make this change? Do you see any differences in how brokers are aligned with nodes? Just interested in why this change now, and I do think it logically makes sense.

redpanda/Chart.yaml
redp01: resolved [46f6b62](https://github.com/redpanda-data/helm-charts/pull/117/commits/46f6b628b22162234c01501ab2e222b359aebc5e)

redpanda/values.yaml
redp01: resolved [46f6b62](https://github.com/redpanda-data/helm-charts/pull/117/commits/46f6b628b22162234c01501ab2e222b359aebc5e)

redpanda/templates/configmap.yaml
```diff
@@ -89,6 +109,9 @@ data:
       - host:
           address: "{{ template "redpanda.fullname" $ }}-{{ . }}.{{ template "redpanda.internal.domain" $ }}"
           port: {{ $values.listeners.rpc.port }}
+{{- end }}
+{{- if not (eq (toYaml .Values.config.node) "null") }}
+  {{ toYaml .Values.config.node | indent 4 }}
```
vuldin, line 114: While testing with `21.11.17` I run into an issue due to this line, where it expects that `.Values.config.node` exists. The issue is that this section shouldn't exist at all in `21.x` configs.

A possible change to solve this would be the following:
```suggestion
{{- if hasKey .Values "config" }}
  {{- if not (eq (toYaml .Values.config.node) "null") }}
    {{ toYaml .Values.config.node | indent 4 }}
  {{- end }}
```

vuldin (changes requested):
One issue while testing with `21.x`

redpanda/templates/configmap.yaml
redp01: @vuldin - I have updated the configmap.yaml to be more tolerant to old versions with newer values.yaml files. New versions will require an explicit config change. Even if the ingest of values into central config is successful the declarative nature of the values file will not necessarily match the config in the central config. Note: upgrading values using the new format with an old values.yaml file was successful.

vuldin (approved):

redpanda/templates/tests/test-api-status.yaml
```diff
@@ -15,7 +15,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */}}
 {{- /* TODO test fails if SASL is enabled */}}
```
vuldin, line 17: Should this line be deleted after your changes to enable running SASL without TLS (non-blocker)?

redpanda/templates/configmap.yaml
```diff
@@ -39,30 +54,49 @@ data:
...
     {{- end }}
     redpanda:
-{{- if .Values.auth.sasl.enabled }}
-  {{- with .Values.auth.sasl.users }}
+{{- if not (include "redpanda.semver" . | semverCompare ">=22.1.1") }}
+  {{- if .Values.auth.sasl.enabled }}
+    {{- with .Values.auth.sasl.users }}
       superusers:
-    {{ toYaml . | indent 8 }}
+    {{ toYaml . | indent 4 }}
```
vuldin, line 61: I believe L60-L61 should be the same as L38:
```suggestion
      superusers: ['{{ (first .).name }}']
```

vuldin (changes requested):
I had one issue with getting a successful deployment when using Redpanda `v21.11.18`, related to properly reading the superuser from `values.yaml`.

vuldin:
Resolves #109 . We will still want to create more tests around SASL outside of this PR.

redpanda/templates/configmap.yaml
redp01: Fixed in [60a4927](https://github.com/redpanda-data/helm-charts/pull/117/commits/60a4927e61db5d117d538345f8b246d1fca73a0b) - I think this was a throw back to the alternative list notation.

redpanda/templates/NOTES.txt
```diff
@@ -65,9 +67,17 @@ Delete the topic:
 
 {{- if and $anyTLS $anySASL }}
```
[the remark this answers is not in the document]
redp01: @vuldin - thanks. Fixed in [60a4927](https://github.com/redpanda-data/helm-charts/pull/117/commits/60a4927e61db5d117d538345f8b246d1fca73a0b)

redpanda/templates/NOTES.txt
```diff
@@ -65,9 +67,17 @@ Delete the topic:
...
+Request the certificates:
+
+CERT=$(kubectl exec -n {{ .Release.Namespace }} {{ template "redpanda.name" . }}-0 -c redpanda -- cat /etc/tls/certs/kafka/ca.crt)
 
-CERT=$(kubectl exec -n {{ .Release.Namespace }} {{ template "redpanda.name" . }}-0 -c redpanda -- cat /etc/tls/certs/ca.crt)
+ADMINCERT=$(kubectl exec -n {{ .Release.Namespace }} {{ template "redpanda.name" . }}-0 -c redpanda -- cat /etc/tls/certs/admin/ca.crt)
+
+Create admin user:
+
+ kubectl -n redpanda run rpk -ti --rm --restart=Never --image {{ .Values.image.repository }}:{{ .Values.image.tag }} --command -- /bin/bash -c "rpk acl user create admin --api-urls {{ template "redpanda.name" . }}-0.{{ include "redpanda.internal.domain" . }}:{{ .Values.listeners.admin.port }} --user admin --password test --sasl-mechanism  SCRAM-SHA-256 --admin-api-tls-enabled --admin-api-tls-truststore <(echo '$ADMINCERT')"
```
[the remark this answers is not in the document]
redp01: @vuldin - fixed in [60a4927](https://github.com/redpanda-data/helm-charts/pull/117/commits/60a4927e61db5d117d538345f8b246d1fca73a0b)

vuldin (approved):

PR #123, opened 2022-07-21
Cleanup storage section

### LINKED ISSUE #79: Expand top-level `storage` parameter

vuldin:
- remove storage.enabled
- lower default PV size to 3Gi
- add initContainer that properly changes ownership on mounted volume

A good side effect of changes in this PR is that minikube now works as expected with persistent volumes (see comments in #79).

closes #79

### DIFF
--- a/redpanda/templates/statefulset.yaml
+++ b/redpanda/templates/statefulset.yaml
@@ -57,6 +57,12 @@ spec:
       securityContext:
 {{- toYaml .Values.statefulset.podSecurityContext | nindent 8 }}
       initContainers:
+        - name: set-datadir-ownership
+          image: busybox:latest
+          command: ["/bin/sh", "-c", "chown 101:101 -R /var/lib/redpanda/data"]
+          volumeMounts:
+            - name: datadir
+              mountPath: /var/lib/redpanda/data
         - name: {{ template "redpanda.name" . }}-configurator
           image: {{ .Values.image.repository }}:{{ .Values.image.tag }}
           command: ["/bin/sh", "-c"]
--- a/redpanda/values.yaml
+++ b/redpanda/values.yaml
@@ -168,10 +168,7 @@ listeners:
 logLevel: info
 #
 # Persistence
-# TODO modify to conform to https://github.com/bitnami/charts/blob/master/bitnami/kafka/values.yaml#L888
 storage:
-  # TODO make use of enabled parameter
-  enabled: true
   # Absolute path on host to store Redpanda's data.
   # If not specified, then `emptyDir` will be used instead.
   # If specified, but `persistentVolume.enabled` is `true`, then has no effect.
@@ -179,10 +176,8 @@ storage:
   # If `enabled` is `true` then a PersistentVolumeClaim will be created and
   # used to store Redpanda's data, otherwise `hostPath` is used.
   persistentVolume:
-    # TODO possibly replace with storage.enabled
     enabled: true
-    # TODO lower default size to a more reasonable size (10 - 50 Mi)
-    size: 100Gi
+    size: 3Gi
     # If defined, then `storageClassName: <storageClass>`.
     # If set to "-", then `storageClassName: ""`, which disables dynamic
     # provisioning.


redpanda/templates/statefulset.yaml
```diff
@@ -57,6 +57,12 @@ spec:
       securityContext:
 {{- toYaml .Values.statefulset.podSecurityContext | nindent 8 }}
       initContainers:
+        - name: set-datadir-ownership
+          image: busybox:latest
+          command: ["/bin/sh", "-c", "chown 101:101 -R /var/lib/redpanda/data"]
+          volumeMounts:
+            - name: datadir
+              mountPath: /var/lib/redpanda/data
```
r-vasquez, line 65: Is there a case where the user might want to change the default data_directory (/var/lib/redpanda/data)?

vuldin: I think changing the default directory in a kubernetes (or helm) situation would be much less likely (than local install) since this is the container path, and this path is just a mount point for a persistent volume which could be multiple places (a completely different directory on the node, a remote storage device through some persistent storage cloud service, etc.). I can't think of a scenario right now where the customer would be using this path, unless they are connecting directly into the pod for some reason.

At the same time, this would be easy to provide as a modifiable parameter in `values.yaml` if we can think of a good reason to do so.

r-vasquez: Thanks! It's clear to me now :)

r-vasquez (approved):

PR #127, opened 2022-07-28
Cleanup how resources and tuning parameters are handled

vuldin:
- new templates for getting container, redpanda, and reserved memory
- new logging section
- new resources section
- calculation of redpanda/reserved memory based on container memory
- remove rpk section
- new tuning section
- remove resources config from statefulset section

### DIFF
--- a/redpanda/ci/01-one-node-cluster-values.yaml
+++ b/redpanda/ci/01-one-node-cluster-values.yaml
@@ -14,9 +14,3 @@
 # limitations under the License.
 statefulset:
   replicas: 1
-  resources:
-    limits:
-      memory: 1Gi
-  redpanda:
-    memory: 700M
-    reservedMemory: 200M
--- a/redpanda/templates/configmap.yaml
+++ b/redpanda/templates/configmap.yaml
@@ -141,9 +141,30 @@ data:
 {{- else }}
       enable_memory_locking: false
 {{- end }}
-      tune_cpu: false
-      tune_disk_irq: false
-{{ toYaml .Values.tuning | indent 6 }}
+{{- if hasKey .Values.tuning "tune_aio_events" }}
+  {{- if .Values.tuning.tune_aio_events }}
+      tune_aio_events: true
+  {{- end }}
+{{- end }}
+{{- if hasKey .Values.tuning "tune_clocksource" }}
+  {{- if .Values.tuning.tune_clocksource }}
+      tune_clocksource: true
+  {{- end }}
+{{- end }}
+{{- if hasKey .Values.tuning "tune_ballast_file" }}
+  {{- if .Values.tuning.tune_ballast_file }}
+      tune_ballast_file: true
+  {{- end }}
+{{- end }}
+{{- if hasKey .Values.tuning "ballast_file_path" }}
+      ballast_file_path: {{ .Values.tuning.ballast_file_path }}
+{{- end }}
+{{- if hasKey .Values.tuning "ballast_file_size" }}
+      ballast_file_size: {{ .Values.tuning.ballast_file_size }}
+{{- end }}
+{{- if hasKey .Values.tuning "well_known_io" }}
+      well_known_io: {{ .Values.tuning.well_known_io }}
+{{- end }}
 {{- if .Values.listeners.http.enabled }}
     pandaproxy:
       pandaproxy_api_tls:
--- a/redpanda/templates/configmap.yaml
+++ b/redpanda/templates/configmap.yaml
@@ -50,8 +50,13 @@ data:
     {{- if gt (len .Values.license_key) 0 }}
     license_key: {{ .Values.license_key }}
     {{- end }}
-    {{- if gt (len .Values.organization) 0 }}
-    organization: {{ .Values.organization }}
+    {{- if .Values.logging.usageStats.enabled }}
+      {{- if hasKey .Values.logging.usageStats "organization" }}
+    organization: {{ .Values.logging.usageStats.organization }}
+      {{- end }}
+      {{- if hasKey .Values.logging.usageStats "clusterId" }}
+    cluster_id: {{ .Values.logging.usageStats.clusterId}}
+      {{- end }}
     {{- end }}
     redpanda:
 {{- if not (include "redpanda.semver" . | semverCompare ">=22.1.1") }}
@@ -125,7 +130,20 @@ data:
           port: {{ $values.listeners.rpc.port }}
 {{- end }}
     rpk:
-{{ toYaml .Values.rpk | indent 6 }}
+      enable_usage_stats: {{ .Values.logging.usageStats.enabled }}
+{{- if hasKey .Values.resources.cpu "overprovisioned" }}
+      overprovisioned: {{ .Values.resources.cpu.overprovisioned }}
+{{- else }}
+      overprovisioned: false
+{{- end }}
+{{- if hasKey .Values.resources.memory "enable_memory_locking" }}
+      enable_memory_locking: {{ .Values.resources.memory.enable_memory_locking }}
+{{- else }}
+      enable_memory_locking: false
+{{- end }}
+      tune_cpu: false
+      tune_disk_irq: false
+{{ toYaml .Values.tuning | indent 6 }}
 {{- if .Values.listeners.http.enabled }}
     pandaproxy:
       pandaproxy_api_tls:
--- a/redpanda/values.yaml
+++ b/redpanda/values.yaml
@@ -39,8 +39,6 @@ image:
   pullPolicy: IfNotPresent
 # Your license key (optional)
 license_key: ""
-# Your organization name (optional)
-organization: ""
 #
 # Authentication
 auth:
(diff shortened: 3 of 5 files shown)


redpanda/ci/01-one-node-cluster-values.yaml
```diff
@@ -14,9 +14,3 @@
 # limitations under the License.
 statefulset:
   replicas: 1
-  resources:
-    limits:
-      memory: 1Gi
-  redpanda:
-    memory: 700M
-    reservedMemory: 200M
```
vuldin, line 22: These values are removed since there are checks on memory given to Redpanda and the container; if theses are lower than recommended values (2Gi per core minimum), then an error is thrown: https://github.com/vuldin/helm-charts/blob/5bcb118e5dacac8aa154a351ae8c6851187931aa/redpanda/templates/_helpers.tpl#L381-L386

redpanda/values.yaml
```diff
@@ -163,9 +161,108 @@ listeners:
...
+    # Redpanda makes use of a thread per core model described here:
+    # https://redpanda.com/blog/tpc-buffers
+    # For this reason, Redpanda should only be given full cores (cores parameter below).
+    #
+    # NOTE: You can increase cores, but decreasing cores is not currently supported:
+    # https://github.com/redpanda-data/redpanda/issues/350
+    #
+    # Equivalent to: --smp, resources.requests.cpu, and resources.limits.cpu
+    # For production: 4 or greater
+    cores: 1
```
vuldin, line 204: Kubernetes allows containers in a statefulset to have a variable CPU value based on `resources.requests.cpu` and `resources.limits.cpu`. Making these values equal results in CPU affinity which is good performance. At the same time, Redpanda takes a single `--smp` setting which can only be increased.

It doesn't seem that a variable CPU setting is very compatible with a Redpanda deployment, so this change provides a single `resources.cpu.cores` parameter for the user. This value is used to set both `requests` and `limits` CPU values in the statefulset, resulting in CPU affinity.

redpanda/values.yaml
```diff
@@ -163,9 +161,108 @@ listeners:
...
+    #
+    # Overprovisioned means Redpanda won't assume it has all of the provisioned CPU.
+    # This should be true unless the container has CPU affinity (eg. min and max above are equal).
+    # Equivalent to: --idle-poll-time-us 0 --thread-affinity 0 --poll-aio 0
+    # overprovisioned: false
```
vuldin, line 209: Because of the above change to how CPU is handled, `overprovisioned` is false by default.

redpanda/values.yaml
```diff
@@ -163,9 +161,108 @@ listeners:
...
+  #
+  memory:
+    # Enables memory locking.
+    # For production: true
+    # enable_memory_locking: false
```
vuldin, line 214: I would have liked to provided more details on what this `enable_memory_locking` parameter does.

redpanda/values.yaml
```diff
@@ -163,9 +161,108 @@ listeners:
...
+      # resources.memory.container.max).
+      # Equivalent to: --memory
+      # For production: 8Gi or greater
+      # memory: 2Gi
+      #
+      # Memory reserved for the Seastar subsystem.
+      # Any value above 1Gi will provide diminishing performance benefits.
+      # Equivalent to: --reserve-memory
+      # For production: 1Gi
+      # reserveMemory: 200Mi
```
vuldin, line 265: All memory values, including these which go to Seastar, are in base-2 digits and also use the "correct" base-2 suffix (Ki, Mi, Gi). This is change from how Seastar handles memory values underneath. These values here are passed through the new helm template `redpanda-memoryToMi` to appropriately handle the base-2 suffix given here. The output is a digit, which is then combined with Seastar's expected suffix `M` in `configmap.yaml`.

redpanda/templates/configmap.yaml
```diff
@@ -91,7 +96,20 @@ data:
...
+{{- else }}
+      overprovisioned: false
+{{- end }}
+{{- if hasKey .Values.resources.memory "enable_memory_locking" }}
+      enable_memory_locking: {{ .Values.resources.memory.enable_memory_locking }}
+{{- else }}
+      enable_memory_locking: false
+{{- end }}
+      tune_cpu: false
+      tune_disk_irq: false
```
r-vasquez, line 111: I think we can delete this , rpk will strip false values anyway (https://github.com/redpanda-data/redpanda/commit/81b6236ec06d7554caec841fe1a89df8bdf1be00), or do we keep them here so the users know that this can't be set as true?

vuldin: Ahh interesting, I didn't realize they were stripped out if `false`. I added them here just because I wasn't sure of the default value, and I knew they needed to always be false when running in a container. If the default for these values are false, then these can be removed.

Another layer we could have is to ensure these parameters are ignored if the user adds them in manually... a long shot but possible given these details can be found throughout our docs.

vuldin: I see the documentation lists these two parameters as `false` by default, so I'll remove them completely.

r-vasquez: Agree with the layer to ignore these parameters, if it's outside of the scope of this PR maybe a comment in `Values.yaml` could be useful.

vuldin: Both (comments in `values.yaml` and ignoring these if manually added) are easy to do and I'll add them in when removing these parameters from `configmap.yaml`.

vuldin: Made this change here: https://github.com/redpanda-data/helm-charts/pull/127/commits/1f9af5277b045b0c6c02376f46290f739f88723c

redpanda/values.yaml
```diff
@@ -311,6 +359,68 @@ serviceAccount:
...
+  # Disables IOPS merging.
+  # tune_disk_nomerges: false
+  #
+  # tune_disk_irq must be false in container environments and will be ignored.
+  #
+  # Installs a systemd service to run fstrim weekly, or starts the default fstrim service
+  # which comes with most Linux distributions.
+  # tune_fstrim: false
+  #
+  # tune_cpu must be false in container environments and will be ignored.
```
vuldin, line 387: This section was expanded to include all `tune_` parameters from the documentation.

Two parameters are mentioned and ignored: `tune_disk_irq` and `tune_cpu`. I've seen messages saying these parameters cannot be used in a container. Let me know if there are other tune parameters that should also be ignored.

redpanda/templates/configmap.yaml
r-vasquez: Awesome, thanks!

redpanda/values.yaml
r-vasquez: I think only aio_events, ballast_file and clocksource work in a containerized environment. Just ran a test and got:

```
$ redpanda git:(v2) kubectl exec -it -n redpanda redpanda-0 -- rpk redpanda tune all                         
Defaulted container "redpanda" out of: redpanda, set-datadir-ownership (init), redpanda-configurator (init)
got an error while writing 1 to /proc/sys/vm/swappiness: open /proc/sys/vm/swappiness: read-only file system
  TUNER                  APPLIED  ENABLED  SUPPORTED  ERROR                                                      
  aio_events             true     true     true                                                                  
  ballast_file           true     true     true                                                                  
  clocksource            true     true     true                                                                  
  coredump               false    true     true       open                                                       
                                                      /proc/sys/kernel/core_pattern:                             
                                                      read-only file system                                      
  cpu                    false    true     true       open                                                       
                                                      /sys/devices/system/cpu/cpufreq/policy0/scaling_governor:  
                                                      read-only file system                                      
  disk_irq               false    true     true       no such file or                                            
                                                      directory                                                  
  disk_nomerges          false    true     false      no such file or                                            
                                                      directory                                                  
  disk_scheduler         false    true     false      no such file or                                            
                                                      directory                                                  
  disk_write_cache       false    true     false      no such file or                                            
                                                      directory                                                  
  fstrim                 false    true     false      dial unix                                                  
                                                      /var/run/dbus/system_bus_socket:                           
                                                      connect: no such file or                                   
                                                      directory                                                  
  net                    false    true     true       open                                                       
                                                      /proc/sys/net/core/rps_sock_flow_entries:                  
                                                      no such file or directory                                  
  swappiness             false    true     true       open                                                       
                                                      /proc/sys/vm/swappiness:                                   
                                                      read-only file system                                      
  transparent_hugepages  false    true     true       open                                                       
                                                      /sys/kernel/mm/transparent_hugepage/enabled:               
                                                      read-only file system    
```
I used V2 branch to setup a 3 node cluster using kind

vuldin: Ahh that's a great that you can find out what can be ran in a container environment based on running the command. I'll handle all these invalid parameters the same as `tune_disk_irq` and `tune_cpu`.

vuldin:
Last commit resolved merge conflicts and addressed feedback from  @r-vasquez (removing tuning parameters not suitable for a container environment).

PR #132, opened 2022-07-29
Remove address from listeners

vuldin:
This is a small PR to remove the address from listeners in `values.yaml`.


joejulian (approved):
This looks great!

PR #140, opened 2022-08-11
docs: values.yaml documentation iteration

joejulian:
As a new user, I found some of the values difficult to understand what they were doing and when and why they were used.
These updates come from my discovery process and will hopefully make it easier on a new user.

### DIFF
--- a/redpanda/values.yaml
+++ b/redpanda/values.yaml
@@ -20,11 +20,14 @@
 
 # Common parameters
 #
-# Override redpanda.name template
+# Override the app.kubernetes.io/name label
 nameOverride: ""
-# Override redpanda.fullname template
+# Objects are named based on the name of the helm release. fullnameOverride,
+# overrides this behavior naming resources as specified.
 fullnameOverride: ""
-# Default kuberentes cluster domain
+# Cluster's default DNS domain.
+# You should overwrite it if you're using a different one,
+# otherwise Redpanda node discovery and TLS certificates won't work.
 clusterDomain: cluster.local
 # Additional labels added to all Kubernetes objects
 commonLabels: {}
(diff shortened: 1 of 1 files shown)


redpanda/values.yaml
```diff
@@ -566,3 +467,114 @@ config:
   # rpc_server_tls                                TLS configuration for RPC server                                validate
   # seed_servers                                List of the seed servers used to join current cluster; If the seed_server list is empty the node will be a cluster root and it will form a new cluster                                None
   # superusers                                List of superuser usernames                                None
+
+#
+# Listener overrides
```
vuldin, line 472: Is there a reason why this section was moved to the bottom? The reason it was closer to the top was because this is an important section that will likely need to be updated by the user based on their environment. With this change, this section is at the bottom past other sections that are both long and also not likely to need to be changed (and are mostly commented out). The original idea was that the sections would be primarily in order based on importance and then alphabetical.

vuldin (changes requested):
Minor question around the movement of the `listeners` section.

vuldin:
Thanks for the PR @joejulian . Have you been able to sign the contributor license? Post if you have any issues with this step.

redpanda/values.yaml
joejulian: I moved it because it overrides generalized settings. Every setting above this has global consequences, the majority of the things that would be changed in the listeners section are overrides for things like tls enablement or external access.

joejulian: From a user perspective, I was going through enabling all the external access for everything so I could test it. After I got done doing that, I found the single global configuration I could have done it with.

joejulian:
Closing this in favor of #143

PR #153, opened 2022-08-26
Add startupProbe and make probes configurable

### LINKED ISSUE #150: allow configuration of liveness probe
1. liveness probe in the chart uses the admin port.
2. admin port is only started up when the partitions are in a health state. (eg.: `cluster - state_machine.cc:29 - Starting state machine for ntp={kafka/{my-topic}/122}` is executed for all topics/partitions)
3. it is not possible to specify liveness probe values in the helm-chart, this becomes an issue when a cluster tries to recover

in our cluster 2. (which have 2 topics, each with 128 partitions) takes about 5 minutes, but the default liveness values are not sufficient (by default there are 3 retries at 10 seconds intervals.)
thus currently in our cluster redpanda keeps crashlooping

_Update:_ 

additional context

once the admin port started up after 5 minutes, the readiness/liveness probe still failed:
```
{"is_healthy": false, "controller_id": 2, "all_nodes": [0,2,1], "nodes_down": [0,1], "leaderless_partitions": [... lots of lots of partitions here...]
```
there were lots of `leaderless_partitions`, which took about 30 minutes to go down. once that list became empty, the cluster became healthy/ready.

joejulian:
Fixes #150

### DIFF
--- a/redpanda/templates/statefulset.yaml
+++ b/redpanda/templates/statefulset.yaml
@@ -109,6 +109,24 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: status.hostIP
+          startupProbe:
+            exec:
+              command:
+                - /bin/sh
+                - -c
+                {{- if (include "admin-internal-tls-enabled" . |fromJson).bool }}
+                - >
+                  curl https://localhost:{{ .Values.listeners.admin.port }}/v1/cluster/health_overview
+                  -svk --cacert /etc/tls/certs/{{ .Values.listeners.admin.tls.cert }}/ca.crt |
+                  grep '"is_healthy": true'
+                {{- else }}
+                - >
+                  curl -sv http://localhost:{{ .Values.listeners.admin.port }}/v1/cluster/health_overview |
+                  grep '"is_healthy": true'
+                {{- end }}
+            initialDelaySeconds: {{ .Values.statefulset.startupProbe.initialDelaySeconds }}
+            failureThreshold: {{ .Values.statefulset.startupProbe.failureThreshold }}
+            periodSeconds: {{ .Values.statefulset.startupProbe.periodSeconds }}
           livenessProbe:
             exec:
               command:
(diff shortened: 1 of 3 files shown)


redpanda/templates/statefulset.yaml
```diff
@@ -109,6 +109,24 @@ spec:
...
+                {{- if (include "admin-internal-tls-enabled" . |fromJson).bool }}
+                - >
+                  curl https://localhost:{{ .Values.listeners.admin.port }}/v1/cluster/health_overview
+                  -svk --cacert /etc/tls/certs/{{ .Values.listeners.admin.tls.cert }}/ca.crt |
+                  grep '"is_healthy": true'
+                {{- else }}
+                - >
+                  curl -sv http://localhost:{{ .Values.listeners.admin.port }}/v1/cluster/health_overview |
+                  grep '"is_healthy": true'
+                {{- end }}
```
ekovacs, line 126: this snippet is now triplicated. it could be put into the `_helpers.tpl` and in each of these places it could be called as a template

ekovacs:
@joejulian
this is a bit tangential question, but i was wondering why the liveness check uses the same snippet as the readiness probe?

during startup redpanda creates a lockfile:
```
Writing pid file "/var/lib/redpanda/data/pid.lock"
```
 liveness probe is meant to check for the existence of a process
i was wondering, wouldn't it be better to `stat` that file?

joejulian:
Liveness checks are also intended to be used for catching deadlocks.

redpanda/templates/statefulset.yaml
joejulian: I think I'd like to leave this as-is, at this time, for readability. I'm also not convinced that one or more of these might change in the near future. Let's leave this for now and revisit after this chart gets more adoption and feedback. wdyt?

vuldin (approved):
LGTM

PR #158, opened 2022-09-07
fix readiness probe to only report readiness of itself

joejulian:
The liveness probe was checking more than just liveness. Changed it to just report success if it can query an api endpoint.
The readiness probe was failing if any node on the cluster was down. This was causing the entire cluster to become unavailable if a pod was restarted. Changed it to just check for itself being reported as down by the redpanda health overview.


ievgen-novytskyi:
Cluster seems to be ok now and can tolerate nodes restarts. Thanks.

PR #162, opened 2022-09-14
add job to set/update sasl superusers and license

### LINKED ISSUE #154: Issue applying license key
A user reported that adding a license key doesn't work as expected. They set `.Values.license_key` and verified that this makes it into `redpanda.yaml`, but once the cluster is up it doesn't detect the license:

```shell
> rpk cluster license info
this cluster is missing a license
```

joejulian:
Remove non-functional license_key from the redpanda.yaml and implement a post-install/post-upgrade job that sets up the superusers and the license.

fixes #154

### DIFF
--- a/redpanda/Chart.yaml
+++ b/redpanda/Chart.yaml
@@ -20,7 +20,7 @@ maintainers:
   - name: redpanda-data
     url: https://github.com/orgs/redpanda-data/people
 type: application
-version: 2.1.2
+version: 2.1.3
 appVersion: 22.2.2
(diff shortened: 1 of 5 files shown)


vuldin (approved):
This looks great. The only non-blocking feedback is that we need to document how to setup SASL so that users know they must set both `auth.sasl.users.name` and `auth.sasl.users.password` in `values.yaml` for it to work.

We could also include how they can verify their license:
```
kubectl exec -it -n redpanda-ns redpanda-0 -c redpanda -- rpk cluster license info
```

PR #163, opened 2022-09-14
do not add an empty nodeport service

### LINKED ISSUE #159: How do you disable external access?
I see numerous PRs and support for explicitly enabling external access.  In my case, I explicitly do NOT want this.

Setting `external.enabled: false` does not globally disable external configuration.  What is the proper values configuration for disabling external access? If I disable the above, the `external-nodeport-enabled` template still iterates over the listeners.  I don't think it should be doing this? If it should still be doing this, what is the proper listener configuration to disable external configuration?

joejulian:
When there are no external services enabled, do not add an empty nodeport service. Fixes #159


vuldin (approved):
LGTM

Issue #138, opened 2022-08-10
Add helm values json schema


joejulian:
It's easy for a user to fail to provide valid values data:

```bash
$ helm install redpanda redpanda --set auth.sasl.enabled=true --set auth.sasl.users.password=password
```
```text
coalesce.go:223: warning: destination for redpanda.auth.sasl.users is a table. Ignoring non-table value ([map[name:admin]])
Error: INSTALLATION FAILED: template: redpanda/templates/statefulset.yaml:52:28: executing "redpanda/templates/statefulset.yaml" at <include (print $.Template.BasePath "/configmap.yaml") .>: error calling include: template: redpanda/templates/configmap.yaml:38:24: executing "redpanda/templates/configmap.yaml" at <first .>: error calling first: Cannot find first on type map
```
We should provide a json schema: https://helm.sh/docs/topics/charts/#schema-files

PR #164, opened 2022-09-15
Add some schema validation

joejulian:
This adds schema validation for the image and auth.sasl values. See #138


dotnwat:
@joejulian i removed the `@dev` team from review. i don't think that works without making some adjustments to the assignment policy (e.g. it is assigning me to review this).

dotnwat (approved):

Issue #73, opened 2022-06-13
Create users listed in `auth.sasl.users`


vuldin:
```yaml
# Authentication
auth:
  #
  # SASL configuration
  sasl:
    enabled: false
    # user list
    # TODO create user at startup: https://github.com/redpanda-data/helm-charts/issues/73
    users:
    - name: admin
      # If password isn't given, then the secretName must point to an already existing secret
      # Work in progress! secretName is not yet used (planned for v2 release)
      # passwordSecret: adminPassword
```
Expand details of each entry in `users` list, possibly pointing to a secret for the user password.

vuldin:
The users list has been expanded to include password or passwordSecret details:
![image](https://user-images.githubusercontent.com/310946/174793289-7e7d27a8-2bd2-48a5-ba54-ee256952d97b.png)


This way the user can define their user, and the helm chart will handle creating this user so that it can be used with SASL to access Redpanda services.

vuldin:
Redapanda 22.2 will allow creating users on first start by setting an environment variable on first, relevant PR: https://github.com/redpanda-data/redpanda/pull/5231

joejulian:
The environment variable only allows setting one user. Since the values has a list, this should be a post-install,post-upgrade job.

Do we have to manage removing users that are not in the list?

vuldin:
I think this is closed by #162 ?

> Do we have to manage removing users that are not in the list?

That could be a separate ticket prioritized by how much (if any) feedback we get.

joejulian:
I don't think this is done. We should still set the first user using the environment variable. As it is now, I think there's a moment where the api is exposed without sasl passwords being set.

joejulian:
https://github.com/redpanda-data/helm-charts/pull/168 should be sufficient to close this

PR #168, opened 2022-09-27
Bootstrap admin user with RP_BOOTSTRAP_USER

mcfearsome:
Removes the post install job in favor of configuring the `RP_BOOTSTRAP_USER` env var.

Some other potential QoL changes while I was in there:

- "common-labels" named template
- configmap to make post upgrade script more readable


.github/workflows/lint-test.yml
```diff
@@ -24,17 +24,17 @@ on:
       - "**"
 jobs:
   lint-test:
-    runs-on: ubuntu-20.04
```
joejulian, line 27: This should be pinned. It should be pinned to a sha, in fact, to ensure reproducibility.

.github/workflows/lint-test.yml
```diff
@@ -24,17 +24,17 @@ on:
...
+    runs-on: ubuntu-latest
     steps:
       - name: Checkout
         uses: actions/checkout@v2
         with:
           fetch-depth: 0
 
       - name: Set up Helm
-        uses: azure/setup-helm@v1
```
joejulian, line 35: Use digest instead of tag

redpanda/templates/_helpers.tpl
```diff
@@ -387,4 +387,19 @@ IP is required for the advertised address.
...
+{{- define "common-labels" -}}
+helm.sh/chart: {{ template "redpanda.chart" . }}
+app.kubernetes.io/name: {{ template "redpanda.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name | quote }}
+app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
+app.kubernetes.io/component: {{ template "redpanda.name" . }}
+{{- with .Values.commonLabels }}
+{{- toYaml . }}
+{{- end }}
+{{- end -}}
```
joejulian, line 401: :+1:

redpanda/templates/post-upgrade-configmap.yaml
```diff
@@ -0,0 +1,11 @@
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "redpanda.fullname" . }}-post-upgrade
+  namespace: {{ .Release.Namespace | quote }}
+  labels:
+{{ include "common-labels" . | nindent 4 }}
+data:
+  post-upgrade.sh: |-
+    {{ .Files.Get "post-upgrade.sh" | nindent 4 }}
```
joejulian, line 11: Why use an extra configmap for this instead of using the file as the args in the job?

redpanda/templates/post-upgrade.yaml
```diff
@@ -67,6 +65,10 @@ spec:
         - name: {{ template "redpanda.fullname" . }}
           configMap:
             name: {{ template "redpanda.fullname" . }}
+        - name: post-upgrade-script
+          configMap:
+            name: {{ include "redpanda.fullname" . }}-post-upgrade
+            defaultMode: 0777
```
joejulian, line 71:
```suggestion
            defaultMode: 0544
```

redpanda/values.yaml
```diff
@@ -41,20 +41,16 @@ image:
...
-        # If password isn't given, then the secretName must point to an already existing secret
-        # secretName: adminPassword
+  enabled: false
+  # Set `bootstrap: {}` to enable sasl but not create the `RP_BOOTSTRAP_USER` envvar
+  bootstrap:
+    user: admin
+    password: ""
+    secretName: ""
+  additionalSuperUsers: []
+  # - name: user
```
joejulian, line 53: This can't be moved for backwards compatibility.

It's also under sasl because future auth options are coming soon.

redpanda/values.yaml
```diff
@@ -41,20 +41,16 @@ image:
...
-        # password: password
-        # If password isn't given, then the secretName must point to an already existing secret
-        # secretName: adminPassword
+  enabled: false
+  # Set `bootstrap: {}` to enable sasl but not create the `RP_BOOTSTRAP_USER` envvar
+  bootstrap:
+    user: admin
+    password: ""
+    secretName: ""
+  additionalSuperUsers: []
```
joejulian, line 52: This could remain under `users` (or `superusers` with a helper that uses `users` for backward compatibility). The creation of the bootstrap secret could be done by taking the `| first` of the list.

How does this set the passwords for the additional superusers?

joejulian (changes requested):
There are some great changes in this, thanks!

redpanda/templates/post-upgrade-configmap.yaml
mcfearsome: I find this way of doing it much more readable and there are the added benefits of normal editor niceties if/when it needs to be touched and easier addition of temporary debugging changes should they be necessary. Never much liked showing shell into yaml.

.github/workflows/lint-test.yml
mcfearsome: I was running the actions locally with act and this was mainly just to get it running at all. Easy change and you are of course right

redpanda/values.yaml
mcfearsome: Makes sense

redpanda/values.yaml
mcfearsome: I wanted to split it out for a scenario where someone didn't want to use the bootstrap env var in favor of some external process. They could set the bootstrap key to an empty object and use the additionalUsers key to mark the users as superusers in the config. My preference is to make the values more explicit rather than do list operations in helm but it's not a huge deal to revert the change.

I actually removed my additions for seeding passwords for all users as I wasn't sure if that was wanted

redpanda/templates/post-upgrade-configmap.yaml
joejulian: I do really like using .File for this, but wouldn't it be just as readable to:
```
kind: Job
spec:
  template:
    spec:
      containers:
        - command: ["/bin/sh", "-c"]
          args:
            - >
              {{ .Files.Get "post-upgrade.sh" | nindent 14 }}
```

redpanda/values.yaml
joejulian: Our desire is to make it as easy as possible for helm/kubernetes novices. The Helm chart should do things in a prescriptive way that helps with best practices. Adding superusers through values is simple and easy. As a list, it's intuitive. For the more advanced users, we want to add the ability to pull passwords from a secret+key that would need to be pre-populated.

redpanda/templates/post-upgrade-configmap.yaml
mcfearsome: Honestly, that never even crossed my mind! I'm blaming too much time in Kustomize, although that brings up the point that it's easier to patch a configmap should someone want to.

joejulian:
Do you mind if I make changes?

mcfearsome:
> Do you mind if I make changes?

Not at all! Apologies, I'm sure you familiar with devs getting distracted! 😂 I was going to try to find some time this week but be my guest!

.github/workflows/lint-test.yml
joejulian: omg, that was dumb. This isn't a docker image. :headdesk:

vuldin:
closes #73

joejulian:
I've not abandoned this. Just a little busy this week.

carlreid (commented):
I think the `auth.bootstrap` related values have not been updated in the `Values.yaml`. The `password` and `secretName`.

joejulian:
The user creation is going through a complete revamp. I'm going to close this in favor of that work. Thank you for your contribution, though. It truly is appreciated. This PR exposed the need for the revamp so it absolutely did provide value.

PR #176, opened 2022-10-11
Add missing namespace to charts template

### LINKED ISSUE #95: Add namespace to secrets and certs generated by helm chart
Right now the objects generated by the `certs.yaml` file don't have a namespace. It doesn't seem to be an issue, but it seems weird these objects would not need to be part of the same namespace as everything else.

joejulian:
Fixes #95


dotnwat (approved):

vuldin (approved):
LGTM

PR #180, opened 2022-10-18
increase default persistent volume size

### LINKED ISSUE #179: Change the storage default size to exceed the default storage_min_free_bytes
`storage_min_free_bytes` defaults to 5GiB but we're only creating PVCs at 3GiB. Change the default to something that can be used in production.

joejulian:
The default volume size of 3Gi was less than the default storage_min_free_bytes Redpanda configuration. This increases the default storage size to a much more reasonable 20GiB and calculates the minimum free bytes as 5% of the storage capacity.

Added a test to ensure topics can be created, produced to, and consumed.

Fixes #179

### DIFF
--- a/redpanda/Chart.yaml
+++ b/redpanda/Chart.yaml
@@ -20,7 +20,7 @@ maintainers:
   - name: redpanda-data
     url: https://github.com/orgs/redpanda-data/people
 type: application
-version: 2.2.1
+version: 2.2.2
 appVersion: v22.2.6
 icon: https://images.ctfassets.net/paqvtpyf8rwu/3cYHw5UzhXCbKuR24GDFGO/73fb682e6157d11c10d5b2b5da1d5af0/skate-stand-panda.svg
 sources:
--- a/redpanda/templates/_helpers.tpl
+++ b/redpanda/templates/_helpers.tpl
@@ -351,28 +351,54 @@ IP is required for the advertised address.
 {{- end -}}
 
 {{- define "rpk-flags" -}}
-  {{- $command := list -}}
-  {{- $command = concat $command (list "--api-urls" (include "api-urls" . )) -}}
+  {{- $admin := list -}}
+  {{- $admin = concat $admin (list "--api-urls" (include "api-urls" . )) -}}
   {{- if (include "admin-internal-tls-enabled" . | fromJson).bool -}}
-    {{- $command = concat $command (list
+    {{- $admin = concat $admin (list
       "--admin-api-tls-enabled"
       "--admin-api-tls-truststore"
       (printf "/etc/tls/certs/%s/ca.crt" .Values.listeners.admin.tls.cert))
     -}}
   {{- end -}}
+  {{- $kafka := list -}}
   {{- if (include "kafka-internal-tls-enabled" . | fromJson).bool -}}
-    {{- $command = concat $command (list
+    {{- $kafka = concat $kafka (list
       "--tls-enabled"
       "--tls-truststore"
       (printf "/etc/tls/certs/%s/ca.crt" .Values.listeners.kafka.tls.cert))
     -}}
   {{- end -}}
+  {{- $sasl := list -}}
   {{- if (include "sasl-enabled" . | fromJson).bool -}}
-    {{- $command = concat $command (list
+    {{- $sasl = concat $sasl (list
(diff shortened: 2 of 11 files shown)


redpanda/templates/configmap.yaml
```diff
@@ -67,10 +70,13 @@ data:
...
   {{- with (dig "cluster" dict .Values.config) }}
-    {{- toYaml . | nindent 6 }}
+      {{- toYaml . | nindent 6 }}
   {{- end }}
   {{- with (dig "tunable" dict .Values.config) }}
-    {{- toYaml . | nindent 6 }}
+      {{- toYaml . | nindent 6 }}
+  {{- end }}
+  {{- if not (hasKey "storage_min_free_bytes" .Values.config.cluster) }}
+      storage_min_free_bytes: {{- include "storage-min-free-bytes" . }}
```
jcsp, line 79: Bear in mind that when setting this in redpanda.yaml it'lll only take effect the first time you start up (this is the legacy upgrade path for importing pre-22.1 style configs)

redpanda/templates/_helpers.tpl
```diff
@@ -404,4 +375,12 @@ IP is required for the advertised address.
     -}}
   {{- end -}}
 {{ $command | join " " }}
+{{- end -}}
+
+{{- define "storage-min-free-bytes" -}}
+{{- if dig "enabled" false .Values.storage.persistentVolume -}}
+  {{- mulf (include "SI-to-bytes" .Values.storage.persistentVolume.size) 0.1 | int64 -}}
```
jcsp, line 382: Maybe put an upper bound on it?  Reserving 500MB on a 5GB drive is probably sensible, but reserving 1.6TB of space on a 16TB drive is wasteful

jcsp: For context: the level of slack needed is more about bandwidth than capacity: we need to notice the issue and stop writes, in time to avoid actually filling the drive.  So on a system doing 1GB/s of writes and checking its disk state every 10 seconds (this should really be more frequent but that's what it is today), we need ~10GB of safety room.

Probably not realistic to calculate this dynamically, but that's the motivation for suggesting to used a <10% number on large drives.

redpanda/templates/configmap.yaml
joejulian: It's undocumented, so how would one maintain this after the initial configuration?

redpanda/templates/_helpers.tpl
joejulian: The default being 5 gig made me guess conservatively. The user can override that setting using the `config.cluster.storage_min_free_bytes` value. Would you suggest a more logarithmic calculation for this or do you think setting a conservative default and allowing users to override that be sufficient?

redpanda/templates/configmap.yaml
jcsp: Cluster vs. node configuration: https://docs.redpanda.com/docs/cluster-administration/configuration/

redpanda.yml is for node config (the file is authoritative).  For cluster configuration, redpanda's internal store is authoritative.  There is a one-time read of cluster configuration properties from redpanda.yaml to accomodate upgrades from versions <22.1 where redpanda.yaml was for everything and there was no concept of cluster-wide properties.

Helm and ansible need updating as/when we have people available to do it -- operator already speaks the cluster config API (see calls to `PatchClusterConfig` for a thread to pull on).

I'm not suggesting you implement support for cluster configuration in this PR, just flagging it as a caveat for anywhere that the redpanda.yaml upgrade path is used in helm/ansible.

joejulian: Added the rpk command to the post-upgrade job.

vuldin (approved):
LGTM

redpanda/Chart.yaml
```diff
@@ -20,7 +20,7 @@ maintainers:
   - name: redpanda-data
     url: https://github.com/orgs/redpanda-data/people
 type: application
-version: 2.2.1
+version: 2.2.2
```
jcsp, line 23: nit: we're at 22.2.6 if you're updating this

jcsp (approved):

redpanda/Chart.yaml
joejulian: This is the chart version. See appVersion for the version of Redpanda.

PR #182, opened 2022-10-19
Change post-install-upgrade-job.yaml and post-upgrade.yaml job names to avoid `RepeatedResourceWarning`

missingcharacter:
When applying this helm chart the name of these jobs is repeated causing a warning.

This change would allow for both jobs to be created in case both are needed


joejulian:
Need to bump the chart version but otherwise lgtm

missingcharacter:
@joejulian give me a moment

joejulian (approved):

PR #184, opened 2022-10-21
#183 Fix reference scope

### LINKED ISSUE #183: Bug in `certs.yaml`: `.Release.Namespace` not available in `range` scope
Version 2.2.1 cannot be used with `tls.enabled=true`, because line 13 of `redpanda/templates/certs.yaml` refers to `.Release.Namespace` instead of to `$ns`.

This causes:

>Error: template: redpanda/templates/certs.yaml:13:24: executing "redpanda/templates/certs.yaml" at <.Release.Namespace>: nil pointer evaluating interface {}.Namespace

vovtz:
Fixes #183

### DIFF
--- a/redpanda/Chart.yaml
+++ b/redpanda/Chart.yaml
@@ -20,7 +20,7 @@ maintainers:
   - name: redpanda-data
     url: https://github.com/orgs/redpanda-data/people
 type: application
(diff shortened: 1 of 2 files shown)


joejulian (approved):
lgtm, thanks!

PR #188, opened 2022-10-24
#187 Check `Capabilities` to render `PodDisruptionBudget` manifest

### LINKED ISSUE #187: `PodDisruptionBudget` resource should be optional - it now prevents deployment on K8s 1.20 and below
The template for the `PodDisruptionBudget` should only render if its resource type is available on the cluster.

vovtz:
Fixes #187

### DIFF
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,13 +23,9 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 2.3.8
+version: 2.3.9
 # The app version is the default version of Redpanda to install.
 appVersion: v22.3.3
-# kubeVersion must be suffixed with "-0" to be able to match cloud providers
-# kubernetes versions like "v1.23.8-gke.1900". Their suffix is interpreted as a
-# pre-release. Our "-0" allows pre-releases to be matched.
-kubeVersion: "^1.21.0-0"
 icon: https://images.ctfassets.net/paqvtpyf8rwu/3cYHw5UzhXCbKuR24GDFGO/73fb682e6157d11c10d5b2b5da1d5af0/skate-stand-panda.svg
 sources:
   - https://github.com/redpanda-data/helm-charts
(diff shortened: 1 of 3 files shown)


joejulian:
The PDB is critical to prevent outages, split brain, and data loss. I think we should just prevent the chart from being used on k8s < 1.21 (which was merged before I saw this issue/PR). What do you think?

vovtz:
For our current use case there is no HA requirement, we plan to use _Redpanda_ only to create ephemeral one-node ‘clusters’ that are used in component testing. In future we might decide to use _Redpanda_ in production too, but for now we use _Kafka_.

(BTW, we obviously shouldn’t run K8s versions that are EOL anyway - we urgently requested to upgrade.)

redpanda/templates/poddisruptionbudget.yaml
```diff
@@ -35,3 +36,4 @@ spec:
       app.kubernetes.io/name: {{ template "redpanda.name" . }}
       app.kubernetes.io/instance: {{ .Release.Name | quote }}
   maxUnavailable: {{ .Values.statefulset.budget.maxUnavailable | int64 }}
+{{- end }}
```
joejulian, line 39: > For our current use case there is no HA requirement, we plan to use _Redpanda_ only to create ephemeral one-node ‘clusters’ that are used in component testing.

In that case, we should probably add a value under "below here be dragons" with a big warning about not using in production, then use that new value in an else, here.

something like:

```
yesyesyesno:
  pod_disruption_budget:
    enabled: true
```    

```suggestion
{{- else }}
  {{- if .Values.yesyesyesno.pod_disruption_budget.enabled }}
  fail "You must use Kubernetes 1.21+ for a production cluster. If you're not using this in production and you know why doing this is a bad idea, you can set yesyesyesno.pod_disruption_budget.enabled=false to continue."
  {{- end }
{{- end }}
```

(The parent key is not what I suggest, it's just a humorous placeholder)

joejulian (changes requested):

redpanda/templates/poddisruptionbudget.yaml
vovtz: I committed largely what you propose above, including an attempt for a more serious key name 😄

joejulian (approved):
Thank you. This seems like a good compromise.

joejulian:
rebased

joejulian:
Will rebase and merge after #202

PR #195, opened 2022-11-04
Support redpanda-nightly builds

vuldin:
Modify `semverCompare` usage to take into account repository. If `redpanda` repo is not used, then assume the latest version.

### DIFF
--- a/redpanda/templates/configmap.yaml
+++ b/redpanda/templates/configmap.yaml
@@ -37,7 +37,7 @@ metadata:
     {{- toYaml . | nindent 4 }}
 {{- end }}
 data:
-{{- if (include "redpanda.semver" . | semverCompare ">=22.1.1") }}
+{{- if (include "redpanda-atleast-22-1-1" . | fromJson).bool }}
   bootstrap.yaml: |
     enable_sasl: {{ dig "sasl" "enabled" false .Values.auth }}
   {{- if $users }}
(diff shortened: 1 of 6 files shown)


redpanda/templates/configmap.yaml
```diff
@@ -37,7 +37,7 @@ metadata:
     {{- toYaml . | nindent 4 }}
 {{- end }}
 data:
-{{- if (include "redpanda.semver" . | semverCompare ">=22.1.1") }}
+{{- if or (eq .Values.image.repository "vectorized/redpanda-nightly") (include "redpanda.semver" . | semverCompare ">=22.1.1") }}
```
joejulian, line 40: Let's go the other direction and only check it if it's `vectorized/redpanda`. That would allow developers to host their own images for testing as well.

redpanda/templates/configmap.yaml
```diff
@@ -64,7 +64,7 @@ data:
   {{- end }}
 {{- end }}
     redpanda:
-{{- if not (include "redpanda.semver" . | semverCompare ">=22.1.1") }}
+{{- if (include "redpanda-atleast-22-1-1" . | fromJson).bool }}
```
vuldin, line 67: @joejulian This is the line that causes this commit to fail linting with the following error:

```
[ERROR] templates/: template: redpanda/templates/statefulset.yaml:55:28: executing "redpanda/templates/statefulset.yaml" at <include (print $.Template.BasePath "/configmap.yaml") .>: error calling include: template: redpanda/templates/configmap.yaml:78:21: executing "redpanda/templates/configmap.yaml" at <"storage_min_free_bytes">: can't handle "storage_min_free_bytes" for arg of type map[string]interface {}
```
I think the error doesn't accurately point to the issue. If this line is reverted then the linting passes. But this line is identical to line 40 above which passes. Do you have an idea as to why?

joejulian (approved):

PR #196, opened 2022-11-09
tweak bootstrapping in v22.3 and above

andrwng:
Starting with v22.3, Redpanda can automatically assign node IDs to empty nodes as they join the cluster, and there is no longer a requirement to configure a single root node to start the cluster.

This commit updates the statefulset to use seeds-driven bootstrap, where all seed servers are expected to be configured with an identical list of seed servers. Once all seed servers are up they verify with each other that they all agree on the seeds, and collectively start a cluster with all seed servers as the initial set of nodes in the controller Raft group.

### DIFF
--- a/redpanda/Chart.yaml
+++ b/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 2.2.8
+version: 2.3.0
 # The app version is the default version of Redpanda to install.
 appVersion: v22.2.7
 # kubeVersion must be suffixed with "-0" to be able to match cloud providers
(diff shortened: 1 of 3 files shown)


andrwng:
For posterity, @vuldin helped test this with the `redpanda-nightly` repo on EKS and confirmed that the redpanda.yaml had the `empty_seed_starts_cluster` set to false, and there was no node ID.

redpanda/Chart.yaml
```diff
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 2.2.8
+version: 2.2.9
```
vuldin, line 26: From semver perspective, it feels like this update should move `2.2.8` to `2.3.0`. But I assume we are trying to keep the major and minor numbers in sync with Redpanda, so we just increment patch.

Maybe we should come up with a plan and include it in some README. Incrementing patch on any chart change that doesn't impact the default Redpanda version installed would be one way to tell at a glance if the chart should work with a given Redpanda version. But it breaks semver, which in itself is useful.

vuldin (approved):
Thanks for the PR! I tested on EKS verifying `node_id` and `seed_servers` were properly configured with the latest build from `redpanda-nightly`.

redpanda/Chart.yaml
vuldin: I'm leaning towards following semver (and making this version `2.3.0`).

andrwng: Yeah I could go either way on this. TBH this does probably skew towards "new feature" rather than "bug fix"

PR #201, opened 2022-11-15
Make SCRAM-SHA-512 default

vuldin:
This PR switches from SCRAM-SHA-256 to SCRAM-SHA-512. I considered making this a variable, but I don't know of any good reasons why it needs to be (please post if you can think of any).


vuldin:
Would there be some legacy Kafka clients that only support SCRAM-SHA-256? If the answer is not "definitely not", then we may want to make this an option.

joejulian (approved):
lgtm. Need to merge #209 first.

joejulian:
Merge after #210

PR #205, opened 2022-11-16
Publish changes merged to main

joejulian:
- Moved the `redpanda` chart under `charts` to fit the tooling expectation and in preparation for adding more charts.
- Reorganized the actions and publish charts to github pages using helm/chart-releaser


vuldin (approved):
LGTM. I didn't test the new chart releaser action (which is great to see added), but I don't think testing that out needs to be part of the review.

joejulian:
I tested it on my branch.

PR #208, opened 2022-11-17
Fix ci values

joejulian:
CI was not running the full set of values files as they were not named correctly for ct. This change renames them and fixes a couple values files, adding the password for the user.


vuldin (approved):
LGTM

PR #215, opened 2022-11-24
Added missing { to statefull set for commonLabels

Scroul:
Common Labels cause the statefulset to be invalid


joejulian (approved):
Thanks @Scroul!

PR #216, opened 2022-11-24
Fix kafka produce consume test

RafalKorepta:
The error in the produce consume test was shwallowed by rpk.
```
unable to create topics [produce.consume.test]: unable to dial: dial tcp 127.0.0.1:9092: connect: connection refused
/bin/bash: line 1: exit: too many arguments
```

`REDPANDA_BROKERS` was added to the pod template spec as rpk in tls test need to have fully qualified name as it is defined in self signed certificate.

As the definition of the produce consume test changed from pod to job the topic needs to be uqnique per job run, that is why pod name is used to randomise the topic name.

Produce and consume test can now failed up to 2 times, before it is considered as failed.

### DIFF
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 2.3.4
+version: 2.3.5
 # The app version is the default version of Redpanda to install.
 appVersion: v22.3.3
 # kubeVersion must be suffixed with "-0" to be able to match cloud providers
--- a/charts/redpanda/templates/tests/test-kafka-produce-consume.yaml
+++ b/charts/redpanda/templates/tests/test-kafka-produce-consume.yaml
@@ -14,8 +14,8 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */}}
-apiVersion: v1
-kind: Pod
+apiVersion: batch/v1
+kind: Job
 metadata:
   name: {{ include "redpanda.fullname" . }}-test-kafka-produce-consume
   namespace: {{ .Release.Namespace | quote }}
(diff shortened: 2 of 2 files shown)


charts/redpanda/templates/tests/test-api-status.yaml
```diff
@@ -31,7 +31,7 @@ metadata:
   {{- end }}
   annotations:
     "helm.sh/hook": test
-    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
```
joejulian, line 34: Doesn't removing `before-hook-creation` cause failures on a re-test if the test has failed before?

RafalKorepta: Your are right. I have a problem running `helm test` command as it always deletes every pod and every job. I couldn't debug failed run easily. 

I'm using 
```
helm version
version.BuildInfo{Version:"v3.10.2", GitCommit:"50f003e5ee8704ec937a756c646870227d7c8b58", GitTreeState:"clean", GoVersion:"go1.19.3"}
```

The `helm test` command doesn't have any useful flag to prevent from deletion the test pods/jobs:
```
helm test redpanda --help

The test command runs the tests for a release.

The argument this command takes is the name of a deployed release.
The tests to be run are defined in the chart that was installed.

Usage:
  helm test [RELEASE] [flags]

Flags:
      --filter strings     specify tests by attribute (currently "name") using attribute=value syntax or '!attribute=value' to exclude a test (can specify multiple or separate values with commas: name=test1,name=test2)
  -h, --help               help for test
      --logs               dump the logs from test pods (this runs after all tests are complete, but before any cleanup)
      --timeout duration   time to wait for any individual Kubernetes operation (like Jobs for hooks) (default 5m0s)

Global Flags:
      --burst-limit int                 client-side default throttling limit (default 100)
      --debug                           enable verbose output
      --kube-apiserver string           the address and the port for the Kubernetes API server
      --kube-as-group stringArray       group to impersonate for the operation, this flag can be repeated to specify multiple groups.
      --kube-as-user string             username to impersonate for the operation
      --kube-ca-file string             the certificate authority file for the Kubernetes API server connection
      --kube-context string             name of the kubeconfig context to use
      --kube-insecure-skip-tls-verify   if true, the Kubernetes API server's certificate will not be checked for validity. This will make your HTTPS connections insecure
      --kube-tls-server-name string     server name to use for Kubernetes API server certificate validation. If it is not provided, the hostname used to contact the server is used
      --kube-token string               bearer token used for authentication
      --kubeconfig string               path to the kubeconfig file
  -n, --namespace string                namespace scope for this request
      --registry-config string          path to the registry config file (default "/Users/korepta/Library/Preferences/helm/registry/config.json")
      --repository-cache string         path to the file containing cached repository indexes (default "/Users/korepta/Library/Caches/helm/repository")
      --repository-config string        path to the file containing repository names and URLs (default "/Users/korepta/Library/Preferences/helm/repositories.yaml")
```

joejulian: I wonder if it's the Job ttl that's deleting when successful? Maybe try adding a [ttlSecondsAfterFinished](https://kubernetes.io/docs/concepts/workloads/controllers/job/#ttl-mechanism-for-finished-jobs)?

RafalKorepta: I have problem with all resources (pods and my new job). When I'm debugging I would like to left resource after they finish. Even if they succeed.

joejulian: Right. Setting the ttlSecondsAfterFinish tells kube-controller to not clean it up for the specified length of time.

RafalKorepta: Thank, I reverted the `before-hook-creation` removal and added `ttlSecondsAfterFinish`

joejulian (approved):

PR #219, opened 2022-12-01
redpanda: add rack awareness feature

### LINKED ISSUE #86: Enable rack awareness
Enabled rack awareness when installed across multiple availability zones.

https://github.com/redpanda-data/helm-charts/blob/29049a7f1da7e8d97ef8563fcd4a44057d004717/redpanda/values.yaml#L355

See https://docs.redpanda.com/docs/data-management/rack-awareness/

joejulian:
Set redpanda.rack to the value of the Kubernetes Node's `topology.kubernetes.io/zone` annotation.

Fixes #86

### DIFF
--- a/charts/redpanda/values.yaml
+++ b/charts/redpanda/values.yaml
@@ -602,7 +602,6 @@ config:
     # api_doc_dir: /usr/share/redpanda/proxy-api-doc               # API doc directory
     # coproc_supervisor_server: 127.0.0.1:43189                    # IpAddress and port for supervisor service
     # dashboard_dir: None                                          # serve http dashboard on / url
-    # rack: None                                                   # Rack identifier
     # developer_mode: optional                                     # Skips most of the checks performed at startup
 
   # Invalid properties
--- a/charts/redpanda/values.yaml
+++ b/charts/redpanda/values.yaml
@@ -19,7 +19,7 @@
 #  https://helm.sh/docs/chart_template_guide/values_files/
 
 #
-# >>> This chart requires Helm version 3.5.0 or greater <<<
+# >>> This chart requires Helm version 3.6.0 or greater <<<
 #
 
 # Common parameters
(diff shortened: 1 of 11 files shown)


charts/redpanda/values.yaml
```diff
@@ -344,6 +354,12 @@ serviceAccount:
   # The name of the service account to use.
   # If not set and create is true, a name is generated using the fullname template
   name: ""
+# Role Based Access Control
+rbac:
+  # enable this for features that need extra priveleges
+  enabled: false
```
vuldin, line 360: Should this be enabled by default? I feel that it should as it is required to enable other features out of the box. For instance, we enable external access out of the box, but if we want it to work then they would need to flip this flag (eventually). The same applies to rack awareness, in that they would need to enable rack awareness and also rbac in order to get just rack awareness.

Another option may be to do a check that doesn't allow users to enable either rack awareness or external access without also enabling rbac. I this would work to ensure clusters work as expected, but at the same time making rbac be enabled automatically when these other features are enabled would be a better default.

vuldin (commented):
One comment around `rbac.enabled` existing and/or it's default being false. I think it would be better to leave this out entirely and then enable it automatically when this (and future features) are enabled by the user.

charts/redpanda/values.yaml
joejulian: The principal of least privilege suggests that of a privilege is not needed it should not be added. Unless rack awareness is enabled, there's no need for the container to be able to get the Node.

vuldin: That's true for rack awareness. I guess another approach would be to enable rbac by default in [another incoming PR](https://github.com/redpanda-data/helm-charts/pull/224) that fixes external access by making use of this (given that external access is enabled by default.

joejulian:
> One comment around rbac.enabled existing and/or it's default being false. I think it would be better to leave this out entirely and then enable it automatically when this (and future features) are enabled by the user.

I made this choice for a couple reasons. One, to ensure that adding privilege is a conscious choice.

The other reason is to allow this chart to be used by unprivileged users. In an enterprise environment, organization units are given a namespace but have not cluster-level privileges. In this scenario, the ServiceAccount, ClusterRole, and ClusterRoleBinding will need to be created by the cluster administrator. The chart user will still be able to use the chart by using the SA assigned to them and not enabling the rbac components.

vuldin (approved):
LGTM

PR #220, opened 2022-12-01
Expand external access capabilities

vuldin:
- subdomain is customizable
- IP address can be used for kafka advertised listener
- new redpanda.external.nodeType, selecting either subdomain or ip
- new redpanda.external.nodes, listing node ips or subdomains
- remove setting kafka advertised listener via flag
- added setting kafka advertised listener from initContainer (via rpk)

Docs for how to use this feature are in a docs PR that I'll link here soon.


charts/redpanda/templates/statefulset.yaml
```diff
@@ -68,28 +68,58 @@ spec:
               mountPath: /var/lib/redpanda/data
         - name: {{ template "redpanda.name" . }}-configurator
           image: {{ .Values.image.repository}}:{{ template "redpanda.tag" . }}
+          # command: ["/bin/bash", "-c"]
```
joejulian, line 71: leftover comment

charts/redpanda/templates/statefulset.yaml
```diff
@@ -68,28 +68,58 @@ spec:
...
               {{- else }}
-              NODE_ID=${SERVICE_NAME##*-};
               rpk --config "$CONFIG" config set redpanda.node_id $NODE_ID;
               if [ "$NODE_ID" = "0" ]; then
                 rpk --config "$CONFIG" config set redpanda.seed_servers '[]' --format yaml;
               fi;
               {{- end }}
+              rpk redpanda --config "$CONFIG" config set redpanda.advertised_kafka_api[0].name internal ;
+              rpk redpanda --config "$CONFIG" config set redpanda.advertised_kafka_api[0].address {{ $advertiseAddress }} ;
+              rpk redpanda --config "$CONFIG" config set redpanda.advertised_kafka_api[0].port {{ .Values.listeners.kafka.port }} ;
```
joejulian, line 98: Why are we doing this for internal?

charts/redpanda/values.yaml
```diff
@@ -92,9 +92,23 @@ external:
...
-  # Default external access type (NodePort is the only option for now)
+  # Default external access type (only NodePort at the moment)
   type: NodePort
+  # Domain to be used for each node
   domain: local
+  #
+  # nodeType can be set to the following: subdomain ip
+  # with subdomain, each node will advertise an address with <external.nodes[replica index]>.<domain>
+  # with ip, each node will advertise only <external.nodes[replica index]> (and domain will be ignored)
+  nodeType: ip
```
joejulian, line 103: This changes the default behavior making it a breaking change.

charts/redpanda/values.yaml
```diff
@@ -92,9 +92,23 @@ external:
...
+  #
+  # The nodes list must have an entry for each statefulset replica
+  # By default, the statefulset is called redpanda and has 3 replicas
+  # If you want to access nodes by their IP: nodes: "18.224.215.250 18.118.163.26 3.142.153.29"
+  # Or if you want to access nodes via DNS: nodes: "apple bacon carrot"
+  # nodes: "apple bacon carrot"
+  nodes: "3.129.52.203 3.12.164.46 3.144.145.170"
```
joejulian, line 110: Can we use a term that's less ambiguous than "nodes"? I'm not sure we should be using routable addresses in the example.

How will a user know which ip addresses (or hostnames) to put here? Won't these be unpredictable?

Also, did you mean for this to be set to these addresses by default?

charts/redpanda/templates/statefulset.yaml
```diff
@@ -68,28 +68,58 @@ spec:
...
+                  {{- if $values.external.nodes -}}
+              NODE_INDEX=`expr $NODE_ID + 1`;
+              NODE_ADDRESS=`echo $EXTERNAL_NODES | cut -d ' ' -f $NODE_INDEX`;
+                    {{- if eq $values.external.nodeType "ip" -}}
+              rpk redpanda --config "$CONFIG" config set redpanda.advertised_kafka_api[{{ $listenerIndex }}].address $NODE_ADDRESS;
+                    {{- else -}}
+              rpk redpanda --config "$CONFIG" config set redpanda.advertised_kafka_api[{{ $listenerIndex }}].address $NODE_ADDRESS.{{ $values.external.domain }};
+                    {{- end -}}
+                  {{- else -}}
+                    {{- $advertiseKafkaHost = printf "$(SERVICE_NAME).%s" $values.external.domain -}}
```
carlreid, line 115: Could make use of the template value `redpanda.kafka.external.advertise.address`? Otherwise it could be removed from `_helpers.tpl` as I don't think it's used anywhere.

carlreid (commented):
I notice there's no changes related to certificates/TLS management in this PR. From what I see, the current certificate doesn't set up anything regarding the external domain. Seems to have some concept of `clusterDomain` but seems related to internal domain also.

Could be that a separate issue needs opening to be addressed later?

charts/redpanda/templates/statefulset.yaml
vuldin: We don't actually configure any `advertised_kafka_api` in the configmap, it was originally in this same file as a startup flag. I considered moving all this `advertised_kafka_api, but like seed servers this configuration depends on the node ID (and the configmap is identical across all nodes).

charts/redpanda/values.yaml
vuldin: This was an accident to keep in (it is now commented out by default). Without this or `node` the current default behavior occurs.

charts/redpanda/values.yaml
vuldin: Same as my `nodeType` comment above, this was mistakenly left in (I've removed it).

charts/redpanda/templates/statefulset.yaml
vuldin: Looking at this and a few other named templates, I see they are using environment variables to allow the container to replace the text with the actual value (`SERVICE_NAME` or `HOST_IP`): https://github.com/redpanda-data/helm-charts/blob/main/charts/redpanda/templates/_helpers.tpl#L115-L141

I think this is an issue because that means these named templates will only work within files that spin up containers, but we don't mention that or block their use anywhere else (in fact it probably isn't possible to do so). I know in the past I've mistakenly tried to use these named templates in ConfigMap or other places and found that the environment variable being referenced is just passed through without ever being associated with the expected value.

I think these named templates should be removed to avoid this confusion, or possibly handled differently so that the chance for confusion is much less. I'll create an issue around this and hopefully that can be handled in a separate PR.

vuldin: https://github.com/redpanda-data/helm-charts/issues/223

carlreid: Sounds like a good idea 😊 Thanks for the investigation work 👍

joejulian (approved):
lgtm. Need a chart version bump before merging. Might as well rebase, too.

PR #222, opened 2022-12-02
Service accounts

Scroul:
Not sure if you want this level of configuration for service accounts but I am running the chart in a service mesh so need good control over the service account. 

I can add/remove some parts if you feel that you only want a part of this.

With this keeping `serviceAccount` in values as it is now will put `serviceAccountName: default` so it will work as  before if you do not use any of the new yaml values.


joejulian:
Would you mind opening an issue to associate with this that describes the symptoms you encounter with using a service mesh without this change? It would be really helpful for our docs team to be able to understand when and how this feature is used.

Scroul:
I added #225 explaining how this can be used in a service mesh.

Issue #237, opened 2022-12-09
Disabling Pandaproxy also disables Schema Registry


carlreid:
I think a [misplaced `if` statement in the configmap](https://github.com/redpanda-data/helm-charts/blob/6fcfb2783c34e839a99d9456956e2c1f3a9ce65d/charts/redpanda/templates/configmap.yaml#L151-L152) has caused this.

I don't believe there is a dependency on Pandaproxy for the Schema Registry? Otherwise maybe the Values.yaml could be refactored to indicate that.

PR #238, opened 2022-12-09
Relocate `if` statement so `schemaRegistry` is still defined if pandaproxy is not enabled

carlreid:
This should hopefully close/fix https://github.com/redpanda-data/helm-charts/issues/237, assuming there is no dependency between schema registry and pandaproxy.

If there is, I guess close dismiss both the issue and this PR 😀


vuldin (approved):
LGTM, thanks!

carlreid:
@vuldin Thanks for review 😊 I just rebased + bumped version to `2.3.16` (was `2.3.15` when you reviewed) due to another PR being merged.

Issue #230, opened 2022-12-07
If user was already created, the post-install/upgrade job fails


carlreid:
The [`rpk acl user create` command](https://github.com/redpanda-data/helm-charts/blob/main/charts/redpanda/templates/post-install-upgrade-job.yaml#L71) will exit code `1` if the user exists (due to HTTP 500), then the whole job fails due to the [`set -e` switch](https://github.com/redpanda-data/helm-charts/blob/main/charts/redpanda/templates/post-install-upgrade-job.yaml#LL67C19-L67C19).

I'm not sure if it makes sense to fail the job if the user is already there? On install, maybe it could make sense, but you then need to assume that there's no lingering volume from a previous install.

In my case, I am using Redpanda as a dependency chart, with a Helm hook of my own which won't be executed due to the fail state reported by the [`post-install-upgrade-job.yaml`](https://github.com/redpanda-data/helm-charts/blob/main/charts/redpanda/templates/post-install-upgrade-job.yaml).

PR #239, opened 2022-12-09
Handle case when user already exists to avoid job failing

carlreid:
Should close/resolve issue https://github.com/redpanda-data/helm-charts/issues/230. Basically I just caught the output, so it can be checked against in the case of the user existing. If it isn't the string we expect, then echo out/exit with the original code. Bit of a bash noob, so maybe there is a more concise way of doing this 😁

I've tried this with creation of two users, and re-running after those two users are created.

First couple job runs when cluster not ready:
```text
Retrying POST for error: Post "https://kafka-0.afka.default.svc.cluster.local.:9644/v1/security/users": dial tcp: lookup kafka-0.kafka.default.svc.cluster.local. on 10.0.0.10:53: no such host
Retrying POST for error: Post "https://kafka-0.kafka.default.svc.cluster.local.:9644/v1/security/users": dial tcp: lookup kafka-0.kafka.default.svc.cluster.local. on 10.0.0.10:53: no such host
unable to create user "admin": Post "https://kafka-0.kafka.default.svc.cluster.local.:9644/v1/security/users": dial tcp: lookup kafka-0.kafka.default.svc.cluster.local. on 10.0.0.10:53: no such host
```

When cluster is ready:
```text
Created user admin.
Created user console.
```

When upgrading the chart:
```text
The admin user already exists, skipping creation.
The console user already exists, skipping creation.
```

### DIFF
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 2.3.14
+version: 2.3.15
 # The app version is the default version of Redpanda to install.
 appVersion: v22.3.5
 # kubeVersion must be suffixed with "-0" to be able to match cloud providers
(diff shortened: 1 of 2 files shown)


joejulian (approved):
LGTM, thanks!

PR #245, opened 2022-12-15
ensure the uid:gid and the fsgroup always match

joejulian:
If the redpanda container changes the uid:gid for the "redpanda" user, it will no longer match the fsGroup and the pod will no longer be able to access its data or configuration. This ensures the ids always match the file ownership.

### DIFF
--- a/charts/redpanda/templates/_helpers.tpl
+++ b/charts/redpanda/templates/_helpers.tpl
@@ -430,3 +430,15 @@ IP is required for the advertised address.
 {{- define "redpanda-atleast-22-3-0" -}}
 {{- toJson (dict "bool" (or (not (eq .Values.image.repository "vectorized/redpanda")) (include "redpanda.semver" . | semverCompare ">=22.3.0"))) -}}
 {{- end -}}
+
+# manage backward compatibility with renaming podSecurityContext to securityContext
+{{- define "pod-security-context" -}}
+fsGroup: {{ dig "podSecurityContext" "fsGroup" .Values.statefulset.securityContext.fsGroup .Values.statefulset }}
+{{- end -}}
+
+# for backward compatibility, force a default on releases that didn't
+# set the podSecurityContext.runAsUser before
+{{- define "container-security-context" -}}
+runAsUser: {{ dig "podSecurityContext" "runAsUser" .Values.statefulset.securityContext.runAsUser .Values.statefulset }}
+runAsGroup: {{ dig "podSecurityContext" "fsGroup" .Values.statefulset.securityContext.fsGroup .Values.statefulset }}
+{{- end -}}
--- a/charts/redpanda/templates/statefulset.yaml
+++ b/charts/redpanda/templates/statefulset.yaml
@@ -57,13 +57,14 @@ spec:
         {{- toYaml . | nindent 8 }}
 {{- end }}
     spec:
-      securityContext:
-        {{- toYaml .Values.statefulset.podSecurityContext | nindent 8 }}
+      securityContext: {{ include "pod-security-context" . | nindent 8 }}
       serviceAccountName: {{ include "redpanda.serviceAccountName" . }}
       initContainers:
         - name: set-datadir-ownership
           image: busybox:latest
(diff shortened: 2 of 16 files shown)


maowerner:
Hi. 
we stumbled over this issue and it is really nice to see the fixes are already coming :) We had a connected issue though: When setting the `user` and `group` to `101` the post-upgrade and post-install pods failed because `kubernetes PodSecurityPolicy set to runAsNonRoot, container has runAsNonRoot and image has non-numeric user (appuser), cannot verify user is non-root`. We could solve this by setting the securityContext there as well. 

Can you reproduce this? Would it make sense to set the `securityContext` in these pods as default?

joejulian:
Good catch. Thanks!

charts/redpanda/templates/_helpers.tpl
```diff
@@ -430,3 +430,15 @@ IP is required for the advertised address.
...
 {{- toJson (dict "bool" (or (not (eq .Values.image.repository "vectorized/redpanda")) (include "redpanda.semver" . | semverCompare ">=22.3.0"))) -}}
 {{- end -}}
+
+# manage backward compatibility with renaming podSecurityContext to securityContext
+{{- define "pod-security-context" -}}
+fsGroup: {{ dig "podSecurityContext" "fsGroup" .Values.statefulset.securityContext.fsGroup .Values.statefulset }}
+{{- end -}}
+
+# for backward compatibility, force a default on releases that didn't
+# set the podSecurityContext.runAsUser before
```
RafalKorepta, line 440:
```suggestion
# set the podSecurityContext.runAsUser nor podSecurityContext.fsGroup before
```

charts/redpanda/templates/_helpers.tpl
```diff
@@ -430,3 +430,15 @@ IP is required for the advertised address.
...
+{{- define "container-security-context" -}}
+runAsUser: {{ dig "podSecurityContext" "runAsUser" .Values.statefulset.securityContext.runAsUser .Values.statefulset }}
+runAsGroup: {{ dig "podSecurityContext" "fsGroup" .Values.statefulset.securityContext.fsGroup .Values.statefulset }}
```
RafalKorepta, line 443: Should this be:
```suggestion
runAsGroup: {{ dig "podSecurityContext" "runAsGroup" .Values.statefulset.securityContext.runAsGroup .Values.statefulset }}
```
?

Or at least
```suggestion
runAsGroup: {{ dig "podSecurityContext" "runAsGroup" .Values.statefulset.securityContext.fsGroup .Values.statefulset }}
```

charts/redpanda/templates/statefulset.yaml
```diff
@@ -140,13 +141,13 @@ spec:
                 rpk --config "$CONFIG" redpanda config set redpanda.rack "${RACK}"
                 {{- end }}
               {{- end }}
+          securityContext: {{ include "container-security-context" . | nindent 12 }}
```
RafalKorepta, line 144: NIT:
Weird that we have almost in all places `nindent 10` for container security content. Should we remove one level of indentation in this file for all containers?

RafalKorepta (approved):
LGTM only few NIT

charts/redpanda/templates/_helpers.tpl
joejulian: No, we want to ensure the runAsGroup matches the fsGroup. To do that we just don't expose the runAsGroup key in values.

charts/redpanda/templates/statefulset.yaml
joejulian: This is the container securityContext. The `nindent 10` ones are Pod securityContext.

joejulian: I find it just as weird that go doesn't indent lists but just about every yaml linter expects it. I think indented lists are more readable.

Since this list is already indented, let's decide this outside this PR.

charts/redpanda/templates/_helpers.tpl
joejulian: `fsGroup` has always been set. It was a required value since the beginning.

PR #254, opened 2022-12-24
Allow full control of advertised listeners

### LINKED ISSUE #256: Support LoadBalancer for connectivity

### LINKED ISSUE #258: external.addresses only picks up the first value
chart version: redpanda-2.4.0

Steps to reproduce:

1. Set `external.addresses` to some IPs (one for each broker).

    ```yaml
    external:
      addressType: ip
      addresses: "203.0.113.3 203.0.113.5 203.0.113.7"
    ```

1. Apply the file to the Helm deployment.

1. Check the `redpanda.yaml` for each broker.

    ```
    kubectl exec redpanda-0 -n redpanda -c redpanda -- cat /etc/redpanda/redpanda.yaml
    ```
    
    ```
    kubectl exec redpanda-1 -n redpanda -c redpanda -- cat /etc/redpanda/redpanda.yaml
    ```
    
    ```
    kubectl exec redpanda-2 -n redpanda -c redpanda -- cat /etc/redpanda/redpanda.yaml
    ```

They all use the same IP address in their`advertised_kafka_api` field.

```yaml
    advertised_kafka_api:
        - address: redpanda-2.redpanda.redpanda.svc.cluster.local.
          port: 9093
          name: internal
        - address: 203.0.113.3
          port: 31092
          name: default
```

The PandaProxy API is also advertising the FQDN of the headless Cluster IP:

```
    advertised_pandaproxy_api:
        - address: redpanda-2.redpanda.redpanda.svc.cluster.local.
          port: 8082
          name: internal
        - address: redpanda-2.redpanda.redpanda.svc.cluster.local.
          port: 30082
          name: default
```

vuldin:
This PR enables configuring advertised listeners without creating a service for external access. This makes it possible for helm chart deployments to work with a variety of external access methods, such as LoadBalancer service.

Fixes #258, fixes #256 

### Changes
1. made `external.type` optional. If undefined, then advertised listeners will be configured without creating a service
3. removed `$externalListener.nodePorts`, added `$externalListener.advertisedPorts`:
  a. when `external.type` is `NodePort`, the first element of the array is the port opened on each node
  b. when `external.type` is undefined, each element of the array is the port to be opened externally for accessing the pod where array index == statefulset replica index
  c. when advertisedPorts is undefined, `listeners.*.external.*.port` then `listeners.*.port` is used
4. removed container start command listener flags, add listeners other than RPC to statefulset initContainer
5. removed unused templates
6. switched `external.addresses` from string to array
7. removed `listeners.*.external.type` and assume `external.type` is used across all externally enabled listeners
8. fix `schema_registry_api` in configmap.yaml (was `schema_registry`)
9. update test-schema-registry-status to use same short hostname URL as other tests (rather than FQDN)
10. fixed issue after `NODE_ID` was renamed to `POD_ORDINAL`

### DIFF
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 2.5.1
+version: 2.5.2
 
 # The app version is the default version of Redpanda to install.
 appVersion: v22.3.10
--- a/charts/redpanda/templates/configmap.yaml
+++ b/charts/redpanda/templates/configmap.yaml
@@ -148,7 +148,7 @@ data:
 {{- end }}
 {{- if .Values.listeners.schemaRegistry.enabled }}
     schema_registry:
-      schema_registry:
+      schema_registry_api:
         - name: internal
           address: 0.0.0.0
           port: {{ .Values.listeners.schemaRegistry.port }}
--- a/charts/redpanda/templates/services.nodeport.yaml
+++ b/charts/redpanda/templates/services.nodeport.yaml
@@ -15,7 +15,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */}}
 {{- $values := .Values }}
-{{- if (include "external-nodeport-enabled" . | fromJson).bool }}
+{{- if (and .Values.external.enabled (eq .Values.external.type "NodePort")) }}
 ---
 apiVersion: v1
 kind: Service
@@ -38,42 +38,38 @@ spec:
   ports:
 {{- range $name, $listener := $values.listeners.admin.external }}
   {{- $enabled := dig "enabled" $values.external.enabled $listener }}
-  {{- $type := dig "type" $values.external.type $listener }}
-  {{- if and $enabled (eq $type "NodePort") }}
+  {{- if $enabled }}
     - name: admin-{{ $name }}
       protocol: TCP
       port: {{ $values.listeners.admin.port }}
-      nodePort: {{ $listener.nodePort }}
+      nodePort: {{ index $listener.advertisedPorts 0 }}
   {{- end }}
 {{- end }}
 {{- range $name, $listener := $values.listeners.kafka.external }}
   {{- $enabled := dig "enabled" $values.external.enabled $listener }}
-  {{- $type := dig "type" $values.external.type $listener }}
-  {{- if and $enabled (eq $type "NodePort") }}
+  {{- if $enabled }}
     - name: kafka-{{ $name }}
       protocol: TCP
       port: {{ $listener.port }}
-      nodePort: {{ $listener.nodePort }}
+      nodePort: {{ index $listener.advertisedPorts 0 }}
   {{- end }}
 {{- end }}
 {{- range $name, $listener := $values.listeners.http.external }}
   {{- $enabled := dig "enabled" $values.external.enabled $listener }}
-  {{- $type := dig "type" $values.external.type $listener }}
-  {{- if and $enabled (eq $type "NodePort") }}
+  {{- if $enabled }}
     - name: http-{{ $name }}
       protocol: TCP
       port: {{ $listener.port }}
-      nodePort: {{ $listener.nodePort }}
+      nodePort: {{ index $listener.advertisedPorts 0 }}
   {{- end }}
 {{- end }}
 {{- range $name, $listener := $values.listeners.schemaRegistry.external }}
   {{- $enabled := dig "enabled" $values.external.enabled $listener }}
-  {{- $type := dig "type" $values.external.type $listener }}
-  {{- if and $enabled (eq $type "NodePort") }}
+  {{- if $enabled }}
     - name: schema-{{ $name }}
       protocol: TCP
-      port: {{ $listener.port }}
-      nodePort: {{ $listener.nodePort }}
+      port: {{ $values.listeners.schemaRegistry.port }}
+      nodePort: {{ index $listener.advertisedPorts 0 }}
   {{- end }}
 {{- end }}
   selector:
--- a/charts/redpanda/templates/services.nodeport.yaml
+++ b/charts/redpanda/templates/services.nodeport.yaml
@@ -15,7 +15,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */}}
 {{- $values := .Values }}
-{{- if (and .Values.external.enabled (eq .Values.external.type "NodePort")) }}
+{{- if (include "external-nodeport-enabled" . | fromJson).bool }}
 ---
 apiVersion: v1
 kind: Service
@@ -42,7 +42,7 @@ spec:
     - name: admin-{{ $name }}
       protocol: TCP
       port: {{ $values.listeners.admin.port }}
-      nodePort: {{ index $listener.advertisedPorts 0 }}
+      nodePort: {{ dig "nodePort" (first (dig "advertisedPorts" (list $values.listeners.admin.port) $listener)) $listener }}
   {{- end }}
 {{- end }}
 {{- range $name, $listener := $values.listeners.kafka.external }}
@@ -51,7 +51,7 @@ spec:
     - name: kafka-{{ $name }}
       protocol: TCP
       port: {{ $listener.port }}
-      nodePort: {{ index $listener.advertisedPorts 0 }}
+      nodePort: {{ dig "nodePort" (first (dig "advertisedPorts" (list $values.listeners.kafka.port) $listener)) $listener }}
   {{- end }}
 {{- end }}
 {{- range $name, $listener := $values.listeners.http.external }}
@@ -60,7 +60,7 @@ spec:
     - name: http-{{ $name }}
       protocol: TCP
       port: {{ $listener.port }}
-      nodePort: {{ index $listener.advertisedPorts 0 }}
+      nodePort: {{ dig "nodePort" (first (dig "advertisedPorts" (list $values.listeners.http.port) $listener)) $listener }}
   {{- end }}
 {{- end }}
 {{- range $name, $listener := $values.listeners.schemaRegistry.external }}
@@ -69,7 +69,7 @@ spec:
     - name: schema-{{ $name }}
       protocol: TCP
       port: {{ $values.listeners.schemaRegistry.port }}
-      nodePort: {{ index $listener.advertisedPorts 0 }}
+      nodePort: {{ dig "nodePort" (first (dig "advertisedPorts" (list $values.listeners.schemaRegistry.port) $listener)) $listener }}
   {{- end }}
 {{- end }}
   selector:
--- a/charts/redpanda/templates/statefulset.yaml
+++ b/charts/redpanda/templates/statefulset.yaml
@@ -15,6 +15,41 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */}}
 
+
+{{- /*
+advertised-port returns either the only advertised port if only one is specified,
+or the port specified for this pod ordinal when there is a full list provided.
+
+This will return a string int or panic if there is more than one port provided,
+but not enough ports for the number of replicas requested.
+*/ -}}
+{{- define "advertised-port" -}}
+  {{- $port := dig "port" .listenerVals.port .externalVals -}}
+  {{- if .externalVals.advertisedPorts -}}
+    {{- if eq (len .externalVals.advertisedPorts) 1 -}}
+      {{- $port = mustFirst .externalVals.advertisedPorts -}}
+    {{- else -}}
+      {{- $port = index .externalVals.advertisedPorts .replicaIndex -}}
+    {{- end -}}
+  {{- end -}}
+  {{ $port }}
+{{- end -}}
+
+{{- /*
+advertised-host returns a json sring with the data neded for configuring the advertised listener
+*/ -}}
+{{- define "advertised-host" -}}
+  {{- $host := dict "name" .externalName "address" .externalAdvertiseAddress "port" .port -}}
+  {{- if .values.external.addresses -}}
+    {{- if .values.external.domain -}}
+      {{- $host = dict "name" .externalName "address" (printf "%s.%s" (index .values.external.addresses .replicaIndex) .values.external.domain) "port" .port -}}
+    {{- else -}}
+      {{- $host = dict "name" .externalName  "address" (index .values.external.addresses .replicaIndex) "port" .port -}}
+    {{- end -}}
+  {{- end -}}
+  {{- toJson $host -}}
+{{- end -}}
+
 {{- $values := .Values }}
 {{- $internalAdvertiseAddress := printf "%s.%s" "$(SERVICE_NAME)" (include "redpanda.internal.domain" .) -}}
 {{- $externalAdvertiseAddress := printf "$(SERVICE_NAME)" -}}
@@ -79,34 +114,27 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.name
-{{- range $listenerName, $listenerVals := $values.listeners }}
+{{- range $listenerName := (list "kafka" "http") }}
+{{- $listenerVals := get $values.listeners $listenerName }}
   {{- $ordList := list }}
+  {{- /* do this for each pod in the statefulset */}}
   {{- range $replicaIndex := until ($values.statefulset.replicas | int) }}
-    {{- $listenerList := list }}
-    {{- if and $listenerVals.external (or (eq $listenerName "kafka") (eq $listenerName "http")) }}
-      {{- $listenerList = append $listenerList (toJson (dict "name" "internal" "address" $internalAdvertiseAddress "port" $listenerVals.port)) }}
+    {{- /* build a list of listeners */}}
+    {{- $listenerList := list (toJson (dict "name" "internal" "address" $internalAdvertiseAddress "port" $listenerVals.port)) }}
+    {{- if $listenerVals.external }}
+        {{- /* add each external listener */}}
         {{- range $externalName, $externalVals := $listenerVals.external }}
-
-          {{- $port := dig "port" $listenerVals.port $externalVals }}
-          {{- if $externalVals.advertisedPorts }}
-            {{- if eq (len $externalVals.advertisedPorts) 1 }}
-              {{- $port = index $externalVals.advertisedPorts 0 }}
-            {{- else }}
-              {{- $port = index $externalVals.advertisedPorts $replicaIndex }}
-            {{- end }}
-          {{- end }}
-
-          {{- $host := dict "name" $externalName "address" $externalAdvertiseAddress "port" $port }}
-          {{- if $values.external.addresses }}
-            {{- if $values.external.domain }}
-              {{- $host = dict "name" $externalName "address" (printf "%s.%s" (index $values.external.addresses $replicaIndex) $values.external.domain) "port" $port }}
-            {{- else }}
-              {{- $host = dict "name" $externalName  "address" (index $values.external.addresses $replicaIndex) "port" $port }}
-            {{- end }}
-          {{- end }}
-          {{- $listenerList = mustAppend $listenerList (toJson $host) }}
+          {{- $tmplVals := dict "listenerVals" $listenerVals "externalVals" $externalVals "replicaIndex" $replicaIndex "externalName" $externalName "externalAdvertiseAddress" $externalAdvertiseAddress "values" $values }}
+          {{- $port := int (include "advertised-port" $tmplVals) }}
+          {{- $host := include "advertised-host" (mustMerge $tmplVals (dict "port" $port)) }}
+          {{- $listenerList = mustAppend $listenerList $host }}
         {{- end }}
     {{- end }}
+    {{- /*
+    This is making a semicolon list of listeners, one list for each pod.
+    We can safely remove all the spaces as there cannot be a space in a name, address, or port, and this fixes
+    a problem where a user can feed in a trailing space on a hostname.
+    */}}
     {{- $ordList = mustAppend $ordList (nospace (join ";" $listenerList)) }}
   {{- end }}
             - name: ADVERTISED_{{ upper $listenerName }}_ADDRESSES
--- a/charts/redpanda/templates/statefulset.yaml
+++ b/charts/redpanda/templates/statefulset.yaml
@@ -15,41 +15,6 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */}}
 
-
-{{- /*
-advertised-port returns either the only advertised port if only one is specified,
-or the port specified for this pod ordinal when there is a full list provided.
-
-This will return a string int or panic if there is more than one port provided,
-but not enough ports for the number of replicas requested.
-*/ -}}
-{{- define "advertised-port" -}}
-  {{- $port := dig "port" .listenerVals.port .externalVals -}}
-  {{- if .externalVals.advertisedPorts -}}
-    {{- if eq (len .externalVals.advertisedPorts) 1 -}}
-      {{- $port = mustFirst .externalVals.advertisedPorts -}}
-    {{- else -}}
-      {{- $port = index .externalVals.advertisedPorts .replicaIndex -}}
-    {{- end -}}
-  {{- end -}}
-  {{ $port }}
-{{- end -}}
-
-{{- /*
-advertised-host returns a json sring with the data neded for configuring the advertised listener
-*/ -}}
-{{- define "advertised-host" -}}
-  {{- $host := dict "name" .externalName "address" .externalAdvertiseAddress "port" .port -}}
-  {{- if .values.external.addresses -}}
-    {{- if .values.external.domain -}}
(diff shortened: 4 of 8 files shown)


vuldin:
I haven't yet gone through the steps for adding a LoadBalancer (or other) service that allows making use of the advertised listeners. That isn't part of this PR, but it will be a very likely use case for some people, so I'd like to document those steps at some point.

Also I'll need to make a few minor changes to the Kubernetes quickstart docs related to this PR's changes.

vuldin:
Steps to make Redpanda's external Kafka listener available via a load balancer (by default port 9092 on each node):

1. Create the file `lb-redpanda-0.yml` with the following content:
```
apiVersion: v1
kind: Service
metadata:
  name: lb-redpanda-0
  namespace: redpanda
spec:
  type: LoadBalancer
  ports:
    - targetPort: 9092
      port: 9092
      nodePort: 31092
  selector:
    statefulset.kubernetes.io/pod-name: redpanda-0
  externalTrafficPolicy: Local
```
2. `kubectl apply -f lb-redpanda-0.yaml`
3.  (For minikube only) `minikube tunnel`
4. Get the external IP with `kubectl get service/lb-redpanda-0 -n redpanda`
5. (From local rpk) `rpk cluster info --broker <external_ip>:31092`

A similar load balancer would need to be created for each broker.

vuldin:
Moving to draft because I'm seeing to issues:
1. The advertised host for each broker is the same
2. Updates to advertised listeners are rejected due to finding a duplicate advertised endpoint. This may be an issue more with the helm upgrade process than this PR, but I need to look more into this.

vuldin:
The two issues mentioned earlier are resolved.
1. This was related to [a partial renaming](https://github.com/redpanda-data/helm-charts/commit/c2145b5e3ea04ca3af7f17fd5ca2674ca1b8c777#diff-6f494fcb2c4edfa7256b33ba89ef8e7c83d92c7f0a6bf6bea53341881e224620R88) of a variable
2. This seems to only be an issue in minikube. I tested upgrading the helm chart multiple times in EKS without any issues. I still want to find out why minikube behaves differently, but at this point I'm sure it is an issue with minikube and not this PR.

JakeSCahill:
Could we introduce `external.type=LoadBalancer`? That way, the Helm chart can control those Services like we do for NodePort and users don't need to worry about manually creating one for each worker node.

charts/redpanda/values.schema.json
```diff
@@ -721,12 +709,12 @@
                     "port": {
                       "type": "integer"
                     },
-                    "type": {
-                      "type": "string",
-                      "pattern": "^NodePort$"
-                    },
-                    "nodePort": {
```
JakeSCahill, line 728: LoadBalancer Services still expose a NodePort, so I think this `nodePort` label is fine to keep.

From the Kubernetes docs:

>To implement a Service of type: LoadBalancer, Kubernetes typically starts off by making the changes that are equivalent to you requesting a Service of type: NodePort. The cloud-controller-manager component then configures the external load balancer to forward traffic to that assigned node port.

charts/redpanda/values.yaml
```diff
@@ -403,21 +399,22 @@ tuning: {}
...
-  # The kafka listener group cannot be disabled
+  # Admin API listener (only one)
   admin:
-    # The port for the admin server
+    # The port for both internal and external connections to the admin API
     port: 9644
     # Optional external section
     external:
+      # Name of the external listener
       default:
```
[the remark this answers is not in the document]
JakeSCahill: do we need the `default` object?

JakeSCahill (commented):
I really like how you've simplified the `external.addresses` and also made enabling/disabling listeners a consistent user experience 😄 

I couldn't see an easy way to set `external.type` to undefined. And I wonder if we can introduce `external.type: LoadBalancer` to offload the management of the Services to the Helm chart.

charts/redpanda/values.yaml
```diff
@@ -102,25 +102,21 @@ tls:
...
-  # Default external access value for all listeners except RPC
-  # External config doesn't apply to RPC listeners as they are never externally accessible
-  # These values can be overridden by each listener if needed
+  # Default external access value
+  # This be overridden by each listener if needed
   enabled: true
-  # Default external access type (only NodePort at the moment)
+  # External access type. `NodePort` is the only acceptable value if defined.
+  # If undefined, then advertised listeners will be configured and a service must be created manually
   type: NodePort
```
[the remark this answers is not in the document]
vuldin: That's a good point, it's difficult (impossible?) to "undefine" a parameter with helm set. I'll make this change.

alejandroEsc (commented):
A few questions and one suggestion.

charts/redpanda/values.yaml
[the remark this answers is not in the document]
vuldin: Actually I was thinking about another possibility rather than creating a `LoadBalancer` value for `external.type` within this PR.

Right now this PR has nothing tying it directly to a LoadBalancer service. It just makes it possible for the user to manually create the service that will provide external connectivity, whatever it is. I do think adding `LoadBalancer` to the helm chart is a good idea, as that will simplify creating a LoadBalancer service... but there are questions to answer and technical issues to solve.

One such issue is one vs many LoadBalancers. This PR makes it possible to define a different port for each node, which means that I think it could work to provide external access via a single LoadBalancer service (where different ports get you to different brokers). This would be a big benefit, but we need to make sure and handle the case where the user defines the same port for each node (which would only work with multiple LoadBalancers). We also need to make sure how it's defined in values.yaml is the most understandable way to do so.

Instead of answering these questions above here, I suggest we add in a `none` value for `external.type`. This would do what it suggests: no service of any type would be created. Instead you are only defining advertised listeners. Later once we have an actual LoadBalancer PR up and have answered the above questions, then we can add a `LoadBalancer` option.

Please post if you have any feedback, or take a look at the incoming (small) change to this PR.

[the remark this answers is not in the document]
JakeSCahill: Sounds sensible, thanks @vuldin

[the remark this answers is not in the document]
joejulian: > That's a good point, it's difficult (impossible?) to "undefine" a parameter with helm set. I'll make this change.

```
--set things.unwanted=null
```

will remove "unwanted" from the "things" dict

charts/redpanda/values.yaml
```diff
@@ -427,47 +424,39 @@ listeners:
...
+        # If undefined, `listeners.kafka.external.port` is used
+        advertisedPorts:
+        - 31092
   # HTTP API listeners (aka PandaProxy)
-  # PandaProxy is a kafka client that connects to an endpoint from listeners.kafka.endpoints
   http:
     enabled: true
-    port: 8082
+    port: 9082
+    # TODO is this for internal and external clients? why can't internal connect to kafka internal, and each external connect to matchin kafka external? then this can be removed
```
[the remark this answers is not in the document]
joejulian: No, it's only for the internal listener. If you go down to the `external` section, you'll see that `default` has its own port. This is a requirement because the port that's connected to determines which listener profile (and advertised listener set) is being used.

charts/redpanda/Chart.yaml
```diff
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 2.4.0
+version: 2.4.1
```
joejulian, line 26:
```suggestion
version: 2.5.0
```

charts/redpanda/templates/configmap.yaml
```diff
@@ -150,7 +150,7 @@ data:
 {{- end }}
 {{- if .Values.listeners.schemaRegistry.enabled }}
     schema_registry:
-      schema_registry:
```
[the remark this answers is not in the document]
joejulian: +1

charts/redpanda/templates/services.nodeport.yaml
```diff
@@ -38,42 +38,38 @@ spec:
...
 {{- range $name, $listener := $values.listeners.admin.external }}
   {{- $enabled := dig "enabled" $values.external.enabled $listener }}
-  {{- $type := dig "type" $values.external.type $listener }}
-  {{- if and $enabled (eq $type "NodePort") }}
+  {{- if $enabled }}
     - name: admin-{{ $name }}
       protocol: TCP
       port: {{ $values.listeners.admin.port }}
-      nodePort: {{ $listener.nodePort }}
+      nodePort: {{ index $listener.advertisedPorts 0 }}
```
joejulian, line 45:
```suggestion
      nodePort: {{ first $listener.advertisedPorts }}
```

joejulian: This is a breaking change. This will need to be templated to ensure the existing users will not have their nodePort redefined.

charts/redpanda/templates/services.nodeport.yaml
```diff
@@ -15,7 +15,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */}}
 {{- $values := .Values }}
-{{- if (include "external-nodeport-enabled" . | fromJson).bool }}
+{{- if (and .Values.external.enabled (eq .Values.external.type "NodePort")) }}
```
joejulian, line 18: Why are we bringing this back in? I prefer to have simple booleans inside the templates when possible, leaving the complexity in the named templates.

charts/redpanda/templates/services.nodeport.yaml
```diff
@@ -38,42 +38,38 @@ spec:
...
 {{- range $name, $listener := $values.listeners.kafka.external }}
   {{- $enabled := dig "enabled" $values.external.enabled $listener }}
-  {{- $type := dig "type" $values.external.type $listener }}
-  {{- if and $enabled (eq $type "NodePort") }}
+  {{- if $enabled }}
     - name: kafka-{{ $name }}
       protocol: TCP
       port: {{ $listener.port }}
-      nodePort: {{ $listener.nodePort }}
+      nodePort: {{ index $listener.advertisedPorts 0 }}
```
joejulian, line 54: same comments as admin section

charts/redpanda/templates/services.nodeport.yaml
```diff
@@ -38,42 +38,38 @@ spec:
...
 {{- range $name, $listener := $values.listeners.http.external }}
   {{- $enabled := dig "enabled" $values.external.enabled $listener }}
-  {{- $type := dig "type" $values.external.type $listener }}
-  {{- if and $enabled (eq $type "NodePort") }}
+  {{- if $enabled }}
     - name: http-{{ $name }}
       protocol: TCP
       port: {{ $listener.port }}
-      nodePort: {{ $listener.nodePort }}
+      nodePort: {{ index $listener.advertisedPorts 0 }}
```
joejulian, line 63: same comments as admin section

charts/redpanda/templates/services.nodeport.yaml
```diff
@@ -38,42 +38,38 @@ spec:
...
   {{- $enabled := dig "enabled" $values.external.enabled $listener }}
-  {{- $type := dig "type" $values.external.type $listener }}
-  {{- if and $enabled (eq $type "NodePort") }}
+  {{- if $enabled }}
     - name: schema-{{ $name }}
       protocol: TCP
-      port: {{ $listener.port }}
-      nodePort: {{ $listener.nodePort }}
+      port: {{ $values.listeners.schemaRegistry.port }}
+      nodePort: {{ index $listener.advertisedPorts 0 }}
```
joejulian, line 72: same comments as admin section

charts/redpanda/templates/statefulset.yaml
```diff
@@ -253,18 +303,14 @@ spec:
...
+            - --advertise-rpc-addr={{ $internalAdvertiseAddress }}:{{ .Values.listeners.rpc.port }}
           ports:
 {{- range $name, $listener := .Values.listeners }}
-            - name: {{ lower $name }}
+            - name: {{ lower $name | trunc 6 }}-internal
               containerPort: {{ $listener.port }}
   {{- range $externalName, $external := $listener.external }}
     {{- if $external.port }}
-            - name: {{ lower $name | trunc 6 }}-{{ lower $externalName | trunc 8}}
+            - name: {{ lower $name | trunc 4 }}-ext-{{ lower $externalName | trunc 6 }}
```
joejulian, line 313: Isn't internal already named "internal"? I'd prefer we not truncate these any more than we already have. If the default names are not descriptive enough (and "default" may be a bit too generic), let's change those.

charts/redpanda/templates/tests/test-schemaregistry-status.yaml
```diff
@@ -49,5 +49,5 @@ spec:
         - "300"
         - --retry-max-time
         - "120"
-        - http://{{ include "redpanda.internal.domain" . }}:{{ .Values.listeners.schemaRegistry.port }}/subjects
+        - http://{{ include "redpanda.fullname" . }}:{{ .Values.listeners.schemaRegistry.port }}/subjects
```
joejulian, line 52: Why are we changing this test? Isn't connecting to the aggregate service preferred?

charts/redpanda/values.schema.json
```diff
@@ -651,23 +645,19 @@
             },
             "external": {
               "type": "object",
-              "minProperties": 1,
```
joejulian, line 654: Add a test with external set to an empty set if you change this. The chart originally depended on there being at least one external listener defined (even if external was disabled).

charts/redpanda/values.schema.json
```diff
@@ -697,7 +687,6 @@
           "type": "object",
           "required": [
             "port",
-            "external",
```
joejulian, line 700: external is required. If it's missing `$listenerVals.external` will panic.

charts/redpanda/values.yaml
```diff
@@ -102,25 +102,21 @@ tls:
 #
 # External access configuration
 external:
-  # Default external access value for all listeners except RPC
-  # External config doesn't apply to RPC listeners as they are never externally accessible
```
joejulian, line 106: Is this no longer true?

charts/redpanda/values.yaml
```diff
@@ -102,25 +102,21 @@ tls:
...
-  # These values can be overridden by each listener if needed
+  # Default external access value
+  # This be overridden by each listener if needed
   enabled: true
-  # Default external access type (only NodePort at the moment)
+  # External access type. `NodePort` is the only acceptable value if defined.
+  # If undefined, then advertised listeners will be configured and a service must be created manually
```
joejulian, line 109:
```suggestion
  # If undefined, then advertised listeners will be configured in Redpanda, but the helm chart will not create a Service. You must create a Service manually.
```

charts/redpanda/values.yaml
```diff
@@ -403,21 +399,22 @@ tuning: {}
 # listeners.
 #
 listeners:
-  # Admin API listener
-  # The kafka listener group cannot be disabled
+  # Admin API listener (only one)
   admin:
-    # The port for the admin server
+    # The port for both internal and external connections to the admin API
```
joejulian, line 404: This cannot be true.

joejulian:
Renumbering all the ports will break existing users. Please either find an alternative, or provide backward compatibility.

charts/redpanda/templates/configmap.yaml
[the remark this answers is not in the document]
vuldin: This is a fix to the helm chart to adhere to the Redpanda API, as described here: https://docs.redpanda.com/docs/platform/reference/node-configuration-sample/

charts/redpanda/templates/services.nodeport.yaml
vuldin: I needed to make a change to the logic on this, but then realized the only place this named template is ever used is in the nodeport service file. So I just included the logic here. I guess it's possible that someday this named template would need to be used in some other file though, and if that's the case it would make sense to keep it around.

But the named template could be brought back in regardless, let me know.

charts/redpanda/templates/statefulset.yaml
vuldin: I didn't want to truncate here, but it is required that port names are 15 or less characters. The deployment will fail otherwise. So I truncated the name in the best way I could figure out (to keep defails from each part of the name) to fit within the 15 character restriction. At one point I found the kubernetes docs for this, but just now I couldn't find the link again. But incrementing either the 4 or 6 (or just adding in a single character anywhere) will show you the error.

charts/redpanda/templates/tests/test-schemaregistry-status.yaml
vuldin: This is just making this test match each of the other similar tests we have. If this is incorrect, then the other tests may also need to be updated similarly. I don't recall why this came up during this PR though... it definitely isn't related and could be handled in a separate PR.

charts/redpanda/values.schema.json
vuldin: I may have mistakenly removed `external` from here (rather than another place I meant to). I'll take a closer look and test it out again.

charts/redpanda/values.yaml
vuldin: This is true, but this line was moved to this location: https://github.com/redpanda-data/helm-charts/blob/d88954e09e4007346a1fbe576fa7896b16e87f4b/charts/redpanda/values.yaml#L459

charts/redpanda/values.yaml
vuldin: There is only ever a single admin API port. Internal and external clients always connect to the same port.

charts/redpanda/values.yaml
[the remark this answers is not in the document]
vuldin: This is a TODO line that I meant to remove beforehand, I'll remove the line.

anggutie-dev:
Hi there,


I'm waiting for this PR to be merged as well, I have the need to expose my Redpanda instance using a LoadBalancer (Private GKE).

Is there an ETA?

vuldin:
> Hi there,
> 
> I'm waiting for this PR to be merged as well, I have the need to expose my Redpanda instance using a LoadBalancer (Private GKE).
> 
> Is there an ETA?

Hey @anggutie-dev ! No ETA yet, but this PR is in active review so shouldn't be much longer. I would expect it to be done within a week tops (unless something major takes priority in the meantime).

charts/redpanda/templates/configmap.yaml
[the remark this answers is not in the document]
joejulian: Ah, bug fix. Glad you caught this. Any thoughts on how we can lint the redpanda.yaml in a test?

charts/redpanda/templates/services.nodeport.yaml
joejulian: Ok, then I'd prefer to leave this change out of this PR. We can discuss it in a separate PR if you feel strongly.

charts/redpanda/templates/statefulset.yaml
joejulian: Right, but aren't we truncating only because we're changing the names and adding extra characters?

charts/redpanda/values.yaml
joejulian: Ah, right, so there's no "advertised listeners" for admin then.

charts/redpanda/values.yaml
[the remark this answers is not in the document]
vuldin: This TODO does have a purpose though, so I will create a related issue.

`kakfaEndpoint` exists to connect the http listener internally to the proper kafka listener based on name of the listener. But we could automate this by ensuring the internal http listener connects to the internal kafka listener, and then each external http listener connects to the kafka listener of the same name as itself (which by default, the one external listener created for both http and kafka is named `default`).

https://github.com/redpanda-data/helm-charts/issues/284

charts/redpanda/templates/services.nodeport.yaml
vuldin: So rather than relying on semver to notify the user of a breaking change, you suggest that we have a template to determine if the user is trying to use a `values.yaml` file from a previous version? I think I may need to reach out and clarify what the template would try to detect here.

charts/redpanda/templates/statefulset.yaml
vuldin: Ahh, you are more asking why I have made this modification in this PR rather than asking why `trunc` is needed.

The change is not strictly needed, I just realized while developing this PR that we did not explicitly name the internal port with anything that mentioned `internal` (and same for external ports).

I'll remove this change to keep the PR more focused.

vuldin:
Regarding the port changes in this PR:

### Existing

Redpanda has listeners for various services, and most of these services have an internal and external variant. At the same time, we could have multiple external listeners. So for example, the kafka and http APIs by default have the following listeners/ports:

| listener | docs | internal | external | advertised |
|---|---|---|---|---|
| admin | 9644 | 9644 | N/A | 31644
| kafka | 9092 | 9093 | 9094 | 31092
| http | 8082 | 8082 | 8083 | 30082
| schema | 8081 | 8081 | 8080 | 30081

The above state is confusing, as the Kafka API is not available anywhere via port 9092. Also, there isn't any similarity or formula for how ports are handled across the different listeners (neither within kafka listeners nor between kafka and http listeners). The listener ports are incremented by 1 for admin and kafka, but decremented by 1 for the schema registry. The reason is because incrementing by 1 is impossible when schema and http services are at 8081 and 8082 respectively. Another issue is the schema registry has an external port defined, but in reality it uses the same port across internal and external requests (same as admin API). Advertised ports must be within the NodePort range, so we are adding 22K to the API's documented port.

### Future

The port changes in this PR apply the same formula across all listeners:

| listener | docs | internal | external | advertised |
|---|---|---|---|---|
| admin | 9644 | 9644 | N/A | 31644
| kafka | 9092 | 10092 | 9092 | 31092
| http | 8082 | 9082 | 8082 | 30082
| schema | 8081 | 8081 | N/A | 30081

- The external ports match our documentation and the default behavior of the Kafka ecosystem. Most people will want to connect to a Redpanda cluster via external clients, so internal ports are modified instead (see below)
- The internal port is `+1K` to the documented port. This is applied across each listener when needed. This modification is needed only when the API has an external port (http and kafka APIs).
- Advertised ports is `+22K` to the documented port to bump it up into the NodePort range.
- The schema registry external port is removed, since all schema clients connect to the same port. This was improperly included in the listener definition in an attempt to keep all listener definitions as similar as possible.

charts/redpanda/values.schema.json
```diff
@@ -201,12 +199,8 @@
           "type": "string",
           "format": "idn-hostname"
         },
-        "addressType": {
-          "type": "string",
-          "pattern": "^(subdomain|ip)$"
-        },
         "addresses": {
-          "type": "string"
+          "type": "array"
```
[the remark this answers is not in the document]
JakeSCahill: I'll make this change as part of https://github.com/redpanda-data/documentation/pull/999

alejandroEsc (commented):
lgtm so far, one ask that we can talk about if you want off channel

charts/redpanda/values.yaml
```diff
@@ -102,25 +102,21 @@ tls:
 #
 # External access configuration
 external:
-  # Default external access value for all listeners except RPC
-  # External config doesn't apply to RPC listeners as they are never externally accessible
-  # These values can be overridden by each listener if needed
+  # Default external access value
+  # This be overridden by each listener if needed
```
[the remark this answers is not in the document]
vuldin: I'll try to change the wording to make it more clear.

`external.enabled` is the default value for all listeners. But each service/API can determine whether they create an external listener or not. If `external.enabled` is true, then you can disable external access individually on each service/API. If it's false, then you can enable in the same way.

charts/redpanda/templates/statefulset.yaml
```diff
@@ -15,8 +15,47 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */}}
 
+
```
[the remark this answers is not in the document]
vuldin: I'll move these named templates in spite of them never being used outside of this file nor likely needing to be used outside of this file, if that's the direction we want to go. But doing so feels wrong for a few reasons:

1. as stated above, these templates aren't used anywhere else and are likely to not be used anywhere else. My default approach is that code used in one location should be kept in that location until the time that it makes sense to share it more broadly... and many times that time never comes.
2. `_helpers.tpl` is grab-all collection of templates, many of which aren't used more than once, some of which aren't used at all, and has been a source of issues in the past where templates added there will cause errors if used anywhere else other than in the one place it was made for. The maintenance of these templates suffers in comparison to other areas of the project.

Maybe another approach that would resolve the above issues while also moving named templates out of this Kubernetes template file is to create a separate `_statefulset.tpl` where these can live. Then we can have named templates separated by where they are used when appropriate, and included in `_helpers.tpl` when they need to be used in more than one place. This PR is likely not the place for this change though (if it's the chosen way forward).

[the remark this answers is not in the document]
alejandroEsc: Thank you for that. I was also going to suggest _statefulset.tpl. @joejulian is this ok with you? If not, then sure let's keep things as is for now but my concern is just how big and unreadable this file is becoming. If we can move internal logic outside and clean up the yaml a bit that would help make this more understandable. But again, thats a matter of my opinion.

alejandroEsc (approved):
Not going to block this. There is an open question regarding template files but we can negotiate this outside of this PR on a later date. Thank you!

joejulian (approved):
lgtm. Very nice!

Issue #203, opened 2022-11-15
Deploying Redpanda to K8s for production


RafalKorepta:

vladoschreiner:
This is a good example of a consumer of this content: https://redpandacommunity.slack.com/archives/C01AJDUT88N/p1669046127254399

PR #259, opened 2023-01-05
Add production tunable configuration

RafalKorepta:
Redpanda configuration that is cluster wide and could be changed on runtime.

|                                | default | this PR | available from   | source code link                                                                                                                       |
|--------------------------------|---------|---------|------------------|----------------------------------------------------------------------------------------------------------------------------------------|
| log_segment_size               | 1 GiB   | 128 Mb  | at least 21.11.x | [line 37](https://github.com/redpanda-data/redpanda/blob/bd322f74606f52ef7cee8dde1f64b32068105ce4/src/v/config/configuration.cc#L37)   |
| log_segment_size_min           | 1 MiB   | 16 Mb   | 22.3.x           | [line 47](https://github.com/redpanda-data/redpanda/blob/bd322f74606f52ef7cee8dde1f64b32068105ce4/src/v/config/configuration.cc#L47)   |
| log_segment_size_max           | null    | 256 Mb  | 22.3.x           | [line 56](https://github.com/redpanda-data/redpanda/blob/bd322f74606f52ef7cee8dde1f64b32068105ce4/src/v/config/configuration.cc#L56)   |
| kafka_batch_max_bytes          | 1 MiB   | 1 Mb    | 22.3.x           | [line 867](https://github.com/redpanda-data/redpanda/blob/bd322f74606f52ef7cee8dde1f64b32068105ce4/src/v/config/configuration.cc#L867) |
| topic_partitions_per_shard     | 7000    | 1000    | 22.2.x           | [line 161](https://github.com/redpanda-data/redpanda/blob/bd322f74606f52ef7cee8dde1f64b32068105ce4/src/v/config/configuration.cc#L161) |
| compacted_log_segment_size     | 256 MiB | 64 Mb   | at least 21.11.x | [line 74](https://github.com/redpanda-data/redpanda/blob/bd322f74606f52ef7cee8dde1f64b32068105ce4/src/v/config/configuration.cc#L74)   |
| max_compacted_log_segment_size | 5 GiB   | 512 Mb  | at least 21.11.x | [line 803](https://github.com/redpanda-data/redpanda/blob/bd322f74606f52ef7cee8dde1f64b32068105ce4/src/v/config/configuration.cc#L803) |
| kafka_connection_rate_limit    | null    | 1000    | 22.1.x           | [line 438](https://github.com/redpanda-data/redpanda/blob/bd322f74606f52ef7cee8dde1f64b32068105ce4/src/v/config/configuration.cc#L438) |
| group_topic_partitions         | 16      | 16      | at least 21.11.x | [line 507](https://github.com/redpanda-data/redpanda/blob/bd322f74606f52ef7cee8dde1f64b32068105ce4/src/v/config/configuration.cc#L507) |

REF:
https://github.com/redpanda-data/helm-charts/issues/203

### DIFF
--- a/charts/redpanda/values.yaml
+++ b/charts/redpanda/values.yaml
@@ -550,7 +550,19 @@ config:
     # tm_sync_timeout_ms: 2000ms                                   # Time to wait state catch up before rejecting a request
     # tm_violation_recovery_policy: crash                          # Describes how to recover from an invariant violation happened on the transaction coordinator level
     # transactional_id_expiration_ms: 10080min                     # Producer ids are expired once this time has elapsed after the last write with the given producer ID
-  tunable: {}
+  tunable:
+    log_segment_size: 134217728                                    # 128 mb
+    log_segment_size_min: 16777216                                 # 16 mb
+    log_segment_size_max: 268435456                                # 256 mb
+    kafka_batch_max_bytes: 1048576                                 # 1 mb
+    topic_partitions_per_shard: 1000
+    compacted_log_segment_size: 67108864                           # 64 mb
+    max_compacted_log_segment_size: 536870912                      # 512 mb
+    kafka_connection_rate_limit: 1000
+    group_topic_partitions: 16
+    # cloud_storage_enable_remote_read: true                       # cluster wide configuration for read from remote cloud storage
+    # cloud_storage_enable_remote_write: true                      # cluster wide configuration for writing to remote cloud storage
+
     # alter_topic_cfg_timeout_ms: 5s                               # Time to wait for entries replication in controller log when executing alter configuration request
     # compacted_log_segment_size: 256MiB                           # How large in bytes should each compacted log segment be (default 256MiB)
     # controller_backend_housekeeping_interval_ms: 1s              # Interval between iterations of controller backend housekeeping loop
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 2.4.4
+version: 2.4.5
 
 # The app version is the default version of Redpanda to install.
 appVersion: v22.3.10
--- a/charts/redpanda/templates/_helpers.tpl
+++ b/charts/redpanda/templates/_helpers.tpl
@@ -419,6 +419,24 @@ IP is required for the advertised address.
 {{- end -}}
 {{- end -}}
 
+{{- define "tunable" -}}
+{{- $tunable := dig "tunable" dict .Values.config }}
+{{- if (include "redpanda-atleast-22-3-0" . | fromJson).bool }}
+{{- toYaml $tunable | nindent 4 }}
+{{- else if (include "redpanda-atleast-22-2-0" . | fromJson).bool }}
+{{- $tunable = unset $tunable "log_segment_size_min" }}
+{{- $tunable = unset $tunable "log_segment_size_max" }}
+{{- $tunable = unset $tunable "kafka_batch_max_bytes" }}
+{{- toYaml $tunable | nindent 4 }}
+{{- else if (include "redpanda-atleast-22-1-1" . | fromJson).bool }}
+{{- $tunable = unset $tunable "log_segment_size_min" }}
+{{- $tunable = unset $tunable "log_segment_size_max" }}
+{{- $tunable = unset $tunable "kafka_batch_max_bytes" }}
+{{- $tunable = unset $tunable "topic_partitions_per_shard" }}
+{{- toYaml $tunable | nindent 4 }}
+{{- end }}
+{{- end }}
+
 {{- define "redpanda-atleast-22-1-1" -}}
 {{- toJson (dict "bool" (or (not (eq .Values.image.repository "vectorized/redpanda")) (include "redpanda.semver" . | semverCompare ">=22.1.1"))) -}}
 {{- end -}}
--- a/charts/redpanda/templates/configmap.yaml
+++ b/charts/redpanda/templates/configmap.yaml
@@ -46,9 +46,7 @@ data:
   {{- with (dig "cluster" dict .Values.config) }}
     {{- toYaml . | nindent 4 }}
   {{- end }}
-  {{- with (dig "tunable" dict .Values.config) }}
-    {{- toYaml . | nindent 4 }}
-  {{- end }}
+  {{- include "tunable" . }}
   {{- if and (not (hasKey .Values.config.cluster "storage_min_free_bytes")) ((include "redpanda-atleast-22-2-0" . | fromJson).bool) }}
     storage_min_free_bytes: {{ include "storage-min-free-bytes" . }}
   {{- end }}
--- a/charts/redpanda/templates/configmap.yaml
+++ b/charts/redpanda/templates/configmap.yaml
@@ -79,7 +79,7 @@ data:
   {{- with (dig "tunable" dict .Values.config) }}
       {{- toYaml . | nindent 6 }}
   {{- end }}
-  {{- if not (hasKey "storage_min_free_bytes" .Values.config.cluster) }}
+  {{- if not (hasKey .Values.config.cluster "storage_min_free_bytes") }}
       storage_min_free_bytes: {{ include "storage-min-free-bytes" . }}
   {{- end }}
 {{- end }}


charts/redpanda/values.yaml
```diff
@@ -550,7 +550,22 @@ config:
...
-  tunable: {}
+  tunable:
+    log_segment_size: 134217728                                    # 128 mb
+    log_segment_size_min: 16777216                                 # 16 mb
+    log_segment_size_max: 268435456                                # 256 mb
+    kafka_batch_max_bytes: 1048576                                 # 1 mb
+    topic_partitions_per_shard: 1000
+    compacted_log_segment_size: 67108864                           # 64 mb
+    max_compacted_log_segment_size: 536870912                      # 512 mb
+    kafka_connections_max: 15100
```
jcsp, line 561: Setting a connection limit probably doesn't make sense, if you don't know how large the machines you're installing onto are.

charts/redpanda/values.yaml
```diff
@@ -550,7 +550,22 @@ config:
...
+    kafka_connection_rate_limit: 1000
+    partition_autobalancing_mode: "continuous"
```
jcsp, line 563: continuous autobalancing requires a license, so should not be on by default unless helm can know whether it's installing with a license.

charts/redpanda/values.yaml
```diff
@@ -550,7 +550,22 @@ config:
...
+    cloud_storage_segment_max_upload_interval_sec: 3600            # 3600 sec = 1 hour
```
jcsp, line 564: The upload interval setting should be removed, this only makes sense in the context of some intended RPO.

charts/redpanda/values.yaml
```diff
@@ -550,7 +550,22 @@ config:
     # tm_sync_timeout_ms: 2000ms                                   # Time to wait state catch up before rejecting a request
     # tm_violation_recovery_policy: crash                          # Describes how to recover from an invariant violation happened on the transaction coordinator level
     # transactional_id_expiration_ms: 10080min                     # Producer ids are expired once this time has elapsed after the last write with the given producer ID
-  tunable: {}
+  tunable:
+    log_segment_size: 134217728                                    # 128 mb
```
jcsp, line 554: There's a docs ticket here https://github.com/redpanda-data/documentation/issues/1000 for describing how Cloud systems have a different default than self hosted systems.

If self hosted (helm) systems start using a different set of defaults too (which will probably diverge from Cloud eventually), then that should be flagged to the docs team as well.

charts/redpanda/values.yaml
```diff
@@ -550,7 +550,22 @@ config:
...
+    kafka_batch_max_bytes: 1048576                                 # 1 mb
+    topic_partitions_per_shard: 1000
+    compacted_log_segment_size: 67108864                           # 64 mb
+    max_compacted_log_segment_size: 536870912                      # 512 mb
+    kafka_connections_max: 15100
+    kafka_connection_rate_limit: 1000
+    partition_autobalancing_mode: "continuous"
+    cloud_storage_segment_max_upload_interval_sec: 3600            # 3600 sec = 1 hour
+    group_topic_partitions: 16
+    # cloud_storage_enable_remote_read: true                       # cluster wide configuration for read from remote cloud storage
```
jcsp, line 566: Commented lines left in by mistake?

jcsp:
Be aware that this will make helm stop working with pre-v22.3.x redpandas: I don't mind, but i don't know if we document anywhere what Redpanda versions the helm chart is meant to work with.

charts/redpanda/values.yaml
joejulian: No, commented lines are left in as documentation.

joejulian:
> Be aware that this will make helm stop working with pre-v22.3.x redpandas: I don't mind, but i don't know if we document anywhere what Redpanda versions the helm chart is meant to work with.

I assume you mean that these defaults will do that. The user can override any of these defaults.

joejulian:
What's the source for these values? I think you said this from some performance tested configuration that's documented somewhere, didn't you?

RafalKorepta:
> What's the source for these values? I think you said this from some performance tested configuration that's documented somewhere, didn't you?

It is in our internal repo https://github.com/redpanda-data/team-kubernetes-internal/issues/3

charts/redpanda/values.yaml
RafalKorepta: I mentioned it to documentation team in the ticket you attached https://github.com/redpanda-data/documentation/issues/1000#issuecomment-1380099331.

charts/redpanda/values.yaml
jcsp: Nope.  In cloud SaaS contexts, the connection count is part of the definition of the product, we do not have a matrix of how many connections a given self-hosted machine type can handle (with memory limits, it's hard to define, because a system with lots of partitions can handle fewer clients, etc, so it depends on the workload).

The default for self hosted clusters today is not to apply a connection  count limit.

charts/redpanda/values.yaml
jcsp: It's complicated.  For things like this where there is no simple answer, I don't think the Helm chart is the place to try and address this for self hosted systems: our lives will be simpler if Helm just inherits the redpanda default (which is null), and thereby matches other self hosted systems.

jcsp:
> I assume you mean that these defaults will do that. The user can override any of these defaults.

Right, it just means the defaults won't work with older clusters.  It might not be super easy for the user to figure out why, they'd have to go look at redpanda logs to see errors like this, when their pods won't stay up:
```
3532208: ERROR 2023-01-12 10:19:36,031 [shard 0] main - application.cc:337 - Failure during startup: std::invalid_argument (Unknown property bloogle)
```

charts/redpanda/values.yaml
alejandroEsc: So now we are adding these items, what are the sideaffects to existing and and upgrading previous versions?

alejandroEsc (commented):
Just a question before i approve. What defaults were taken before we added these fields? Are the aligned? What is the side effects of adding these now? If there is nothing to worry about im OK in approving.

charts/redpanda/values.yaml
RafalKorepta: This comment is pinned to lines 566. It is parameter available in Redpanda from 22.11.X version as far as I remember. If you have license (not enforced yet) and configures tiered storage https://docs.redpanda.com/docs/platform/data-management/remote-read-replicas/#creating-a-topic-with-archival-storage-or-tiered-storage

RafalKorepta:
> What defaults were taken before we added these fields?

I will update cover letter to address this comment with the appropriate link to the source code.

>  What is the side effects of adding these now?

You mean what are the side effects to add this if some cluster is already running?

It depends on the cluster version. As helm chart doesn't have validation webhook I will probably use clunky templating system to guard this.

alejandroEsc:
> Redpanda configuration that is cluster wide and could be changed on runtime.
> 
> default	this PR	available from	source code link
> log_segment_size	1 GiB	128 Mb	at least 21.11.x	[line 37](https://github.com/redpanda-data/redpanda/blob/bd322f74606f52ef7cee8dde1f64b32068105ce4/src/v/config/configuration.cc#L37)
> log_segment_size_min	1 MiB	16 Mb	22.3.x	[line 47](https://github.com/redpanda-data/redpanda/blob/bd322f74606f52ef7cee8dde1f64b32068105ce4/src/v/config/configuration.cc#L47)
> log_segment_size_max	null	256 Mb	22.3.x	[line 56](https://github.com/redpanda-data/redpanda/blob/bd322f74606f52ef7cee8dde1f64b32068105ce4/src/v/config/configuration.cc#L56)
> kafka_batch_max_bytes	1 MiB	1 Mb	22.3.x	[line 867](https://github.com/redpanda-data/redpanda/blob/bd322f74606f52ef7cee8dde1f64b32068105ce4/src/v/config/configuration.cc#L867)
> topic_partitions_per_shard	7000	1000	22.2.x	[line 161](https://github.com/redpanda-data/redpanda/blob/bd322f74606f52ef7cee8dde1f64b32068105ce4/src/v/config/configuration.cc#L161)
> compacted_log_segment_size	256 MiB	64 Mb	at least 21.11.x	[line 74](https://github.com/redpanda-data/redpanda/blob/bd322f74606f52ef7cee8dde1f64b32068105ce4/src/v/config/configuration.cc#L74)
> max_compacted_log_segment_size	5 GiB	512 Mb	at least 21.11.x	[line 803](https://github.com/redpanda-data/redpanda/blob/bd322f74606f52ef7cee8dde1f64b32068105ce4/src/v/config/configuration.cc#L803)
> kafka_connection_rate_limit	null	1000	22.1.x	[line 438](https://github.com/redpanda-data/redpanda/blob/bd322f74606f52ef7cee8dde1f64b32068105ce4/src/v/config/configuration.cc#L438)
> group_topic_partitions	16	16	at least 21.11.x	[line 507](https://github.com/redpanda-data/redpanda/blob/bd322f74606f52ef7cee8dde1f64b32068105ce4/src/v/config/configuration.cc#L507)
> REF: #203

Wow thank you for that, did not need you to do that, just wanted to know if these were known/defaults etc.. but awesome!

alejandroEsc:
> > I assume you mean that these defaults will do that. The user can override any of these defaults.
> 
> Right, it just means the defaults won't work with older clusters. It might not be super easy for the user to figure out why, they'd have to go look at redpanda logs to see errors like this, when their pods won't stay up:
> 
> ```
> 3532208: ERROR 2023-01-12 10:19:36,031 [shard 0] main - application.cc:337 - Failure during startup: std::invalid_argument (Unknown property bloogle)
> ```

If this is the case, we can pre-empt these changes in a release note for our users. Is there an easy way for them to determine what their current defaults are so they can take actions pre-emptively against these changes? I am assuming of course that they are not upgrading their redpanda images.

RafalKorepta:
> Is there an easy way for them to determine what their current defaults are so they can take actions pre-emptively against these changes?

Using Admin API they can see the values of the running cluster. There is even `rpk cluster config export` command to do this.

emaxerrno:
@RafalKorepta  why. not have all of the segment size match at 128MB? `compacted_log_segment_size` - @jcsp  thoughts?

joejulian (approved):

charts/redpanda/values.yaml
joejulian: You can just check that one of these is set: https://github.com/redpanda-data/helm-charts/blob/e228c875180d5aa6ac3918de45de02cc8b2583d3/charts/redpanda/values.yaml#L45-L46

alejandroEsc (approved):
thank you for your responses!

RafalKorepta:
Added template that will strip unknown property based on the cover letter
```
$ helm template redpanda charts/redpanda --set image.tag=v22.2.1
---
# Source: redpanda/templates/configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: redpanda
  namespace: "redpanda"
  labels:
    helm.sh/chart: redpanda-2.4.5
    app.kubernetes.io/instance: "redpanda"
    app.kubernetes.io/managed-by: "Helm"
    app.kubernetes.io/component: redpanda
data:
  bootstrap.yaml: |
    enable_sasl: false
    compacted_log_segment_size: 67108864
    group_topic_partitions: 16
    kafka_connection_rate_limit: 1000
    log_segment_size: 134217728
    max_compacted_log_segment_size: 536870912
    topic_partitions_per_shard: 1000
    storage_min_free_bytes: 1073741824
...

$ helm template redpanda charts/redpanda --set image.tag=v22.3.1
---
# Source: redpanda/templates/configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: redpanda
  namespace: "redpanda"
  labels:
    helm.sh/chart: redpanda-2.4.5
    app.kubernetes.io/instance: "redpanda"
    app.kubernetes.io/managed-by: "Helm"
    app.kubernetes.io/component: redpanda
data:
  bootstrap.yaml: |
    enable_sasl: false
    compacted_log_segment_size: 67108864
    group_topic_partitions: 16
    kafka_batch_max_bytes: 1048576
    kafka_connection_rate_limit: 1000
    log_segment_size: 134217728
    log_segment_size_max: 268435456
    log_segment_size_min: 16777216
    max_compacted_log_segment_size: 536870912
    topic_partitions_per_shard: 1000
    storage_min_free_bytes: 1073741824
...

$ helm template redpanda charts/redpanda --set image.tag=v22.1.1
---
# Source: redpanda/templates/configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: redpanda
  namespace: "redpanda"
  labels:
    helm.sh/chart: redpanda-2.4.5
    app.kubernetes.io/instance: "redpanda"
    app.kubernetes.io/managed-by: "Helm"
    app.kubernetes.io/component: redpanda
data:
  bootstrap.yaml: |
    enable_sasl: false
    compacted_log_segment_size: 67108864
    group_topic_partitions: 16
    kafka_connection_rate_limit: 1000
    log_segment_size: 134217728
    max_compacted_log_segment_size: 536870912
```

And I fixed wrong invocation of `hasKey` function.

PR #260, opened 2023-01-06
fix: use pod ordinal for the node-index to select correct address

### LINKED ISSUE #258: external.addresses only picks up the first value

alejandroEsc:
Attempting to fix #258 

I verified that the pods are using the correct entry on the address string after making this change.

### DIFF
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 2.4.2
+version: 2.4.3
 
 # The app version is the default version of Redpanda to install.
 appVersion: v22.3.9
--- a/charts/redpanda/templates/statefulset.yaml
+++ b/charts/redpanda/templates/statefulset.yaml
@@ -117,8 +117,8 @@ spec:
                 {{- $advertiseKafkaPort := $listener.nodePort }}
                 {{- if $listenerNodePortEnabled }}
                   {{- if $values.external.addresses }}
-              NODE_INDEX=`expr $NODE_ID + 1`
-              NODE_ADDRESS=`echo $EXTERNAL_ADDRESSES | cut -d ' ' -f $NODE_INDEX`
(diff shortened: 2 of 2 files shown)


JakeSCahill:
Thanks @alejandroEsc . This fixes the issue for the `advertised_kafka_api`, but the issue still remains for the `advertised_pandaproxy_api`. It seems to advertise the same internal FQDN of the headless ClusterIP Service twice. I think one of them should advertise the given `<external.address>.<external.domain>` in the case of subdomains (`external.addressType: subdomain`)or just the `<external.address>` in the case of IP addresses (`external.addressType: ip`).

alejandroEsc:
Closing this in favor of https://github.com/redpanda-data/helm-charts/pull/254 since it has these issues as part of its overall solution.

PR #266, opened 2023-01-09
redpanda: add missed example command to tests, and fix command

### LINKED ISSUE #267: NOTES.txt template suggests rpk commands use unavailable --api-urls option when interacting with the cli
When working with the helm charts for the first time using the [docs](https://docs.redpanda.com/docs/platform/quickstart/kubernetes-qs-dev/) , the NOTES.txt template produces the following command as a suggested way of interacting with the cluster

```
kubectl -n redpanda exec -ti redpanda-0 -c redpanda -- rpk cluster info --brokers redpanda-0.redpanda.redpanda.svc.cluster.local.:9093,redpanda-1.redpanda.redpanda.svc.cluster.local.:9093,redpanda-2.redpanda.redpanda.svc.cluster.local.:9093 --api-urls redpanda-0.redpanda.redpanda.svc.cluster.local.:9644
```

It appears that the --api-urls flag is unavailable in 22.3, so not sure if this was ever needed.  Dropping that flag makes `rpk cluster info` work correctly

Digging through the code, it appears this is used in `_helpers.tpl` in the following code chunk, but I'm not sure if this is required elsewhere with other subcommands that may be produced in the various templates. 

```
{{- define "rpk-flags" -}}
  {{- $root := . -}}
  {{- $admin := list -}}
  {{- $admin = concat $admin (list "--api-urls" (include "api-urls" . )) -}}
  {{- if (include "admin-internal-tls-enabled" . | fromJson).bool -}}
    {{- $admin = concat $admin (list
      "--admin-api-tls-enabled"
      "--admin-api-tls-truststore"
      (printf "/etc/tls/certs/%s/ca.crt" .Values.listeners.admin.tls.cert))
    -}}
```

Best guess right now is that removing the following is sufficient to fixup the NOTES output, but needs eyes from someone more familiar with the code:

```
  {{- $admin = concat $admin (list "--api-urls" (include "api-urls" . )) -}}
```

joejulian:
Fixes #267

### DIFF
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 2.4.2
+version: 2.4.3
 
 # The app version is the default version of Redpanda to install.
 appVersion: v22.3.9
--- a/charts/redpanda/templates/_example-commands.tpl
+++ b/charts/redpanda/templates/_example-commands.tpl
@@ -33,7 +33,7 @@ and tested in a test.
 
 {{/* tested in tests/test-kafka-sasl-status.yaml */}}
 {{- define "rpk-cluster-info" -}}
-{{ .rpk }} cluster info {{ include "rpk-common-flags" . }}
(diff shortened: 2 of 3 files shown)


alejandroEsc (approved):
so slight naming mistake? lgtm

PR #268, opened 2023-01-10
feat: add preStart and postStop scripts

### LINKED ISSUE #262: Add preStop/postStart lifecycle hooks for managing maintenance mode
There are preStop and postStart scripts in the operator: https://github.com/redpanda-data/redpanda/blob/e51d5b7f2ef76d5160ca01b8c7a8cf07593d29b6/src/go/k8s/pkg/resources/secret.go

Duplicate that functionality in the helm chart.

It would involve generating the scripts in a secret, then mounting them in the sts and linking them to the sts [lifecycle](https://github.com/redpanda-data/redpanda/blob/54bccf3a29b173ef50134fbc128985cc7cf1fd11/src/go/k8s/pkg/resources/statefulset.go#L569).

alejandroEsc:
Work in progress:

Adding postStart and preStop scripts as found on the operator code. 

Fixes #262

### DIFF
--- a/charts/redpanda/templates/_helpers.tpl
+++ b/charts/redpanda/templates/_helpers.tpl
@@ -460,3 +460,23 @@ fsGroup: {{ dig "podSecurityContext" "fsGroup" .Values.statefulset.securityConte
 runAsUser: {{ dig "podSecurityContext" "runAsUser" .Values.statefulset.securityContext.runAsUser .Values.statefulset }}
 runAsGroup: {{ dig "podSecurityContext" "fsGroup" .Values.statefulset.securityContext.fsGroup .Values.statefulset }}
 {{- end -}}
+
+{{- define "tls-curl-flags" -}}
+  {{- $result := "" -}}
+  {{- if (include "tls-enabled" . | fromJson).bool -}}
+    {{- $path := (printf "/etc/tls/certs/%s" .Values.listeners.admin.tls.cert) -}}
+    {{- $result = (printf "--cacert %s/tls.crt" $path) -}}
+    {{- if .Values.listeners.admin.tls.requireClientAuth -}}
+      {{- $result = (printf "--cacert %s/ca.crt --cert %s/tls.crt --key %s/tls.key" $path $path $path) -}}
+    {{- end -}}
+  {{- end -}}
+  {{- $result -}}
+{{- end -}}
+
+{{- define "http-protocol" -}}
+  {{- $result := "http" -}}
+  {{- if (include "tls-enabled" . | fromJson).bool -}}
+    {{- $result = "https" -}}
+  {{- end -}}
+  {{- $result -}}
+{{- end -}}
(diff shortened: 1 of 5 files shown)


charts/redpanda/templates/statefulset.yaml
```diff
@@ -164,6 +163,19 @@ spec:
...
+          lifecycle:
+            postStart:
+              exec:
+                command:
+                  - "/bin/sh"
+                  - "-c"
+                  - "./var/lifecycle/postStart.sh"
+            preStop:
+              exec:
+                command:
```
joejulian, line 175: `sh` here, but `bash` in the scripts themselves. Let's just use bash for both.

charts/redpanda/templates/statefulset.yaml
```diff
@@ -164,6 +163,19 @@ spec:
...
+                  - "/bin/sh"
+                  - "-c"
+                  - "./var/lifecycle/preStop.sh"
```
joejulian, line 178: `sh` here, but `bash` in the scripts themselves. Let's just use bash for both.

joejulian (commented):
Looking good! Now we just have to figure out how to test this.

alejandroEsc:
> Looking good! Now we just have to figure out how to test this.

We could just try to run each command on one of the machines (this checks that they exist and that they run without error). The other tests that bring up and down a cluster will be enough to tell us if there is an issue with the scripts because any failure in the scripts will show up as unable to bring up a redpanda cluster or take down a redpanda cluster. Since the scripts depend on TLS and TLS related options they will definitely fail with broken changes, bad options, or unexpected changes to the logic.

joejulian:
> > Looking good! Now we just have to figure out how to test this.
> 
> We could just try to run each command on one of the machines (this checks that they exist and that they run without error). The other tests that bring up and down a cluster will be enough to tell us if there is an issue with the scripts because any failure in the scripts will show up as unable to bring up a redpanda cluster or take down a redpanda cluster. Since the scripts depend on TLS and TLS related options they will definitely fail with broken changes, bad options, or unexpected changes to the logic.

Yeah, that's probably good enough for now.

joejulian (approved):

charts/redpanda/templates/_helpers.tpl
```diff
@@ -460,3 +460,23 @@ fsGroup: {{ dig "podSecurityContext" "fsGroup" .Values.statefulset.securityConte
 runAsUser: {{ dig "podSecurityContext" "runAsUser" .Values.statefulset.securityContext.runAsUser .Values.statefulset }}
 runAsGroup: {{ dig "podSecurityContext" "fsGroup" .Values.statefulset.securityContext.fsGroup .Values.statefulset }}
 {{- end -}}
+
+{{- define "tls-curl-flags" -}}
+  {{- $result := "" -}}
+  {{- if (include "tls-enabled" . | fromJson).bool -}}
+    {{- $path := (printf "/etc/tls/certs/%s" .Values.listeners.admin.tls.cert) -}}
+    {{- $result = (printf "--cacert %s/tls.crt" $path) -}}
```
jan-g, line 468: Noting for posterity - this looks like a C&P error. Should be `ca.crt` here.

PR #272, opened 2023-01-12
#263 - workaround for k8s node taint and rack label having the same k…

### LINKED ISSUE #263: Improve rack awareness label discovery
Hello,

in our environment we are using taints on nodes with the same name as rack label. This is breaking this piece of code: https://github.com/redpanda-data/helm-charts/blob/main/charts/redpanda/templates/statefulset.yaml#L140
`RACK=$(curl --silent --cacert /run/secrets/kubernetes.io/serviceaccount/ca.crt --fail -H 'Authorization: Bearer '$(cat /run/secrets/kubernetes.io/serviceaccount/token) "https://${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT_HTTPS}/api/v1/nodes/${KUBERNETES_NODE_NAME}?pretty=true" | grep {{ .Values.rackAwareness.nodeAnnotation | quote | squote }} | grep -v '\"key\":' | sed 's/.*": "\([^"]\+\).*/\1/')`

As a workaround we have added additional grep:
`.... | grep {{ .Values.rackAwareness.nodeAnnotation | quote | squote }} | grep -v '\"key\":' | sed 's/.*": "\([^"]\+\).*/\1/'`

It would be better to do it using for example jq but it would need to be added to the image.

burandobata:
Workaround for k8s node taint and rack label having the same key name.
Fixes #263

### DIFF
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 2.10.4
+version: 2.10.5
 
 # The app version is the default version of Redpanda to install.
 appVersion: v22.3.13
--- a/charts/redpanda/templates/statefulset.yaml
+++ b/charts/redpanda/templates/statefulset.yaml
@@ -149,7 +149,7 @@ spec:
               {{- if (include "redpanda-atleast-22-3-0" . | fromJson).bool }}
                 {{- if .Values.rackAwareness.enabled }}
                 # Configure Rack Awareness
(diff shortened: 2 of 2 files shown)


joejulian (approved):

joejulian:
@burandobata If you could sign the CLA and rebase, we could get this merged.

joejulian:
Sorry, didn't see the comment in the issue about the CLA already being signed. I'll get this merged.

ivotron:
hi @burandobata , the problem is that the commit has a different email to the one you use for github. this needs to be reconciled so that the CLA check passes

joejulian:
rebased

RafalKorepta:
rebased

RafalKorepta (approved):

PR #275, opened 2023-01-12
redpanda: add console as subchart to redpanda

joejulian:


charts/redpanda/templates/configmap.yaml
```diff
@@ -351,3 +351,60 @@ data:
...
+{{- range (include "seed-server-list" . | mustFromJson) }}
+  {{ $urls = append $urls (printf "%s://%s:%d" $proto . $port) }}
+{{- end }}
+{{/* add sasl */}}
+{{/* add tls */}}
+{{ $schemaRegistry := dict "enabled" true "urls" $urls }}
+{{ $consoleConfig := dict "kafka" (dict "brokers" $brokers "schemaRegistry" $schemaRegistry) }}
+{{ $config := dict "Values" (dict "console" (dict "config" $consoleConfig)) }}
+{{ $console := merge $config $console }}
+{{ tpl (.Subcharts.console.Files.Get "files/console_configmap.yaml") $console }}
```
alejandroEsc, line 410: very nice.

charts/redpanda/templates/configmap.yaml
```diff
@@ -351,3 +351,60 @@ data:
...
+
+  {{- if .Values.console.roleBindings }}
+  role-bindings.yaml: |
+    roleBindings:
+      {{- tpl (toYaml .Values.console.roleBindings) $ | nindent 6 }}
+  {{- end }}
+{{- end }}
+*/}}
+---
+# console config
```
alejandroEsc, line 383: Is it possible to move this to a helper file and just "get" the configmap from there?

alejandroEsc (commented):
Looks awesome so far, very clever changes, one comment/question but nothing blocking so far.

charts/redpanda/templates/configmap.yaml
joejulian: Still thinking about placement. This section is more-or-less is a helper file.

joejulian: I'm hesitant because it creates a second layer of indirection where we already have one layer with the configmap file from the subchart.

alejandroEsc (approved):
lgtm!

alejandroEsc:
merging this as the last commit was me removing a file that was accidentally pushed. The PR was tested both locally and CI and passed.

PR #277, opened 2023-01-17
console: set service targetPort to config listenPort

adrianmester:
The service `targetPort` is currently set to `.Values.service.targetPort` which doesn't exist in the values file, so it ends up as `null` unless overwritten.

Not sure many people change the listenPort in the config file, but since that value is [already used](https://github.com/redpanda-data/helm-charts/blob/acec1790fec99f320e08a8385ece84817bc73c17/charts/console/templates/deployment.yaml#L60) as the `containerPort` in the deployment, it should also be used here.

### DIFF
--- a/charts/console/Chart.yaml
+++ b/charts/console/Chart.yaml
@@ -27,7 +27,7 @@ type: application
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
 
 # Chart versions do not track appVersion
-version: 0.4.0
+version: 0.4.1
 
 # The app version is the version of the Chart application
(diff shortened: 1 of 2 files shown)


joejulian:
@adrianmester This will need a chart version bump as part of the PR. Just rebase on main and bump it a patch version. Thank you!

adrianmester:
> @adrianmester This will need a chart version bump as part of the PR. Just rebase on main and bump it a patch version. Thank you!

Done! Anything else I need to do to be able to merge?

joejulian (approved):

PR #278, opened 2023-01-17
Add tiered storage configuration

RafalKorepta:
When the Kafka client fetches an offset range that isn’t available locally in the Redpanda data directory, Redpanda downloads remote segments from cloud storage. These downloaded segments are stored in the cloud storage cache directory. The cache directory might be created in the Redpanda data directory, but iops will be split among cache and hot segments. In the case where performance and latency is important separated local storage should be used for cache directory.

The tiered storage mimic Redpanda data dir configuration.

Tiered storage is guarded by the presence of license_key or license_secret_ref.


charts/redpanda/values.yaml
```diff
@@ -252,6 +271,120 @@ storage:
...
+    cloud_storage_enabled: false
+    # Cluster wide config option to enable data upload from Redpanda to cloud storage if set to true.
+    cloud_storage_enable_remote_write: true
+    # Cluster wide config option to enable Redpanda to fetch data from cloud storage if set to true.
+    cloud_storage_enable_remote_read: true
+    # This configuration option is available starting from 22.3.X
+    #
+    # Possible values:
+    # `config` - If IAM roles are not available, specify credentials in the cluster configuration file.
+    # `aws_instance_metadata` - For an AWS EC2 instance, use the instance metadata API from AWS.
```
jcsp, line 301: I'm concerned about all this copy-paste.  How is this going to stay up to date when we evolve the configuration options, e.g. adding new possible values to this one?

If we decide as a team that maintaining a copy of the docs in these .yaml files makes sense then it's doable, but it needs thought, and some kind of process similar to how the docs team tracks new features and makes updates for them.

charts/redpanda/values.yaml
```diff
@@ -252,6 +271,120 @@ storage:
...
+    # API endpoint.
+    #
+    # - For AWS, this can be left blank. It’s generated automatically using the region and bucket.
+    #
+    # - For Google Cloud Service, use storage.googleapis.com.
+    cloud_storage_api_endpoint: ""
+
+    # Maximum size of the disk cache used by Tiered Storage.
+    # Default is 20 GiB.
+    cloud_storage_cache_size: 20GiB
```
jcsp, line 324: Does helm have knowledge of the size of drives?  If so, it should do something dynamic here, ideally coordinating the partition-per-shard limit with the disk size.

If not, then helm should not specify a value here, so that we can evolve the default in core.

charts/redpanda/values.yaml
```diff
@@ -252,6 +271,120 @@ storage:
...
+    # The directory for the Tiered Storage cache. You must specify the full path. Default is: <redpanda-data-directory>/cloud_storage_cache.
+    # cloud_storage_cache_directory: ""
+    # The time, in milliseconds, between cache checks. The size of the cache can grow quickly,
+    # so it’s important to have a small interval between checks, but if the checks are too frequent,
+    # they consume a lot of resources.
+    # Default is 30000 ms.
+    cloud_storage_cache_check_interval: 30000
```
jcsp, line 331: For any setting that you're not setting to something non-default, please do not set it unless there is a strong reason to do so.  Letting redpanda handle the defaulting means we can evolve it in the future, whereas explicitly setting values will pin them forever to whatever the default in the helm chart was when the system was installed.

charts/redpanda/values.schema.json
```diff
@@ -321,6 +324,114 @@
               "type": "object"
             }
           }
+        },
+        "tieredStoragePersistentVolume": {
```
jcsp, line 328: I just learned that this schema exists.  I'm concerned about the maintenance burden of keeping it in sync.

To help folks doing operator/cloud, we added a schema to Redpanda builds for using in situations just like this, e.g.: https://cloudsmith.io/~redpanda/repos/redpanda/packages/detail/raw/redpanda-configuration-schema/22.3.11/

It's in a swagger-like format rather than json-schema, but you could probably convert it pretty easily, and never have to edit this by hand again (yay!)

charts/redpanda/templates/statefulset.yaml
```diff
@@ -300,6 +314,18 @@ spec:
...
+{{- if and (or (not (empty .Values.license_key)) (not (empty .Values.license_secret_ref)) .Values.storage.tieredConfig.cloud_storage_enabled) (hasKey .Values.storage.tieredConfig "cloud_storage_cache_directory") (not (empty .Values.storage.tieredConfig.cloud_storage_cache_directory)) }}
+        - name: tiered-storage-dir
+{{- if .Values.storage.tieredStoragePersistentVolume.enabled }}
+          persistentVolumeClaim:
+            claimName: tiered-storage-dir
+{{- else if .Values.storage.tieredStorageHostPath }}
+          hostPath:
+            path: {{ .Values.storage.tieredStorageHostPath | quote }}
+{{- else }}
+          emptyDir: {}
```
jcsp, line 327: `emptyDir` is a risky default here.  The tiered storage cache size is currently 20GiB (it's not very smart, the assumption is that the thing deploying will have opinions), and if you run that on a platform that uses 1GB emptyDirs then it could well end up asserting out and/or misbehaving.

We do also use that as the default for redpanda data dirs, but: I think people are more likely to intuitively understand that using emptyDir for their data dirs isn't going to work, whereas for a cache dir they might reasonably think this is a usable configuration.

charts/redpanda/values.schema.json
jcsp: Thinking a bit further... you could probably also generate some of values.yaml this way.  What you get from "rpk cluster config edit" (inlined docs and all) is all just built from the redpanda config schema -- this could be your path to having those inline docs without having to deal with maintaining them.

charts/redpanda/values.yaml
mattschumpert: Agreed. We shouldn't embed config docs within Helm but instead link straight to the platform docs for description.

RafalKorepta: > I'm concerned about all this copy-paste. How is this going to stay up to date when we evolve the configuration options, e.g. adding new possible values to this one?

Once the Redpanda make official release with the appropriate docs update, then K8S team would pick the new Redpanda version and adjust comment with the Redpanda container tag (Redpanda version).

> If we decide as a team that maintaining a copy of the docs in these .yaml files makes sense then it's doable, but it needs thought, and some kind of process similar to how the docs team tracks new features and makes updates for them.

Release notes, from the Redpanda release, should drive changes in this values.yaml file.

> Agreed. We shouldn't embed config docs within Helm but instead link straight to the platform docs for description.

If users start from [artifact hub](https://artifacthub.io/packages/helm/redpanda-data/redpanda), then it add friction as user would need to jump into many docs to correctly configure Redpanda cluster. This is only tiered storage section, but currently we have 3 different link to our docs. This PR will add 3 more.

I'm used to have this `value.yaml` definition included in artifact hub similar to https://artifacthub.io/packages/helm/softonic/kyverno#values

We don't generate such table of content, but we could use [values modal](https://artifacthub.io/packages/helm/redpanda-data/redpanda?modal=values) to show what is available to customers. That way we can include as many description as we need. 

Another option is to create automation that would generate [values.yaml](https://github.com/redpanda-data/helm-charts/blob/main/charts/redpanda/values.yaml) from the [values.schema.json](https://github.com/redpanda-data/helm-charts/blob/main/charts/redpanda/values.schema.json), but each field needs to have description. You can see in [values schema modal](https://artifacthub.io/packages/helm/redpanda-data/redpanda?modal=values-schema) that not many fields have description. As per https://github.com/redpanda-data/helm-charts/pull/278#discussion_r1072799550 comment.

I'm trying to show you that there would be friction for users if they are discovering Redpanda from artifiact hub.

-------

To add more I included constraint for this particular filed in https://github.com/redpanda-data/helm-charts/pull/278/files#diff-7573c8e3ada7cb2fe4a9a6b40e135a1d5568a9d128d15a57b124d32568126819R368.

That way user would know the `cloud_storage_credentials_source` field have only 4 available options. The command `helm install` or `helm upgrade` will reject invalid input. It's much better user experience.

Example:
```
helm template redpanda charts/redpanda --set storage.tieredConfig.cloud_storage_enabled=true --set image.tag=v22.1.1 --set license_key=test --set storage.tieredConfig.cloud_storage_credentials_source=test
redpanda:
- storage.tieredConfig.cloud_storage_credentials_source: Does not match pattern '^(config|aws_instance_metadata|sts|gcp_instance_metadata)$'
```

If we are making some effort to add input validation, why we should not follow up with clear description?

charts/redpanda/values.yaml
RafalKorepta: > Does helm have knowledge of the size of drives? 

Yes, it does. This exact property will drive the PersistentVolumeClaim which has size included. 
https://github.com/redpanda-data/helm-charts/blob/c4bbf8a34910bf4b0c984340a3688405024afb6d/charts/redpanda/templates/statefulset.yaml#L464

> If so, it should do something dynamic here, ideally coordinating the partition-per-shard limit with the disk size.

I'm happy to implement some logic here to drive partition-per-shard limit. Could you share more concrete example or algorithm that calculates it? 

> At some point we may change to not having an explicit cap by default, and if helm systems have all got an explicit 20GiB they'll get stuck that way.

If that happen we will update `values.yaml` accordingly to the new requirements. The volume size will always be necessary if user would like to configure persistent volumes. It can be driven from another place like Redpanda data volume size is defined https://github.com/redpanda-data/helm-charts/blob/c4bbf8a34910bf4b0c984340a3688405024afb6d/charts/redpanda/values.yaml#L262

charts/redpanda/values.yaml
RafalKorepta: Given my previous comment https://github.com/redpanda-data/helm-charts/pull/278#discussion_r1073454630 if I don't list the options users could not discover them easily if they start from artifact hub. 

To add more I mentioned release for Redpanda should have enough docs or release notes for k8s team to pick it up and adjust any property in helm.

Core team can evolve it in future I don't see any problem here. I would rather be always prescriptive.

charts/redpanda/values.schema.json
RafalKorepta: > To help folks doing operator/cloud, we added a schema to Redpanda builds for using in situations just like this, e.g.: https://cloudsmith.io/~redpanda/repos/redpanda/packages/detail/raw/redpanda-configuration-schema/22.3.11/

That's great! My concerns is you provided schema for particular Redpanda configuration version where helm is about to support multiple Redpanda version in Kubernetes environment, so it needs to:

- drop some properties in older versions
- generate `bootstrap.yaml` config for specific range of Redpanda versions
- set properties at runtime not only as properties defined in `bootstrap.yaml` and `redpanda.yaml`
- support properites that are tight to Kuberentes (statefulset, ingress, DNS, TLS secrets etc.)

> Thinking a bit further... you could probably also generate some of values.yaml this way.

Yes, some of the values yes, but it would require custom tool. For me manual copy and paste few properties is easier. At least this PR didn't trigger in me the need for automation 🙃 

> this could be your path to having those inline docs without having to deal with maintaining them.

True. As new feature in this repo defiantly worth considering.

charts/redpanda/templates/statefulset.yaml
RafalKorepta: > emptyDir is a risky default here. The tiered storage cache size is currently 20GiB (it's not very smart, the assumption is that the thing deploying will have opinions), and if you run that on a platform that uses 1GB emptyDirs then it could well end up asserting out and/or misbehaving.

True, but I'm about to remove the explicit default and remove the comments, so where I should mention this caveat?

> We do also use that as the default for redpanda data dirs, but: I think people are more likely to intuitively understand that using emptyDir for their data dirs isn't going to work, whereas for a cache dir they might reasonably think this is a usable configuration.

The nice think about kubernetes is if user configure `emptyDir` for tiered storage cache it could end up running out of disk space. This event is reported by kubelet and node evicts some pods. If user create auto scaling node group, then VM should be in degraded state, which would trigger new instance creation. During that time Redpanda cluster will be in degraded state. Redpanda would picked Redpanda ID from data volume if it's backed by persistent disk, or generate new ID. Helm doesn't have ways to detect that Redpanda lost whole data folder, so no decommission of Redpanda node/broker would happen. That way, I would argue, we can not prevent from miss configuration. I'm not sure if documentation would be enough to prevent this. 

----

I did more and found emptyDir can have sizeLimit property set to prevent from your scenario. I will add this.

REF: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#emptydirvolumesource-v1-core

charts/redpanda/values.yaml
joejulian: Yes having a well documented values.yaml is [considered best practice](https://helm.sh/docs/chart_best_practices/values/#document-valuesyaml) within the helm community. A poorly documented values.yaml, or a link to external documentation will create an endless stream if issue requests asking for it to be documented. This has been tried in helm charts for other popular software and did not go over well.

charts/redpanda/values.yaml
```diff
@@ -252,6 +271,120 @@ storage:
...
+  # REF: https://docs.redpanda.com/docs/manage/tiered-storage/#enable-tiered-storage-for-a-topic
+  tieredConfig:
+    # Global flag that enables Tiered Storage feature if license is provided.
+    # Default is false.
+    cloud_storage_enabled: false
+    # Cluster wide config option to enable data upload from Redpanda to cloud storage if set to true.
+    cloud_storage_enable_remote_write: true
+    # Cluster wide config option to enable Redpanda to fetch data from cloud storage if set to true.
+    cloud_storage_enable_remote_read: true
+    # This configuration option is available starting from 22.3.X
```
alejandroEsc, line 297: If the comment is referring the the previous configuration, let's move it above the config so that it is clear.

charts/redpanda/values.yaml
alejandroEsc: In most ideal situations we can give users an idea of how to calculate that based on emperical data (metrics, previous experience with customers etc...) If we have some input today, this is the best time to go write it down and provide an way to calculate but give users the option to override (by making this an override) our opinionated solution.

charts/redpanda/values.yaml
alejandroEsc: agreed, let's comment/document the defaults and make it explicit that we can override defaults.

alejandroEsc (commented):
lgtm, just a few comments but i think simplifying some of the templating is my only major ask for this pr. Thank you!

charts/redpanda/values.yaml
RafalKorepta: ok

charts/redpanda/templates/statefulset.yaml
```diff
@@ -67,6 +67,16 @@ spec:
           volumeMounts:
             - name: datadir
               mountPath: /var/lib/redpanda/data
+{{- if and (or (not (empty .Values.license_key)) (not (empty .Values.license_secret_ref)) .Values.storage.tieredConfig.cloud_storage_enabled) (hasKey .Values.storage.tieredConfig "cloud_storage_cache_directory") (not (empty .Values.storage.tieredConfig.cloud_storage_cache_directory)) }}
```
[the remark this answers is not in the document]
RafalKorepta: it will not fail if we will always have `cloud_storage_cache_directory` included in our `values.yaml` file. I'm not so used to it, so I will change it.

charts/redpanda/values.yaml
RafalKorepta: This lines refers to https://github.com/redpanda-data/helm-charts/blob/1e53583ba306ff64d073a86e05e1ce0321b49a91/charts/redpanda/values.yaml#L306

I will change the comment to replace word `this` with `cloud_storage_credentials_source`

RafalKorepta:
I found that upgrades will be not possible with this version. I will write the release notes in the `README.md` how to enable tiered storage cache via manual update.

```
helm upgrade --install --namespace redpanda --create-namespace redpanda charts/redpanda --set storage.tieredConfig.cloud_storage_enabled=true --set image.tag=v22.3.10 --set license_key=test --set storage.tieredConfig.cloud_storage_cache_directory=/some-path/ --set storage.tieredStoragePersistentVolume.enabled=true --set storage.tieredConfig.cloud_storage_cache_size=20Gi
Error: UPGRADE FAILED: cannot patch "redpanda" with kind StatefulSet: StatefulSet.apps "redpanda" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', 'updateStrategy', 'persistentVolumeClaimRetentionPolicy' and 'minReadySeconds' are forbidden
```

charts/redpanda/values.yaml
```diff
@@ -252,6 +271,120 @@ storage:
...
+    # Possible values:
+    # `config` - If IAM roles are not available, specify credentials in the cluster configuration file.
+    # `aws_instance_metadata` - For an AWS EC2 instance, use the instance metadata API from AWS.
+    # `sts` - For AWS on Kubernetes, use the Secure Token Service (STS).
+    # `gcp_instance_metadata` - For a VM running on GCP, or for Google Kubernetes Engine (GKE), use the instance metadata API from GCP.
+    #
+    # REF: https://docs.redpanda.com/docs/manage/security/iam-roles/#configuring-iam-roles
+    cloud_storage_credentials_source: config
+    # Cloud storage region. Required.
+    cloud_storage_region: ""
```
vuldin, line 308: Users may have the expectation that by enabling tiered storage and adding a bucket name in `values.yaml`, tiered storage will be enabled. But these changes also require the user to create the bucket, since we aren't creating the bucket via the helm chart.

Maybe we could add a comment (either close to this variable or at the start of the tiered storage config section) that mentioned the prerequisite step of creating a bucket.

charts/redpanda/values.yaml
```diff
@@ -252,6 +271,120 @@ storage:
...
+    # Cloud storage bucket name. Required.
+    cloud_storage_bucket: ""
+    # Cloud storage access key used to authenticated if `cloud_storage_credentials_source` is set to `config`
+    # cloud_storage_access_key: ""
+    # Cloud storage secret key used to authenticated if `cloud_storage_credentials_source` is set to `config`
+    # cloud_storage_secret_key: ""
```
vuldin, line 314: I know at one point not too long ago these values needed to be set (even when using `aws_instance_metadata` for instance), is this still the case? If that is still the case, then should we ensure these keys have values?

charts/redpanda/values.yaml
```diff
@@ -236,6 +236,25 @@ storage:
   # If not specified, then `emptyDir` will be used instead.
   # If specified, but `persistentVolume.enabled` is `true`, then has no effect.
   hostPath: ""
+  # Absolute path on host to store Redpanda's tiered storage cache data.
+  # If not specified, then `emptyDir` will be used instead.
+  # If specified, but `tieredStoragePersistentVolume.enabled` is `true`, then has no effect.
+  tieredStorageHostPath: ""
+  # If `enabled` is `true` then a PersistentVolumeClaim will be created for tiered storage cache and
+  # used to store data retrieved from cloud storage (eg. S3), otherwise `hostPath` is used.
+  tieredStoragePersistentVolume:
```
vuldin, line 245: We have the `storage.tieredStoragePersistentVolume`, and then we also have `storage.persistentVolume.tieredConfig`. Would it not make sense to move all `storage.persistentVolume.tieredConfig` into `storage.tieredStoragePersistentVolume`?

vuldin (commented):
A few questions on new parameter layout in `values.yaml` and tiered storage default config.

charts/redpanda/values.yaml
RafalKorepta: In this PR there is no `storage.persistentVolume.tieredConfig`, but rather `storage.tieredConfig`. The `storage.tieredConfig` is representation of Redpanda tiered storage configuration. If we move `storage.tieredConfig` under `storage.tieredStoragePersistentVolume` then it would be miss leading as you could create ephemeral tiered storage cache volume.

charts/redpanda/values.yaml
RafalKorepta: Sure I can add the prerequisite step, but this will mimic our doc. This contradict with other review and I'm not sure what your comment is about this thread https://github.com/redpanda-data/helm-charts/pull/278#discussion_r1072782702

vuldin: I just expect that users will assume they can enable tiered storage by making modifications in `values.yaml`, and if we don't let them know that's not all they need to do by some comment in this file, then there will be users who run into an error around tiered storage hopefully will reach out via some channel for help rather than just deciding they either 1) don't need tiered storage or 2) don't need Redpanda.

I'll make a comment in the thread you link to.

charts/redpanda/values.yaml
vuldin: This thread came up [here](https://github.com/redpanda-data/helm-charts/pull/278#discussion_r1081344609). I don't see a huge problem with copying in details that can already be found in our docs, nor with linking to docs. I'm fine with either of these two options.

But at least one of them is needed, as users need to have the information available in `values.yaml` to successfully deploy Redpanda, as this file could likely be the first source of information on deploying Redpanda that some users have.

vuldin: I think having a well-defined `values.yaml` will become more important in the future once Redpanda readily available in other environments like OpenShift.

jcsp: Please bear in mind that the core team's release notes are not intended to be at a level of detail that would enable accurately maintaining a copy of docs for tunables.  Picking an example from this PR, if we changed cloud_storage_initial_backoff_ms's default for some reason, I would probably not release note that.

I think we're likely to get a higher quality outcome by using the json schema to do an update of this stuff at each major feature release, rather than sifting it out of release notes and updating by hand.  Ultimately the docs on the website will work this way too, for things like keeping the defaults etc in sync.

charts/redpanda/values.yaml
jcsp: > I'm happy to implement some logic here to drive partition-per-shard limit. Could you share more concrete example or algorithm that calculates it?

Probably makes sense to make it work the same way that the cloud team does, although they only have to worry about volume sizes of particular instance types.  For arbitrary volume sizes, you'd need to tune the partition limit, segment size + cache size together.  This is a bit too subtle for a PR comment thread: let's have a chat about it offline.

charts/redpanda/values.yaml
joejulian: I think I understand what you're suggeting, @jcsp, but would you please create an issue so we can track that?

charts/redpanda/values.yaml
joejulian: > > I'm happy to implement some logic here to drive partition-per-shard limit. Could you share more concrete example or algorithm that calculates it?
> 
> Probably makes sense to make it work the same way that the cloud team does, although they only have to worry about volume sizes of particular instance types. For arbitrary volume sizes, you'd need to tune the partition limit, segment size + cache size together. This is a bit too subtle for a PR comment thread: let's have a chat about it offline.

Is that a prerequisite for merging this PR, or just a general principal we should work toward?

charts/redpanda/values.yaml
RafalKorepta: This is interesting, but our docs https://docs.redpanda.com/docs/manage/tiered-storage doesn't mention that bucket needs to be created before any configuration changes.

charts/redpanda/templates/statefulset.yaml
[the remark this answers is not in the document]
RafalKorepta: I implemented the default template, so that this condition is now removed.

charts/redpanda/values.yaml
RafalKorepta: I can confirm that version 22.3.10 works without `cloud_storage_secret_key` nor `cloud_storage_access_key`

charts/redpanda/README.md
```diff
@@ -34,6 +34,37 @@ To learn how to override the default values in the `values.yaml` file, see the [
...
+kubectl delete sts redpanda --cascade=orphan
+```
+
+The `helm upgrade` should be able to succeed, but you need to manually do rolling update starting from ordinal 0 to ordinal N.
+
+```bash
+kubectl delete pod redpanda-0
+```
+
+Please wait for deleted Pod to be restarted and become ready in order to move to next Pod.
```
vuldin, line 66: Thanks for working out a way to perform these types of updates for a helm deployment. But how will recommending user delete the statefulset change their ongoing interactions with the cluster? I assume that from now on they will need to manually perform rolling updates from this point forward?

RafalKorepta: Not at all. If user want to have tiered storage cache backed by persistent volume, then user would be required to perform manual deletion of each pod to roll out new pod definition. After that all automation build in statefulset and our post install helm hook jobs will run as always. Nothing else will be manual.

charts/redpanda/templates/configmap.yaml
```diff
@@ -50,6 +50,15 @@ data:
   {{- if and (not (hasKey .Values.config.cluster "storage_min_free_bytes")) ((include "redpanda-atleast-22-2-0" . | fromJson).bool) }}
     storage_min_free_bytes: {{ include "storage-min-free-bytes" . }}
   {{- end }}
+{{- end }}
+{{- if and (include "is-licensed" .) .Values.storage.tieredConfig.cloud_storage_enabled }}
```
joejulian, line 54: "is-licensed" will return a string: "true" or "false" which will always test as true. This is why we use the fromJson/bool strategy.

joejulian (changes requested):
`{{ if "false" }}` will succeed as true because only empty strings are false. Named templates always return strings. We work around that by following the `( include ... | fromJson).bool` strategy.

charts/redpanda/templates/configmap.yaml
RafalKorepta: You are right. I follow your suggestion

joejulian (approved):
lgtm

alejandroEsc (approved):
lgtm so far, some actions should come out of this. Do you mind walking a demo with me sometime?

RafalKorepta:
Yes, please ping me and I can do ad-how presentation.

charts/redpanda/values.yaml
jcsp: >Is that a prerequisite for merging this PR, or just a general principal we should work toward?

I'm late to the party (the PR already merged), but yes, I think this should have been a blocking issue.

As soon as this is merged + visible to users, we will start putting clusters in the field with an explicit 20GiB limit, which can be too big for the disk, and can prevent Redpanda from migrating smoothly in future as/when we change the default to something smarter than a fixed number.

PR #279, opened 2023-01-18
fix: add external.domain to SAN list

### LINKED ISSUE #269: Custom domains not included in SAN list for self-signed certificates
Steps to reproduce:

1. Configure an external domain and enable TLS with self-signed certificates:
    ```bash
    helm upgrade --install redpanda redpanda/redpanda -n redpanda --create-namespace  --set external.domain=example.com --set tls.enabled=true
    ```
1. Export the root CA certificate:
    ```bash
    kubectl -n redpanda get secret redpanda-default-root-certificate -o go-template='{{ index .data "ca.crt" | base64decode }}' > ca.crt
    ```
1. Try to connect to the cluster from outside Kubernetes:
    ```bash
    rpk cluster info --brokers redpanda-0.example.com:31092 --tls-truststore ca.crt
    ```
    
Result:

```
unable to request metadata: unable to dial: x509: certificate is valid for redpanda-cluster.redpanda.redpanda.svc.cluster.local, redpanda-cluster.redpanda.redpanda.svc, redpanda-cluster.redpanda.redpanda, *.redpanda-cluster.redpanda.redpanda.svc.cluster.local, *.redpanda-cluster.redpanda.redpanda.svc, *.redpanda-cluster.redpanda.redpanda, redpanda.redpanda.svc.cluster.local, redpanda.redpanda.svc, redpanda.redpanda, *.redpanda.redpanda.svc.cluster.local, *.redpanda.redpanda.svc, *.redpanda.redpanda, not redpanda-0.example.com
```

alejandroEsc:
Should fix #269

### DIFF
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 2.5.0
+version: 2.5.1
 
 # The app version is the default version of Redpanda to install.
 appVersion: v22.3.10
--- a/charts/redpanda/templates/certs.yaml
+++ b/charts/redpanda/templates/certs.yaml
@@ -3,6 +3,7 @@
   {{- $ns := .Release.Namespace -}}
   {{- $domain := .Values.clusterDomain | trimSuffix "." -}}
   {{- $listeners := .Values.listeners -}}
+  {{- $values := .Values }}
   {{- range $name, $data := .Values.tls.certs }}
     {{- $d := $data.duration }}
 ---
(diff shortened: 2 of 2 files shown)


JakeSCahill:
Tested this with the same steps as the original issue and it works.

Thanks for the fix @alejandroEsc 🎉

RafalKorepta (approved):

RafalKorepta:
Just FYI:

This definitely will solve self sign certificate issue, but if user would like to use ACME (Let's encrypt ) issuer the following SANs will prevent for creating certificate:
https://github.com/redpanda-data/helm-charts/blob/fb446c470745ca4887fec1f4169ada5aef71cd35/charts/redpanda/templates/certs.yaml#L17-L28

external issuer reference:
https://github.com/redpanda-data/helm-charts/blob/fb446c470745ca4887fec1f4169ada5aef71cd35/charts/redpanda/templates/certs.yaml#L40-L42

PR #289, opened 2023-01-20
fix: certificate issues

alejandroEsc:
Here I am fixing a few issues I have found so far that may be causing @JakeSCahill from being able to properly create a certificate with ACME.

Part of #203

### DIFF
--- a/charts/redpanda/templates/statefulset.yaml
+++ b/charts/redpanda/templates/statefulset.yaml
@@ -209,19 +209,21 @@ spec:
                 - -c
                 {{- if (include "admin-internal-tls-enabled" . |fromJson).bool }}
                 - >
-                  curl https://localhost:{{ .Values.listeners.admin.port }}/v1/cluster/health_overview
-                  -svk --cacert /etc/tls/certs/{{ .Values.listeners.admin.tls.cert }}/ca.crt |
-                  awk '{
-                    id = $0; gsub(/.*"controller_id": /, "", id); gsub(/,.*/, "", id)
-                    nd_str = $0; gsub(/.*"nodes_down": \[/, "", nd_str); gsub(/\].*/, "", nd_str)
-                    FS=","
-                    split(nd_str, nd_list)
-                    for (i in nd_list) nodes_down[nd_list[i]]=""
-                    exit (id in nodes_down)
-                  }'
+                  PROTO="https";
+                  CURL_FLAGS="-svk";
+                  {{- $cert := get .Values.tls.certs .Values.listeners.admin.tls.cert }}
+                  {{- if $cert.caEnabled }}
+                  CA_CERT_FLAG={{ printf "--cacert /etc/tls/certs/%s/ca.crt" .Values.listeners.admin.tls.cert | quote }};
+                  {{- else }}
+                  CA_CERT_FLAG="";
+                  {{- end }}
                 {{- else }}
                 - >
-                  curl -sv http://localhost:{{ .Values.listeners.admin.port }}/v1/cluster/health_overview |
+                  PROTO="http";
+                  CURL_FLAGS="-sv";
+                  CA_CERT_FLAG="";
+                {{- end }}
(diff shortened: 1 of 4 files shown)


charts/redpanda/templates/certs.yaml
```diff
@@ -26,7 +26,10 @@ spec:
   - {{ printf "*.%s.%s.svc.%s" $service $ns $domain | quote }}
   - {{ printf "*.%s.%s.svc" $service $ns | quote }}
   - {{ printf "*.%s.%s" $service $ns | quote }}
```
joejulian, line 28: wrap this in `{{- if not (empty $data.issuerRef) }}`

joejulian:
Also remove `commonName` https://github.com/redpanda-data/helm-charts/pull/289/files#diff-ec07f07a47aa911db2b476d9fcdd33907646aadab5e83db1e9aa13d36ab671e5R37

charts/redpanda/templates/statefulset.yaml
```diff
@@ -232,12 +232,15 @@ spec:
...
                 - -c
                 {{- if (include "admin-internal-tls-enabled" . |fromJson).bool }}
                 - >
-                  curl https://localhost:{{ .Values.listeners.admin.port }}/v1/cluster/health_overview
-                  -svk --cacert /etc/tls/certs/{{ .Values.listeners.admin.tls.cert }}/ca.crt
+                  CERT_FLAGS=""
+                  if [ -f /etc/tls/certs/{{ .Values.listeners.admin.tls.cert }}/ca.crt ]; then
+                      CERT_FLAGS={{ printf "--cacert /etc/tls/certs/%s/ca.crt .Values.listeners.admin.tls.cert | quote }}
+                  fi
+                  curl https://localhost:{{ .Values.listeners.admin.port }}/v1/cluster/health_overview -svk ${CERT_FLAGS}
```
joejulian, line 243:
```suggestion
CERT_FLAGS=""
{{- $cert := get .Values.listeners.admin.tls.cert .Values.tls.certs }}
{{- if $cert.caEnabled }}
CERT_FLAGS={{ printf "--cacert /etc/tls/certs/%s/ca.crt .Values.listeners.admin.tls.cert | quote }}
{{- end }}
```

JakeSCahill:
Thanks @alejandroEsc!

It looks like `caEnabled` is also ignored.

I'm seeing a root certificate being generated even if I specify `tls.cert.redpanda.caEnabled=false`.

This is causing a failed Certificate to be generated that is looking for an Issuer called `redpanda-redpanda-self-signed-issuer`:

``` 
Normal  IssuerNotFound      12m   cert-manager-certificaterequests-issuer-selfsigned  Referenced "Issuer" not found: issuer.cert-manager.io "redpanda-redpanda-selfsigned-issuer" not found
```

JakeSCahill:
I'm now able to get TLS working with Let's Encrypt and a custom domain 🎉 thanks @alejandroEsc !

This isn't a blocker, but I still get this disused root Certificate being created: https://github.com/redpanda-data/helm-charts/pull/289#issuecomment-1404934499

alejandroEsc:
> redpanda       redpanda-default-root-certificate         True    redpanda-default-root-certificate         15m

What I am seeing is what I was hoping not to see, so we may have to change a few things about certs but when i compute the values file from what you are doing, and even when i use my own values file, i get the following:

```
...
tls:
  certs:
    default:
      caEnabled: true
    redpanda:
      caEnabled: false
      issuerRef:
        kind: ClusterIssuer
        name: letsencrypt-prod
  enabled: true
tuning: {}
...
```

This is why the default cert is created. Ill investigate this a bit further.

joejulian:
This is a breaking change and I think it's still broken if someone uses a different issuer for the admin cert. We need to find a way to test this with different issuers.

alejandroEsc:
Closing this PR. This will fix the issues we see when trying to use certs meant for external clients internally.

PR #291, opened 2023-01-23
rp: add required resources to clusterrole for rpk debug bundle

r-vasquez:
We are introducing rpk debug bundle for k8s environments in 23.1, this functionality depends on the k8s API and it needs to be able to 'get' the listed resources of this commit.

Here is where we use them: https://github.com/redpanda-data/redpanda/blob/dev/src/go/rpk/pkg/cli/cmd/debug/bundle/bundle_k8s_linux.go#L265-L274

That means that rpk debug bundle will only work if you have `Values.rbac.enabled=true`

### DIFF
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 2.6.1
+version: 2.6.2
 
 # The app version is the default version of Redpanda to install.
 appVersion: v22.3.10
--- a/charts/redpanda/templates/rbac.yaml
+++ b/charts/redpanda/templates/rbac.yaml
@@ -43,6 +43,42 @@ rules:
       - list
 ---
 apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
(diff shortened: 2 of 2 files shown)


joejulian (changes requested):
For auditability, I'd prefer to have these in their own clusterrole.

r-vasquez:
@joejulian Should I use the same flag (rbac.enabled) from the Values.yaml ? I guess this is something that shouldn't be enabled by default.

joejulian:
> @joejulian Should I use the same flag (rbac.enabled) from the Values.yaml ? I guess this is something that shouldn't be enabled by default.

Yes. Should we add a subtree to rbac? Maybe `rbac.rpk_debug.enable` or something like that?

r-vasquez:
Rebased and created a separate clusterrole : {redpanda.fullname}-rpk-bundle but reused the same values.rbac.enabled.

joejulian (approved):
lgtm

r-vasquez:
[Force Push 1:](https://github.com/redpanda-data/helm-charts/compare/6902fd696d2c1cbdf2e7c4003ea9974f0926f1a8..da2839f43c7c14d1024727e36ae1a94579c04765) Version Bump

[Force Push 2:](https://github.com/redpanda-data/helm-charts/compare/da2839f43c7c14d1024727e36ae1a94579c04765..63742847638fdf40363b215baa4d2ed0c5eaf51a) Rebase with main.

PR #292, opened 2023-01-24
redpanda: allow super small memory settings for development

### LINKED ISSUE #229: Deploy dev/test cluster?
Hello, I'm new to Redpanda and I'm trying to figure out how to deploy a test cluster to see if my application works with it as a drop-in-replacement.

I'm trying to deploy Redpanda on extremely resource-constrained VMs. I was initially surprised that it simply would not deploy (because the default limits were too high).

Investigating further, I found that the `values.yaml` file has the following comment:
```
  # The default values below are what should work for a development environment.
  # Production-level values and other considerations are provided in comments
  # if those values are different from the default.
```
and:
```
    container:
      # Minimum memory count for each Redpanda broker
      # If omitted, the min value will equal the max value (requested resources defaults to limits)
      # Equivalent to: resources.requests.memory
      # For production: 10Gi or greater
      # min: 2.5Gi

```

This makes sense. 

However, as of now, there's no way to change these values to dev/test values like a 0.5Gi test cluster, as the helm validation fails it with the following error:

```
Error: INSTALLATION FAILED: execution error at (redpanda/templates/_helpers.tpl:341:84):
81 is below the minimum recommended value for Redpanda
```

It definitely makes sense to have strong defaults and enforce best practices, but as far as my understanding goes, it is very much possible to run a Redpanda node with just 0.5 Gi of memory allocated. It doesn't make sense to take away power from the user and make it completely impossible to do so from the Helm deployment.

I also found some discussion about this at https://github.com/redpanda-data/redpanda/issues/175 but that's just for setting the `developer_mode` config parameter, which would probably be required in this case as well, but setting this to true does not help me bypass the check, given that a comment says:
```
# This section allows modifying various Redpanda settings not covered in other sections above.
# These values do not pertain to the kubernetes objects created with helm.
# Instead these parameters get passed directly to the Redpanda binary at startup.
# See https://docs.redpanda.com/docs/cluster-administration/configuration/
```

joejulian:
This allows you to set memory as small as 256Mb. If you do, however, it will display a warning in the NOTES.txt output.

I also added a framework for adding warnings to that output.

Fixes #229

### DIFF
--- a/charts/redpanda/templates/_helpers.tpl
+++ b/charts/redpanda/templates/_helpers.tpl
@@ -283,8 +283,8 @@ Generate configuration needed for rpk
   {{- if eq $result 0 -}}
     {{- "unable to get memory value" | fail -}}
   {{- end -}}
-  {{- if lt $result 2000 -}}
-    {{- printf "\n%d is below the minimum recommended value for Redpanda" $result | fail -}}
+  {{- if lt $result 256 -}}
+    {{- printf "\n%d is below the minimum value for Redpanda" $result | fail -}}
   {{- end -}}
   {{- if gt (add $result (include "redpanda-reserve-memory" .)) (include "container-memory" . | int64) -}}
     {{- printf "\nNot enough container memory for Redpanda memory values\nredpanda: %d, reserve: %d, container: %d" $result (include "redpanda-reserve-memory" . | int64) (include "container-memory" . | int64) | fail -}}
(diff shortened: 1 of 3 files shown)


emaxerrno:
Joe what about copying the container mode from rpk settings (or some of them)

charts/redpanda/templates/_helpers.tpl
```diff
@@ -366,22 +366,22 @@ Generate configuration needed for rpk
...
-{{- toYaml $tunable | nindent 4 }}
-{{- else if (include "redpanda-atleast-22-1-1" . | fromJson).bool }}
-{{- $tunable = unset $tunable "log_segment_size_min" }}
-{{- $tunable = unset $tunable "log_segment_size_max" }}
-{{- $tunable = unset $tunable "kafka_batch_max_bytes" }}
-{{- $tunable = unset $tunable "topic_partitions_per_shard" }}
-{{- toYaml $tunable | nindent 4 }}
-{{- end }}
-{{- end }}
+{{- $tunable := dig "tunable" dict .Values.config -}}
```
RafalKorepta, line 369: Why those `-` are necessary at the end?

RafalKorepta (approved):
I wish we could add tests for such small Redpanda instances.

charts/redpanda/templates/_helpers.tpl
joejulian: In _helpers.tpl, these are used to prevent the return from having any excess newlines. It's mostly just a good standard to follow.

joejulian:
> what about copying the container mode from rpk settings

@emaxerrno Could you please elaborate? I'm not sure what you're asking here.

Issue #227, opened 2022-12-06
Issues with user secrets, and the `secretName` defined in `Values` not being used


carlreid:
I'd like to generate a random secure password and store it in a secret, for it to be used by Redpanda. I had read the section:
```yaml
        # If password isn't given, then the secretName must point to an already existing secret
        # secretName: adminPassword
```

However it seems this isn't used/referenced at all? If I try use it, the `post-install` [script user creation](https://github.com/redpanda-data/helm-charts/blob/60659632be09be685e7adcaf405ffbac85ddbf31/charts/redpanda/templates/post-install-upgrade-job.yaml#L70-L72) outputs:
```bash
Error: accepts at most 1 arg(s), received 3
Usage:
  rpk acl user create [USER] -p [PASS] [flags]
```

To expand, it would also be nice if these users had associated secrets created by the chart. For example, if I create a user by using the `--set` argument, then a secret named `{{ include "redpanda.fullname" . }}-{{ .name }}-user-secret` could be nice, so I can reference it from another deployment in the cluster.

Likewise, it may also be nice to allow a consumer just specify the name of the user without a password, such as:
```yaml
    users:
      - name: service_one
      - name: external_service
```

Then have the chart generate a random secure password and create an associated secret.

This may relate to some changes in https://github.com/redpanda-data/helm-charts/pull/168, or the issue https://github.com/redpanda-data/helm-charts/issues/73 "_possibly pointing to a secret for the user password._"

PR #295, opened 2023-01-24
fix: enable sasl users from secrets.

alejandroEsc:
Note: This is a breaking change.

Tested this with the following values file:

```
auth:
  sasl:
    enabled: true
    secretRef: "redpanda-users"
    users:
    - name: admin
       password: change-me
       mechanism: SCRAM-SHA-512
    - name: user1
       password: change-me-2
       mechanism: SCRAM-SHA-512
    - name: user2
       password: change-me-3
       mechanism: SCRAM-SHA-512
```

and created the appropriate secret. The users were created, you can and should probably use the wait flag:

```
helm install ... --wait
```
The flag waits for the brokers to be up and running, otherwise you will have post-install jobs running. The response 

```
Creating user "admin"...
Created user "admin".
Creating user "user1"...
Created user "user1".
Creating user "user2"...
Created user "user2".
```

and the pods volumes that were added:

```
    Type:        Secret (a volume populated by a Secret)
    SecretName:  panda-users
    Optional:    false
 ```

I also included some nits and a few improvements for ease of error solving.

Fixes: #227 

Part of #203

### DIFF
--- a/charts/redpanda/ci/04-one-node-cluster-no-tls-sasl-values.yaml
+++ b/charts/redpanda/ci/04-one-node-cluster-no-tls-sasl-values.yaml
@@ -19,9 +19,8 @@ tls:
 auth:
   sasl:
     enabled: true
-    users:
-      - name: admin
-        password: hunter2
+    secretRef: "redpanda-users"
+
 storage:
   persistentVolume:
     size: 3Gi
--- a/charts/redpanda/ci/04-one-node-cluster-no-tls-sasl-values.yaml
+++ b/charts/redpanda/ci/04-one-node-cluster-no-tls-sasl-values.yaml
@@ -20,6 +20,9 @@ auth:
   sasl:
     enabled: true
     secretRef: "redpanda-users"
+    users:
+      - name: admin
+        password: hunter2
 
 storage:
   persistentVolume:
--- a/charts/redpanda/templates/_helpers.tpl
+++ b/charts/redpanda/templates/_helpers.tpl
@@ -346,6 +346,16 @@ Generate configuration needed for rpk
 {{ join " " (list $flags.brokers $flags.admin $flags.sasl $flags.kafka)}}
 {{- end -}}
 
+{{- define "rpk-common-flags-no-admin" -}}
+{{- $flags := fromJson (include "rpk-flags" .) -}}
+{{ join " " (list $flags.brokers $flags.kafka $flags.sasl)}}
+{{- end -}}
+
+{{- define "rpk-common-flags-no-admin-no-sasl" -}}
+{{- $flags := fromJson (include "rpk-flags" .) -}}
+{{ join " " (list $flags.brokers $flags.kafka)}}
+{{- end -}}
+
 {{- define "rpk-common-flags-no-sasl" -}}
 {{- $flags := fromJson (include "rpk-flags" .) -}}
 {{ join " " (list $flags.brokers $flags.admin $flags.kafka)}}
--- a/charts/redpanda/templates/_helpers.tpl
+++ b/charts/redpanda/templates/_helpers.tpl
@@ -346,17 +346,12 @@ Generate configuration needed for rpk
 {{ join " " (list $flags.brokers $flags.admin $flags.sasl $flags.kafka)}}
 {{- end -}}
 
-{{- define "rpk-common-flags-no-admin" -}}
+{{- define "rpk-flags-no-admin" -}}
 {{- $flags := fromJson (include "rpk-flags" .) -}}
 {{ join " " (list $flags.brokers $flags.kafka $flags.sasl)}}
 {{- end -}}
 
-{{- define "rpk-common-flags-no-admin-no-sasl" -}}
-{{- $flags := fromJson (include "rpk-flags" .) -}}
-{{ join " " (list $flags.brokers $flags.kafka)}}
-{{- end -}}
-
-{{- define "rpk-common-flags-no-sasl" -}}
+{{- define "rpk-flags-no-sasl" -}}
 {{- $flags := fromJson (include "rpk-flags" .) -}}
 {{ join " " (list $flags.brokers $flags.admin $flags.kafka)}}
 {{- end -}}
--- a/charts/redpanda/templates/_helpers.tpl
+++ b/charts/redpanda/templates/_helpers.tpl
@@ -19,19 +19,19 @@ Expand the name of the chart.
 */}}
 {{- define "redpanda.name" -}}
 {{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
-{{- end }}
+{{- end -}}
 
 {{/*
 Create a default fully qualified app name.
 We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
 */}}
 {{- define "redpanda.fullname" -}}
-{{- if .Values.fullnameOverride }}
-{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" }}
-{{- else }}
-{{- printf "%s" .Release.Name | trunc 63 | trimSuffix "-" }}
-{{- end }}
-{{- end }}
+{{- if .Values.fullnameOverride -}}
+{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" -}}
+{{- else -}}
+{{- printf "%s" .Release.Name | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+{{- end -}}
 
 {{/*
 Create chart name and version as used by the chart label.
@@ -300,10 +300,6 @@ Generate configuration needed for rpk
 {{- dig "sasl" "mechanism" "SCRAM-SHA-512" .Values.auth -}}
 {{- end -}}
 
-{{- define "sasl-user-mechanism" -}}
-{{- dig "mechanism" (include "sasl-mechanism" $) $.user -}}
-{{- end -}}
-
 {{- define "rpk-flags" -}}
   {{- $root := . -}}
   {{- $admin := list -}}
@@ -325,11 +321,10 @@ Generate configuration needed for rpk
   {{- end -}}
   {{- $sasl := list -}}
   {{- if (include "sasl-enabled" . | fromJson).bool -}}
-    {{- $root := . | toJson | fromJson -}}
     {{- $sasl = concat $sasl (list
-      "--user" (dig "auth" "username" (first .Values.auth.sasl.users).name $root)
-      "--password" (dig "auth" "password" (first .Values.auth.sasl.users).password $root)
-      "--sasl-mechanism " (include "sasl-mechanism" .)
+      "--user" ( print "$(find /etc/secrets/users/* -print | sed -n 1p | xargs cat | sed -n 1p | tr ':' '\n' | sed -n 1p )" | quote )
+      "--password" ( print "$(find /etc/secrets/users/* -print | sed -n 1p | xargs cat | sed -n 1p | tr ':' '\n' | sed -n 2p )" | quote )
+      "--sasl-mechanism" ( printf "$(find /etc/secrets/users/* -print | sed -n 1p | xargs cat | sed -n 1p | tr ':' '\n' | sed -n 3p | grep . || echo %s )" (include "sasl-mechanism" .) | quote )
     )
     -}}
   {{- end -}}
@@ -351,9 +346,26 @@ Generate configuration needed for rpk
 {{ join " " (list $flags.brokers $flags.admin $flags.sasl $flags.kafka)}}
 {{- end -}}
 
+{{- define "rpk-common-flags-no-sasl" -}}
+{{- $flags := fromJson (include "rpk-flags" .) -}}
+{{ join " " (list $flags.brokers $flags.admin $flags.kafka)}}
+{{- end -}}
+
+{{- define "rpk-dummy-sasl" -}}
+{{- if (include "sasl-enabled" . | fromJson).bool -}}
+{{ "--user <admin-user-in-secret> --password <admin-password-in-secret> --sasl-mechanism <mechanism-in-secret>" -}}
+{{- else -}}
+{{ "" }}
+{{- end -}}
+{{- end -}}
+
 {{- define "rpk-topic-flags" -}}
 {{- $flags := fromJson (include "rpk-flags" .) -}}
-{{ join " " (list $flags.brokers $flags.sasl $flags.kafka)}}
+    {{- if (include "sasl-enabled" . | fromJson).bool -}}
+        {{- join " " (list $flags.brokers $flags.kafka $flags.sasl) -}}
+    {{- else -}}
+        {{- join " " (list $flags.brokers $flags.kafka) -}}
+    {{- end -}}
 {{- end -}}
 
 {{- define "storage-min-free-bytes" -}}
--- a/charts/redpanda/templates/_helpers.tpl
+++ b/charts/redpanda/templates/_helpers.tpl
@@ -356,6 +356,11 @@ Generate configuration needed for rpk
 {{ join " " (list $flags.brokers $flags.admin $flags.kafka)}}
 {{- end -}}
 
+{{- define "rpk-flags-no-admin-no-sasl" -}}
+{{- $flags := fromJson (include "rpk-flags" .) -}}
+{{ join " " (list $flags.brokers $flags.kafka)}}
+{{- end -}}
+
 {{- define "rpk-dummy-sasl" -}}
 {{- if (include "sasl-enabled" . | fromJson).bool -}}
 {{ "--user <admin-user-in-secret> --password <admin-password-in-secret> --sasl-mechanism <mechanism-in-secret>" -}}
--- a/charts/redpanda/templates/post-install-upgrade-job.yaml
+++ b/charts/redpanda/templates/post-install-upgrade-job.yaml
@@ -109,9 +109,9 @@ spec:
 
         {{- if (include "redpanda-atleast-22-2-0" . | fromJson).bool }}
           {{- if not (empty .Values.license_secret_ref) }}
-            rpk cluster license set "$REDPANDA_LICENSE" {{ template "rpk-common-flags" $ }}
+            rpk cluster license set "$REDPANDA_LICENSE" {{ template "rpk-common-flags-no-sasl" $ }}
           {{- else if not (empty .Values.license_key) }}
-            rpk cluster license set {{ .Values.license_key | quote }} {{ template "rpk-common-flags" $ }}
+            rpk cluster license set {{ .Values.license_key | quote }} {{ template "rpk-common-flags-no-sasl" $ }}
           {{- end }}
         {{- end }}
         {{- with .Values.post_install_job.resources }}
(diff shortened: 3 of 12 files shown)

### CODE REVIEW

**Review by RafalKorepta — COMMENTED** [review]
If you defining the `auth.sasl.users` object, could you change the values-schema.json definition to help end user set the right properites?

File: charts/redpanda/templates/_helpers.tpl [hunk 1]
```diff
@@ -304,6 +304,21 @@ Generate configuration needed for rpk
 {{- dig "mechanism" (include "sasl-mechanism" $) $.user -}}
 {{- end -}}
 
+{{- define "sasl-first-users" -}}
+{{- toJson ( dict "list" .Values.auth.sasl.users ) -}}
```
RafalKorepta, line 308: I don't understand how this can be `first` sasl user template. If I read this correctly, whole users list will be marshalled to json, right?

charts/redpanda/templates/_helpers.tpl
```diff
@@ -304,6 +304,21 @@ Generate configuration needed for rpk
...
+{{- end -}}
+
+{{- define "sasl-user-password" -}}
+{{- $pass := "" -}}
+{{- $user := (first (include "sasl-first-users" .| fromJson).list) -}}
```
RafalKorepta, line 313: Why we have conversion from list to json to list? Am I missing something obvious in helm?

RafalKorepta: Why not
```suggestion
{{- $user := (first .Values.auth.sasl.users -}}
```

alejandroEsc:
> If you defining the `auth.sasl.users` object, could you change the values-schema.json definition to help end user set the right properites?

I did not change the schema, nor changed the object here.

charts/redpanda/templates/post-install-upgrade-job.yaml
```diff
@@ -76,23 +76,27 @@ spec:
 {{- if .Values.auth.sasl.enabled }}
   {{- $values := .Values }}
   {{- range $user := .Values.auth.sasl.users }}
-            # To avoid `set -e` from exiting the command when a user exists; catch the stderr output and exit codes into `creation_result`
-            # and `creation_result_exit_code` for use later
-            creation_result=$(rpk acl user create {{ $user.name }} -p {{ $user.password | quote }} --mechanism {{ include "sasl-user-mechanism" (dict "user" $user "Values" $values) }} {{ template "rpk-common-flags" $ }} 2>&1) && creation_result_exit_code=$? || creation_result_exit_code=$?
-
-            # On a non-success exit code
+    {{- if $user.password }}
+            PASSWORD={{ $user.password | quote }}
```
RafalKorepta, line 80: Should we even allow that? In my understanding any Password should be protected and not given as a plain text of the job resource.

charts/redpanda/templates/_helpers.tpl
alejandroEsc: The first would presumably be the admin user, this is the previous way this was "working" which i would argue never really did. Regardless I need one username/pass combination to pass into flags, a choice has to be made.

charts/redpanda/templates/post-install-upgrade-job.yaml
```diff
@@ -76,23 +76,27 @@ spec:
...
+    {{- else if not (empty $user.secretName) }}
+            PASSWORD=$(cat {{ printf "/etc/secret-name/%s/password" $user.secretName }})
+    {{- else }}
+            PASSWORD=""
```
RafalKorepta, line 84: If user does not provide `secretName` nor `password` could we add Notes for them to set the password. I don't even now if Redpanda Admin API allows service account creation without any password.

charts/redpanda/templates/_helpers.tpl
alejandroEsc: I tried this, i kept getting a {}.auth nil interface error.

charts/redpanda/templates/post-install-upgrade-job.yaml
alejandroEsc: There is already a validation check. I left the default password there uncommented because we should use defaults. The validation error is 
```
Error: INSTALLATION FAILED: values don't meet the specifications of the schema(s) in the following chart(s):
redpanda:
- auth.sasl: Must validate "then" as "if" was valid
- auth.sasl.users.0: Must validate one and only one schema (oneOf)
- auth.sasl.users.0: password is required

```

charts/redpanda/templates/post-install-upgrade-job.yaml
alejandroEsc: That was the previous behavior? I can change this to just push any password into a secret, but that would be a larger PR and probably one that addresses this directly.

charts/redpanda/templates/post-install-upgrade-job.yaml
```diff
@@ -76,23 +76,27 @@ spec:
...
+    {{- end }}
+            echo {{ printf "Creating user %q..." $user.name | quote }}
+            creation_result=$(rpk acl user create {{ $user.name }} --password=${PASSWORD} --mechanism {{ include "sasl-user-mechanism" (dict "user" $user "Values" $values) }} {{ template "rpk-common-flags-no-sasl" $ }} 2>&1) && creation_result_exit_code=$? || creation_result_exit_code=$?            # On a non-success exit code
             if [[ $creation_result_exit_code -ne 0 ]]; then
               # Check if the stderr contains "User already exists"
               if [[ $creation_result == *"User already exists"* ]]; then
-                printf "The %s user already exists, skipping creation.\n" {{ $user.name }}
+                echo {{ printf "The %s user already exists, skipping creation." $user.name | quote }}
```
RafalKorepta, line 91: NIT: It would be good to add `rpk acl user update` command, but we can use curl to PATCH the password.

charts/redpanda/values.yaml
```diff
@@ -58,27 +58,25 @@ rackAwareness:
...
-#
 # Authentication
 auth:
-  #
   # SASL configuration
   sasl:
-    enabled: false
+    enabled: true
+    # if unspecified, mechanism defaults to SCRAM-SHA-512
+    # mechanism: SCRAM-SHA-512
```
RafalKorepta, line 67: What this line of comment mean?

charts/redpanda/values.yaml
```diff
@@ -58,27 +58,25 @@ rackAwareness:
...
     # user list
-    # TODO create user at startup
+    # The chart assumes that the first user is "admin"
```
RafalKorepta, line 69: What do you mean by "admin"?

Redpanda has only super-users notion. Any other service account needs to have ACL set up.

charts/redpanda/values.yaml
RafalKorepta: Sorry, I see this is commented value.

charts/redpanda/values.yaml
alejandroEsc: thank you for clarifying that, It wasnt clear to me.

RafalKorepta (commented):
I might not understand how "admin user" can be used in template `rpk-flags`. This "admin" is not used anywhere.

charts/redpanda/templates/post-install-upgrade-job.yaml
RafalKorepta: Thanks, comment resolved 👍

alejandroEsc:
> I might not understand how "admin user" can be used in template `rpk-flags`. This "admin" is not used anywhere.

This is used in rpk-common-flags, and it is used in a few places. The admin is a default user when sasl is enabled.

charts/redpanda/templates/post-install-upgrade-job.yaml
alejandroEsc: Quick question, do you know if `rpk acl user update` is idempotent? Can it also "create" a user even it it doesn't exist?

charts/redpanda/values.yaml
alejandroEsc: Updated the text to account for this.

charts/redpanda/templates/_helpers.tpl
RafalKorepta: Then I would rename it to `sasl-users`

charts/redpanda/templates/post-install-upgrade-job.yaml
RafalKorepta: What was the values provided? `--set auth.sasl.enable=true` or more?

alejandroEsc:
```
--set auth.sasl.enable=true
```

Yes, i uncommented the password for that particular case so as to avoid this issue, but now reverted it as i think having a user manually set their passwords would be more appropriate. Personally I would prefer to avoid "creating" passwords for users.

RafalKorepta:
> This is used in rpk-common-flags, and it is used in a few places. The admin is a default user when sasl is enabled.

I agree that this first service account is used in the template definition. When you take a closer look:
`--user` and `--password` is used only in `rpk-flags` https://github.com/redpanda-data/helm-charts/blob/6bada0dd4546c706f2ea8f868c67d5bb5236e041/charts/redpanda/templates/_helpers.tpl#L344

The `rpk-flags` is called in `rpk-common-flags` and `rpk-topic-flags`
https://github.com/redpanda-data/helm-charts/blob/6bada0dd4546c706f2ea8f868c67d5bb5236e041/charts/redpanda/templates/_helpers.tpl#L363-L365
https://github.com/redpanda-data/helm-charts/blob/6bada0dd4546c706f2ea8f868c67d5bb5236e041/charts/redpanda/templates/_helpers.tpl#L373-L375

## rpk-common-flags

The common flags is used wrongly in post-upgrade job. 
https://github.com/redpanda-data/helm-charts/blob/6bada0dd4546c706f2ea8f868c67d5bb5236e041/charts/redpanda/templates/post-upgrade.yaml#L3
Nor `rpk config set` nor `rpk config import` is using service account to authenticate. Those commands are using Admin API endpoint and it can be protected by mTLS only.
https://github.com/redpanda-data/helm-charts/blob/6bada0dd4546c706f2ea8f868c67d5bb5236e041/charts/redpanda/templates/post-upgrade.yaml#L49
https://github.com/redpanda-data/helm-charts/blob/6bada0dd4546c706f2ea8f868c67d5bb5236e041/charts/redpanda/templates/post-upgrade.yaml#L51
https://github.com/redpanda-data/helm-charts/blob/6bada0dd4546c706f2ea8f868c67d5bb5236e041/charts/redpanda/templates/post-upgrade.yaml#L54

And again in post install upgrade job as `rpk cluster license set` is using not using sasl credentials to set the license. Under the hood the Admin API port/endpoint is called.
https://github.com/redpanda-data/helm-charts/blob/6bada0dd4546c706f2ea8f868c67d5bb5236e041/charts/redpanda/templates/post-install-upgrade-job.yaml#L105
https://github.com/redpanda-data/helm-charts/blob/6bada0dd4546c706f2ea8f868c67d5bb5236e041/charts/redpanda/templates/post-install-upgrade-job.yaml#L107

Even in example commands the `rpk-common-flags` is wrongly used as `rpk acl user` subcommand is using Admin API (no sasl parameters required). 
https://github.com/redpanda-data/helm-charts/blob/6bada0dd4546c706f2ea8f868c67d5bb5236e041/charts/redpanda/templates/_example-commands.tpl#L24-L27

## rpk-topic-flags

With this definition I have mixed feelings as sasl is included. 

This example of `rpk-topic-flags` usage should fail test our pipeline as first user and password does not have set ACLs
https://github.com/redpanda-data/helm-charts/blob/6bada0dd4546c706f2ea8f868c67d5bb5236e041/charts/redpanda/templates/tests/test-kafka-produce-consume.yaml#L60-L62

This should fail in our CI as ACLs are not set for the first user
https://github.com/redpanda-data/helm-charts/blob/6bada0dd4546c706f2ea8f868c67d5bb5236e041/charts/redpanda/templates/_example-commands.tpl#L29-L31
https://github.com/redpanda-data/helm-charts/blob/6bada0dd4546c706f2ea8f868c67d5bb5236e041/charts/redpanda/templates/_example-commands.tpl#L34-L36

Here we are replacing the `--user` flag, so we should change the included template with `rpk-common-flags-no-sasl` template
https://github.com/redpanda-data/helm-charts/blob/6bada0dd4546c706f2ea8f868c67d5bb5236e041/charts/redpanda/templates/tests/test-kafka-sasl-status.yaml#L18

Here we are doing hot swap of username and password for tests. I would like to not change sasl credentials for tests and use the first user defined in users list.
https://github.com/redpanda-data/helm-charts/blob/6bada0dd4546c706f2ea8f868c67d5bb5236e041/charts/redpanda/templates/_example-commands.tpl#L39-L43
https://github.com/redpanda-data/helm-charts/blob/6bada0dd4546c706f2ea8f868c67d5bb5236e041/charts/redpanda/templates/_example-commands.tpl#L46-L51
https://github.com/redpanda-data/helm-charts/blob/6bada0dd4546c706f2ea8f868c67d5bb5236e041/charts/redpanda/templates/_example-commands.tpl#L53-L57

P.S. I edited PR cover letter and added Fixes: #227. If you think this not solves all issues, then you can change word `Fixes` with `Refers`.

charts/redpanda/templates/post-install-upgrade-job.yaml
d3k4z: @RafalKorepta is right, we need to protect users credentials. As proposed by @alejandroEsc the right way to do it is to push it into etcd(vault, etc) as a secret and use it from there.

charts/redpanda/templates/_helpers.tpl
joejulian: `{"list": [...]}` is the return string. You're not converting to list (that would be `| list`), you're selecting the "list" key.

joejulian: The key should probably have a better name.

charts/redpanda/templates/post-install-upgrade-job.yaml
joejulian: That would be a breaking change. We can issue a warning, but not break existing users.

alejandroEsc:
> rpk-acl-user-create

I believe I addressed these, please take a look.

charts/redpanda/ci/04-one-node-cluster-no-tls-sasl-values.yaml
```diff
@@ -19,9 +19,8 @@ tls:
 auth:
   sasl:
     enabled: true
-    users:
-      - name: admin
-        password: hunter2
+    secretRef: "redpanda-users"
+
```
joejulian, line 23: Let's leave this one as name/password. Test 05 will test with secretRef and this way we have both cases covered.

alejandroEsc: that is my fault, it was me testing defaults and i forgot to revert this, good catch!

charts/redpanda/templates/_helpers.tpl
```diff
@@ -351,9 +346,36 @@ Generate configuration needed for rpk
 {{ join " " (list $flags.brokers $flags.admin $flags.sasl $flags.kafka)}}
 {{- end -}}
 
+{{- define "rpk-common-flags-no-admin" -}}
```
joejulian, line 349: These aren't really common to anything else, so let's just drop the common part.

```suggestion
{{- define "rpk-flags-no-admin" -}}
```

charts/redpanda/templates/_helpers.tpl
```diff
@@ -351,9 +346,36 @@ Generate configuration needed for rpk
...
+{{- $flags := fromJson (include "rpk-flags" .) -}}
+{{ join " " (list $flags.brokers $flags.kafka $flags.sasl)}}
+{{- end -}}
+
+{{- define "rpk-common-flags-no-admin-no-sasl" -}}
```
joejulian, line 354:
```suggestion
{{- define "rpk-flags-no-admin-no-sasl" -}}
```

charts/redpanda/templates/_helpers.tpl
```diff
@@ -351,9 +346,36 @@ Generate configuration needed for rpk
...
+{{- $flags := fromJson (include "rpk-flags" .) -}}
+{{ join " " (list $flags.brokers $flags.kafka)}}
+{{- end -}}
+
+{{- define "rpk-common-flags-no-sasl" -}}
```
joejulian, line 359:
```suggestion
{{- define "rpk-flags-no-sasl" -}}
```

charts/redpanda/templates/secrets.yaml
```diff
@@ -75,3 +76,17 @@ stringData:
         sleep 0.5
     done
 {{- end }}
+{{- if and .Values.auth.sasl.enabled ( gt ( len .Values.auth.sasl.users ) 0 ) }}
```
joejulian, line 79: Should we have a value that allows the user to administer this secret outside of helm?

alejandroEsc: Not sure what that would get us, wouldnt just an empty list and just a secret referencing things be enough? It does require a bit more logic but i think its tolerable.

charts/redpanda/values.yaml
```diff
@@ -104,7 +107,7 @@ tls:
 external:
   # Default external access value for each service
   # Services can toggle external access per listener (ie. `listeners.<service name>.external.<listener name>.enabled`)
-  enabled: true
+  enabled: false
```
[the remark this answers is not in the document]
alejandroEsc: Nope, this was carried over and caused an issue to my testing. Should this be enabled by default?

charts/redpanda/templates/secrets.yaml
```diff
@@ -20,6 +20,7 @@ apiVersion: v1
 kind: Secret
```
joejulian, line 20: If the user is going to manage this secret manually, they need a way to prevent the helm chart from failing when it tries to create the same secret.
```suggestion
{{- if .Values.auth.sasl.createSecret }}
kind: Secret
```

charts/redpanda/templates/secrets.yaml
```diff
@@ -91,3 +92,21 @@ stringData:
         sleep 0.5
     done
 {{- end }}
+{{- if and .Values.auth.sasl.enabled ( gt ( len .Values.auth.sasl.users ) 0 ) }}
+---
+apiVersion: v1
```
joejulian, line 97: If the user is going to manage this secret manually, they need a way to prevent the helm chart from failing when it tries to create the same secret.
```suggestion
{{- if .Values.auth.sasl.createSecret }}
apiVersion: v1
```

charts/redpanda/templates/secrets.yaml
joejulian: An empty list is false

```suggestion
{{- if and .Values.auth.sasl.enabled .Values.auth.sasl.users }}
```

charts/redpanda/templates/secrets.yaml
```diff
@@ -91,3 +92,21 @@ stringData:
...
+kind: Secret
+metadata:
+  name: {{ .Values.auth.sasl.secretRef | quote }}
```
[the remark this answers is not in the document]
alejandroEsc: Thank you! this is what PRs are for, i completely missed this :face_with_spiral_eyes:

charts/redpanda/templates/secrets.yaml
alejandroEsc: alright

charts/redpanda/values.yaml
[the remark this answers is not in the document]
joejulian: Heh, that's a whole other discussion topic. :grin:

joejulian (approved):

PR #297, opened 2023-01-25
redpanda: default to putting redpanda on separate nodes

joejulian:
Set the default anti-affinity rule to prevent broker pods from being created on the same Node.


alejandroEsc (approved):
thank you!

PR #300, opened 2023-01-26
Remove note about unsetting podAntiAffinity

JakeSCahill:
The chart doesn't allow this action, so we shouldn't document it.

```
redpanda:
- statefulset.podAntiAffinity: type is required

redpanda:
- statefulset.podAntiAffinity.type: Does not match pattern '^(hard|soft)$'
```


charts/redpanda/values.yaml
```diff
@@ -477,8 +477,7 @@ statefulset:
     # The topologyKey to be used.
     # Can be used to spread across different nodes, AZs, regions etc.
     topologyKey: kubernetes.io/hostname
-    # Type of anti-affinity rules: either `soft`, `hard` or empty value (which
-    # disables anti-affinity rules).
+    # Type of anti-affinity rules: either `soft`, `hard`.
```
JakeSCahill, line 480:
```suggestion
    # Type of anti-affinity rules: either `soft` or `hard`.
```

RafalKorepta (approved):

charts/console/Chart.yaml
```diff
@@ -27,7 +27,7 @@ type: application
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
 
 # Chart versions do not track appVersion
-version: 0.4.1
+version: 0.4.2
```
RafalKorepta, line 30: Why we are bumping console chart version?

PR #306, opened 2023-01-31
fix: allow jobs and sts to have the same tolerations and nodeselectors

### LINKED ISSUE #305: Jobs don't set tolerations or nodeSelector (if set)
If a cluster has nodes that have taints and labels applied which limit what can be scheduled on them, then jobs in the helm chart won't run. The easiest solution may be to add the same nodeSelector and tolerations that are on the statefulset if they exist.

I think this is a P0 but want to discuss with the team. I just found this issue, and it is blocking a customer deliverable.

alejandroEsc:
Tested this locally:

- Creates tolerations and node-selectors for all pods/jobs from a global tolerations and nodeSelectors field.
- Allow for sts to override these values if so needed

Fixes #305

### DIFF
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 2.6.5
+version: 2.6.6
 
 # The app version is the default version of Redpanda to install.
 appVersion: v22.3.10
--- a/charts/redpanda/templates/_statefulset-helpers.tpl
+++ b/charts/redpanda/templates/_statefulset-helpers.tpl
@@ -1,0 +1,38 @@
+{{/*
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
+(the "License"); you may not use this file except in compliance with
+the License.  You may obtain a copy of the License at
+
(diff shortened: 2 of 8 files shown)


vuldin (approved):
Looks great, thanks for getting this PR up so quickly! Also thanks for creating `_statefulset-helpers.tpl`, I think this approach to organizing named templates will help over time.

I've tested this change out and it works as intended. My only feedback is that the new named templates are an either/or approach when maybe they could allow an aggregate for both nodeSelector and tolerations. In other words. If values are found in both statefulset and the top-level parameter then both are used. This isn't a blocker though, we can talk about it and improve the implementation over time if we agree.

joejulian:
> My only feedback is that the new named templates are an either/or approach when maybe they could allow an aggregate for both nodeSelector and tolerations. In other words. If values are found in both statefulset and the top-level parameter then both are used. This isn't a blocker though, we can talk about it and improve the implementation over time if we agree.

This would break if the user wanted only the sts to be on one set of nodes, and all the ancillary work to run on another, ie. Jobs run on `mylabel: jobs` and redpanda runs on `mylabel: redpanda`. Merging them would create an un-matchable combination.

joejulian:
I'm working on a PR to ensure that the jobs and statefulset pods use distinct labels so as not to have jobs match the statefulset selector. That PR needs merged first.

RafalKorepta:
@alejandroEsc Can you add test that uses `nodeSelector` and `tolerations`? You could annotate kind workers to achieve that.

alejandroEsc:
> tolerations

I know i can use the labels on kind workers. What do you mean by "use annotations"?

RafalKorepta:
I wrongly write `annotate`. I should write `label them`

PR #310, opened 2023-02-02
use pod labels that don't interfere with each other

joejulian:
The same labels were being used for the statefulset pods and the job pods. This was causing them to interfere with the pod anti-affinity and the podDisruptionBudget. This changes the labels so they can be correctly matched.

### DIFF
--- a/charts/redpanda/templates/_statefulset.tpl
+++ b/charts/redpanda/templates/_statefulset.tpl
@@ -1,0 +1,36 @@
+{{/*
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
+(the "License"); you may not use this file except in compliance with
+the License.  You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/}}
+
+{{- define "statefulset-pod-labels" -}}
+app.kubernetes.io/name: {{ template "redpanda.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name | quote }}
+app.kubernetes.io/component: {{ (include "redpanda.name" .) | trunc 51 }}-statefulset
+{{- with .Values.commonLabels }}
+{{ toYaml . }}
+{{- end }}
+{{- end -}}
+
+{{/*
+Set default path for tiered storage cache or use one provided
+*/}}
+{{- define "tieredStorage.cacheDirectory" -}}
+{{- if empty .Values.storage.tieredConfig.cloud_storage_cache_directory -}}
+  {{- printf "/var/lib/redpanda/data/cloud_storage_cache" -}}
+{{- else -}}
+  {{- .Values.storage.tieredConfig.cloud_storage_cache_directory -}}
+{{- end -}}
+{{- end -}}
--- a/charts/redpanda/templates/_statefulset.tpl
+++ b/charts/redpanda/templates/_statefulset.tpl
@@ -16,12 +16,22 @@ limitations under the License.
 */}}
 
 {{- define "statefulset-pod-labels" -}}
+{{- /*
+  StatefulSets cannot change their selector. Use the existing one even if it's broken.
+  New installs will get better selectors.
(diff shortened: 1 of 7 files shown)


alejandroEsc:
I believe this is the error here:

```
2023-02-02T19:38:07.9052231Z Error: UPGRADE FAILED: cannot patch "redpanda-fjyang5y9v" with kind StatefulSet: StatefulSet.apps "redpanda-fjyang5y9v" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', 'updateStrategy', 'persistentVolumeClaimRetentionPolicy' and 'minReadySeconds' are forbidden
```

charts/redpanda/templates/statefulset.yaml
```diff
@@ -65,23 +59,19 @@ spec:
       initContainers:
         - name: set-datadir-ownership
           image: busybox:latest
-          {{- $uid := dig "podSecurityContext" "runAsUser" .Values.statefulset.securityContext.runAsUser .Values.statefulset }}
```
alejandroEsc, line 68: thank you for that, its been annoying me and was going to change this on a nit PR.

alejandroEsc (commented):
So this breaks updates because labels cannot be changed in this manner, which is very unfortunate. Could we keep the sts pod labels as they were change the pod labels of the jobs instead?

joejulian:
> So this breaks updates because labels cannot be changed in this manner, which is very unfortunate. Could we keep the sts pod labels as they were change the pod labels of the jobs instead?

It's not the labels, it's the selector. The selector would match the jobs unless we removed the standard labels. I've found an alternative that should be a good workaround for existing users, but will be correct for new installs. There's also a path to fixing it for existing users that requires a manual step - but we can handle that with documentation.

cc @JakeSCahill

charts/redpanda/templates/statefulset.yaml
```diff
@@ -331,7 +322,7 @@ spec:
 {{- if gt ( .Values.statefulset.replicas | int64 ) 2 }}
         - name: lifecycle-scripts
           secret:
-            secretName: {{ template "redpanda.fullname" . }}-sts-lifecycle
+            secretName: {{ trunc 50 (include "redpanda.fullname" .) }}-sts-lifecycle
```
RafalKorepta, line 325: NIT: Just to be consistent
```suggestion
            secretName: {{ (include "redpanda.fullname" . | trunc 50 ) }}-sts-lifecycle
```

charts/redpanda/templates/statefulset.yaml
```diff
@@ -37,23 +39,16 @@ metadata:
...
     metadata:
-      labels:
-        app.kubernetes.io/name: {{ template "redpanda.name" . }}
-        app.kubernetes.io/instance: {{ .Release.Name | quote }}
-        app.kubernetes.io/component: {{ template "redpanda.name" . }}
-{{- with .Values.commonLabels }}
-        {{- toYaml . | nindent 8 }}
-{{- end }}
+      labels: {{ (include "statefulset-pod-labels" .) | nindent 8 }}
+        redpanda.com/poddisruptionbudget: {{ template "redpanda.name" . }}
```
RafalKorepta, line 51: Why `redpanda.com/poddisruptionbudget` is not included in `_statefulset.tpl` file under `"statefulset-pod-labels"` template?

RafalKorepta (approved):
Over all LGTM

alejandroEsc (approved):
lgtm!

charts/redpanda/templates/_statefulset.tpl
```diff
@@ -0,0 +1,46 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/}}
+
+{{- define "statefulset-pod-labels" -}}
+{{- /*
+  StatefulSets cannot change their selector. Use the existing one even if it's broken.
+  New installs will get better selectors.
+*/ -}}
+{{- $sts := lookup "apps/v1" "StatefulSet" .Release.Namespace (include "redpanda.fullname" .) -}}
```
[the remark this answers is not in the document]
RafalKorepta: I'm starting to question aren't we too clever in templating

charts/redpanda/templates/statefulset.yaml
joejulian: Because we _can_ change labels, but we cannot change the selector. Since we want to use the same template for labels and selector, but existing installs are broken, this at least allows the PDB to only target the STS pods since the PDB selector _is_ mutable.

charts/redpanda/templates/_statefulset.tpl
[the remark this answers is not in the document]
joejulian: I do agree, @RafalKorepta. But I want to avoid completely breaking existing users since we just did that back in September. I think this is the only way to do that.

Issue #313, opened 2023-02-07
postStart.sh scripts fails with custom cluster domain


jrkinley:
When using [Cloud DNS for GKE](https://cloud.google.com/kubernetes-engine/docs/how-to/cloud-dns#vpc_scope_dns) VPC scoped clusters have a custom DNS name instead of the default `cluster.local` domain. This causes the `postStart.sh` script to fail because it assumes the pod addresses end in `svc.cluster.local`. The script is defined in [secrets.yaml](https://github.com/redpanda-data/helm-charts/blob/main/charts/redpanda/templates/secrets.yaml#L45-L65).

For example, changing the default Kubernetes cluster domain in `chart.yaml`:

```yaml
# Default kubernetes cluster domain
clusterDomain: cluster.redpanda
```

Results in this error:

`
FailedPostStartHook	Exec lifecycle hook ([/bin/bash -c ./var/lifecycle/postStart.sh]) for Container "redpanda" in Pod "redpanda-dwh-redpanda-1_dwh-redpanda(e4db9a3c-2e94-4770-8707-22f4a8d5ad26)" failed - error: command '/bin/bash -c ./var/lifecycle/postStart.sh' exited with 137: ++ curl --silent --fail http://redpanda-dwh-redpanda-1.redpanda-dwh-redpanda.dwh-redpanda.svc.cluster.local:9644/v1/node_config ++ grep -o '\"node_id\":[^,}]*' ++ grep -o '[^: ]*$' , message: "++ curl --silent --fail http://redpanda-dwh-redpanda-1.redpanda-dwh-redpanda.dwh-redpanda.svc.cluster.local:9644/v1/node_config\n++ grep -o '\\\"node_id\\\":[^,}]*'\n++ grep -o '[^: ]*$'\n"
`

PR #314, opened 2023-02-07
Fix postStart.sh hook with custom Service domain (kubeadm's network.dnsDomain)

jrkinley:
This PR fixes the postStart.sh script when using a custom Kubernetes cluster domain. Specifically, when using Cloud DNS in GKE with VPC scoped domain names as described in issue #313.

### DIFF
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 2.8.0
+version: 2.8.1
 
 # The app version is the default version of Redpanda to install.
 appVersion: v22.3.10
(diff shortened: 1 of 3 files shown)


joejulian:
This would require external dns for all users, would require network access exceptions, and will route traffic outside the kubernetes cluster. Many of these things would break existing users.

We're working on a revamp of the cert framework to resolve this differently.

For now, if the user can revert the `default` cert and, instead, add a new certificate for their external domain and use that certificate for the external listeners, that should solve this for them.

joejulian:
> This would require external dns for all users, would require network access exceptions, and will route traffic outside the kubernetes cluster. Many of these things would break existing users.
> 
> We're working on a revamp of the cert framework to resolve this differently.
> 
> For now, if the user can revert the `default` cert and, instead, add a new certificate for their external domain and use that certificate for the external listeners, that should solve this for them.

scratch all that, I completely misread the intent.

joejulian (approved):
Thanks!

PR #318, opened 2023-02-07
chore: remove extra dummy helper vars and simplify code

alejandroEsc:
Removes the dummy variables, fixes the logic on tests that kept it from running properly.


joejulian (approved):
happy-dance

PR #319, opened 2023-02-07
add the "M" to the end of the reserve-memory flag

joejulian:
In theory, this should be a no-op. Mostly this is for consistency and disambiguation, to make the command line easier for humans to interpret.


charts/redpanda/templates/statefulset.yaml
```diff
@@ -275,7 +275,7 @@ spec:
             - start
             - --smp={{ .Values.resources.cpu.cores }}
             - --memory={{ template "redpanda-memory" . }}M
-            - --reserve-memory={{ template "redpanda-reserve-memory" . }}
+            - --reserve-memory={{ template "redpanda-reserve-memory" . }}M
```
alejandroEsc, line 278: The values file describe this unit, and should it always be an M?

alejandroEsc (commented):
Only one question.

charts/redpanda/templates/statefulset.yaml
joejulian: Yes, we convert it in the template from SI units to a int64 (in MiB)

alejandroEsc (approved):
lgtm, we can discuss tomorrow.

RafalKorepta (approved):

PR #320, opened 2023-02-07
Add LoadBalancer service

vuldin:
- bump minor chart version
- created new template external-loadbalancer-enabled
- created new template file service.loadbalancer.yaml
- add `external.type: LoadBalancer`
- add `external.sourceRanges` for whitelisting addresses in LoadBalancer

### DIFF
--- a/charts/redpanda/templates/NOTES.txt
+++ b/charts/redpanda/templates/NOTES.txt
@@ -49,6 +49,13 @@ The pods will rollout in a few seconds. To check the status:
 
   kubectl -n {{ .Release.Namespace }} rollout status statefulset {{ template "redpanda.fullname" . }} --watch
 
+{{- if and (toJson (dict "bool" .Values.external.enabled)) (eq .Values.external.type "LoadBalancer") }}
+
+If you are using the load balancer service with a cloud provider, the services will likely have automatically-generated addresses. In this scenario the advertised listeners must be updated in order for external access to work. Run the following command once Redpanda is deployed:
+
+  {{ printf "helm upgrade %s redpanda/redpanda -n %s --set $(kubectl get svc -n %s -o jsonpath='{\"external.addresses={\"}{ range .items[*]}{.status.loadBalancer.ingress[0].ip }{.status.loadBalancer.ingress[0].hostname}{\",\"}{ end }{\"}\\n\"}')" (include "redpanda.name" .) .Release.Namespace .Release.Namespace }}
+{{- end }}
+
 Try some sample commands:
 
 {{- if and $anySASL }}
--- a/charts/redpanda/templates/service.loadbalancer.yaml
+++ b/charts/redpanda/templates/service.loadbalancer.yaml
@@ -1,0 +1,84 @@
+{{/*
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
+(the "License"); you may not use this file except in compliance with
+the License.  You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/}}
+{{- if (include "external-loadbalancer-enabled" . | fromJson).bool }}
+  {{- $values := .Values }}
+  {{- $root := . -}}
+  {{- range $replicaIndex := untilStep 0 ($values.statefulset.replicas|int) 1 }}
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: lb-{{ include "redpanda.fullname" $root }}-{{ $replicaIndex }}
+  namespace: {{ $root.Release.Namespace | quote }}
+  labels:
+{{- with include "full.labels" $root }}
+  {{- . | nindent 4 }}
+{{- end }}
+    repdanda.com/type: "loadbalancer"
+spec:
+  type: LoadBalancer
+{{- if not ( empty $root.Values.external.sourceRanges ) }}
+  loadBalancerSourceRanges:
+  {{- range $values.external.sourceRanges }}
+  - {{ . }}
+  {{- end }}
+{{- end }}
+  externalTrafficPolicy: Local
+  sessionAffinity: None
+  ports:
+{{- range $name, $listener := $values.listeners.admin.external }}
+  {{- $enabled := dig "enabled" $values.external.enabled $listener }}
+  {{- if $enabled }}
+    - name: admin-{{ $name }}
+      protocol: TCP
+      targetPort: {{ $values.listeners.admin.port }}
+      port: {{ dig "nodePort" (first (dig "advertisedPorts" (list $values.listeners.admin.port) $listener)) $listener }}
+  {{- end }}
+{{- end }}
+{{- range $name, $listener := $values.listeners.kafka.external }}
+  {{- $enabled := dig "enabled" $values.external.enabled $listener }}
+  {{- if $enabled }}
+    - name: kafka-{{ $name }}
+      protocol: TCP
+      targetPort: {{ $listener.port }}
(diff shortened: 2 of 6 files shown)


charts/redpanda/values.yaml
```diff
@@ -611,7 +613,7 @@ listeners:
         # enabled: true
         # The port used for external client connections
         port: 9094
-        # If undefined, `listeners.kafka.external.port` is used
+        # If undefined, `listeners.kafka.external.default.port` is used
```
joejulian, line 616: sanity check... If advertisedPorts is undefined, we advertise "9094" instead of the nodeport port?

joejulian:
I think we should add a warning (see the [warning template](https://github.com/redpanda-data/helm-charts/blob/cc6dd4a0e8316c1780b5070adccdf6b6e896f3df/charts/redpanda/templates/_helpers.tpl#L490-L514)) if LoadBalancer is used warning of latency and packet loss when using cloud load balancers, wdyt?

charts/redpanda/values.yaml
```diff
@@ -115,9 +115,11 @@ external:
   # Default external access value for each service
   # Services can toggle external access per listener (ie. `listeners.<service name>.external.<listener name>.enabled`)
   enabled: true
-  # External access type. `NodePort` is the only acceptable value if defined.
+  # External access type. `NodePort` and `LoadBalancer` are supported
```
alejandroEsc, line 118:
```suggestion
  # External access type. Only `NodePort` and `LoadBalancer` are supported
```

alejandroEsc (approved):
One nit, but not blocking.

charts/redpanda/values.yaml
joejulian: > Related to this, while creating this PR I realized that the schema registry listeners have a port defined, and this is unnecessary.

These additional listeners can have different TLS configurations so they do need their own port, iirc.

charts/redpanda/templates/service.loadbalancer.yaml
```diff
@@ -0,0 +1,88 @@
...
+  namespace: {{ $root.Release.Namespace | quote }}
+  labels:
+    helm.sh/chart: {{ template "redpanda.chart" $root }}
+    app.kubernetes.io/name: {{ template "redpanda.name" $root }}
+    app.kubernetes.io/instance: {{ $root.Release.Name | quote }}
+    app.kubernetes.io/managed-by: {{ $root.Release.Service | quote }}
+    app.kubernetes.io/component: {{ template "redpanda.name" $root }}
+{{- with $root.Values.commonLabels }}
+  {{- toYaml . | nindent 4 }}
+{{- end }}
```
joejulian, line 35:
```suggestion
{{- with include "full.labels" . }}
  {{- . | nindent 4 }}
{{- end }}
```

charts/redpanda/values.schema.json
```diff
@@ -961,7 +961,6 @@
             },
             "external": {
               "type": "object",
-              "minProperties": 1,
```
joejulian, line 964: why can this be removed?

charts/redpanda/templates/service.loadbalancer.yaml
```diff
@@ -0,0 +1,88 @@
...
+      port: {{ dig "nodePort" (first (dig "advertisedPorts" (list $listener.port) $listener)) $listener }}
+  {{- end }}
+{{- end }}
+{{- range $name, $listener := $values.listeners.schemaRegistry.external }}
+  {{- $enabled := dig "enabled" $values.external.enabled $listener }}
+  {{- if $enabled }}
+    - name: schema-{{ $name }}
+      protocol: TCP
+      targetPort: {{ $values.listeners.schemaRegistry.port }}
+      port: {{ dig "nodePort" (first (dig "advertisedPorts" (list $values.listeners.schemaRegistry.port) $listener)) $listener }}
```
joejulian, line 80:
```suggestion
      port: {{ dig "nodePort" (first (dig "advertisedPorts" (list $listener.port) $listener)) $listener }}
```

charts/redpanda/values.yaml
```diff
@@ -626,7 +629,7 @@ listeners:
         # enabled: true
         # The port used for external client connections
         port: 9094
-        # If undefined, `listeners.kafka.external.port` is used
+        # If undefined, `listeners.kafka.external.default.port` is used
```
joejulian, line 632: I also just realized this comment change doesn't make sense. This change makes the comment say:

If `listeners.kafka.external.default.port` is undefined `listeners.kafka.external.default.port` will be used.

joejulian:
Am I correct in assuming that this is still a two-step process where the user is expected to edit the values after installing to apply the loadbalancer hostnames as advertized listeners?

charts/redpanda/values.schema.json
vuldin: This removes the need for a user to have a populated `external` section when defining a schemaRegistry listener. I included this change because I wanted the ability to comment out (or not include) the listener's external configuration, and instead have the templates rely on the default listener values (and the external section) to determine how to configure this listener for external access. This works for because both the LoadBalancer template defaults to values in the following order:
1. `listeners.*.external.*.advertisedPorts[0]`
1. `listeners.*.external.*.port`
1. `listeners.*.port`

https://github.com/redpanda-data/helm-charts/pull/320/files#r1113645104

The other template that uses this section is the Nodeport service, and it default to values in the following order:
1. `listeners.*.external.*.advertisedPorts[0]`
1. `listeners.*.port`

https://github.com/redpanda-data/helm-charts/blob/main/charts/redpanda/templates/services.nodeport.yaml#L66-L67

But a good question would be: why is this included in this PR? I included it just because I was wanting to test the above behavior and the schema wouldn't let me. But it doesn't necessarily need to be in this PR. Another reason for not including it would be that there are lots of changes like this that could be made throughout to ensure users can easily use a subset `values.yaml` file without running into issues.

joejulian: I understand. Yes, let's please do this as a separate PR and as you say, address the usability of this section in general. Would you mind creating an issue so we don't lose the idea?

vuldin: Here is the issue tracking this https://github.com/redpanda-data/helm-charts/issues/359

vuldin:
> Am I correct in assuming that this is still a two-step process where the user is expected to edit the values after installing to apply the loadbalancer hostnames as advertized listeners?

Setting up external access via LoadBalancer doesn't have to be a two-step process. You can set the appropriate values when first deploying and external access will work.

vuldin:
The latest update includes a requested change to print the command a user would need to update advertised listeners once the LoadBalancer services are deployed. This command would upgrade the deployment in order to reconfigure the brokers, and it would set advertised listerners to either the hostname or IP address of each LoadBalancer service. This can be skipped if the user is certain what they want the advertised listeners to be set as, and this would remove the need to upgrade the deployment.

charts/redpanda/templates/NOTES.txt
```diff
@@ -49,6 +49,13 @@ The pods will rollout in a few seconds. To check the status:
 
   kubectl -n {{ .Release.Namespace }} rollout status statefulset {{ template "redpanda.fullname" . }} --watch
 
+{{- if and (toJson (dict "bool" .Values.external.enabled)) (eq .Values.external.type "LoadBalancer") }}
+
+If you are using the load balancer service with a cloud provider, the services will likely have automatically-generated addresses. In this scenario the advertised listeners must be updated in order for external access to work. Run the following command once Redpanda is deployed:
+
+  {{ printf "helm upgrade %s redpanda/redpanda -n %s --set $(kubectl get svc -n %s -o jsonpath='{\"external.addresses={\"}{ range .items[*]}{.status.loadBalancer.ingress[0].ip }{.status.loadBalancer.ingress[0].hostname}{\",\"}{ end }{\"}\\n\"}')" (include "redpanda.name" .) .Release.Namespace .Release.Namespace }}
```
vuldin, line 56: This prints out a command similar to the following:
```
helm upgrade redpanda charts/redpanda -n redpanda --set $(kubectl get svc -n redpanda -o jsonpath='{"external.addresses={"}{ range .items[*]}{.status.loadBalancer.ingress[0].ip }{.status.loadBalancer.ingress[0].hostname}{","}{ end }{"}"}')
```

This is a pretty ugly command and I wish we could resolve the output from `kubectl` rather than having the user run it as a sub-command. This would make the command much easier to understand.

But the issue is the service will be deployed after we ask kubernetes what the service values are, so we won't actually get the correct values.

joejulian (approved):
Looks great. Thanks for all your time and effort on this.

PR #323, opened 2023-02-08
chore: make checking if we are using sasl secret easier to understand

alejandroEsc:
improve readability.


joejulian (approved):
Beautiful!

PR #333, opened 2023-02-10
feat: allow cpu cores to be fractional whole cores and millicores

anapsix:
The Redpanda Helm chart is great in doing validations and ensuring sane values are used. The validation of CPU and Memory requests and limits is a wonderful safety net to ensure best experience with Redpanda. However, users should be able to go outside of recommended values.

While there is a way to override the chart's memory minimums, it's not possible to set CPU value to anything below a full core, due to current values schema constraints, and the way cores are passed to the `--smp` CLI flag.

Let's allow advanced users to experiment with Redpanda in K8s as they are able to do when running Docker images on their local workstations.

* updating values schema to allow integers and strings for `resources.cpu.cores`
* adding `redpanda-cores-in-millis` named templates to convert the `resources.cpu.cores` to millicores
* adding `redpanda-smp` named template to calculate the value passed to `--smp` CLI argument, while defaulting to 1, and making sure to automatically set `resources.cpu.overprovisioned: true` when working with less than 1 full core, and round down to number of full cores when working with fractional cores over 1 full core
* adding `redpanda-cpu-warning` named template to warn on lower than recommended value of cores, and adding it to the list of warnings

Related to https://github.com/redpanda-data/redpanda/issues/350.

Perhaps, when working with less than 1 full core the `--smp` flag should not be passed altogether?
Let's me know and I can make that change.

### DIFF
--- a/charts/redpanda/templates/_helpers.tpl
+++ b/charts/redpanda/templates/_helpers.tpl
@@ -322,18 +322,19 @@ Returns the value of "resources.cpu.cores" in millicores.
 {{- end -}}
 
 {{/*
-Returns the SMP CPU count in whole cores, and sets "overprovisioned: true" when
-the "resources.cpu.cores" is less than 1 core.
+Returns the SMP CPU count in whole cores, with minimum of 1, and sets
+"resources.cpu.overprovisioned: true" when the "resources.cpu.cores" is less
+than 1 core.
 */}}
 {{- define "redpanda-smp" -}}
   {{- $coresInMillies := include "redpanda-cores-in-millis" . | int -}}
-  {{- if lt $coresInMillies 1000 -}}
-    {{- $_ := set $.Values.resources.cpu "overprovisioned" true -}}
-  {{- end -}}
   {{- if $coresInMillies -}}
-    {{- floor (divf $coresInMillies 1000) | int -}}
-  {{- else -}}
+    {{- if lt $coresInMillies 1000 -}}
+      {{- $_ := set $.Values.resources.cpu "overprovisioned" true -}}
+    {{- end -}}
     {{- int "1" -}}
+  {{- else -}}
+    {{- floor (divf $coresInMillies 1000) | int -}}
   {{- end -}}
 {{- end -}}
 
--- a/charts/redpanda/templates/_helpers.tpl
+++ b/charts/redpanda/templates/_helpers.tpl
@@ -312,12 +312,12 @@ Returns the value of "resources.cpu.cores" in millicores.
   {{- $cores := .Values.resources.cpu.cores -}}
   {{- if $cores -}}
     {{- if (hasSuffix "m" (toString $cores)) -}}
-      {{- trimSuffix "m" .Values.resources.cpu.cores | int -}}
+      {{- trimSuffix "m" .Values.resources.cpu.cores -}}
     {{- else -}}
-      {{- mulf 1000.0 ($cores | float64) | int -}}
+      {{- mulf 1000.0 ($cores | float64) -}}
     {{- end -}}
   {{- else -}}
-    {{ int "0" }}
+    {{ "0" }}
   {{- end -}}
 {{- end -}}
 
@@ -332,9 +332,9 @@ than 1 core.
     {{- if lt $coresInMillies 1000 -}}
       {{- $_ := set $.Values.resources.cpu "overprovisioned" true -}}
     {{- end -}}
-    {{- int "1" -}}
+    {{- "1" -}}
   {{- else -}}
-    {{- floor (divf $coresInMillies 1000) | int -}}
+    {{- floor (divf $coresInMillies 1000) -}}
   {{- end -}}
 {{- end -}}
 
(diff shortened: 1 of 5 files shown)


anapsix:
current implementation prohibits setting CPU cores to less than 1 full core

```
❯ helm template redpanda redpanda/redpanda \
  --set config.node.developer_mode=true \
  --set resources.cpu.cores=200m \
  --set resources.memory.container.max=576Mi \
  --set  resources.memory.redpanda.memory=500Mi \
  --set resources.memory.redpanda.reserveMemory=64Mi
redpanda:
- resources.cpu.cores: Invalid type. Expected: integer, given: string

❯ helm template redpanda redpanda/redpanda \
  --set config.node.developer_mode=true \
  --set resources.cpu.cores=0.2 \
  --set resources.memory.container.max=576Mi \
  --set  resources.memory.redpanda.memory=500Mi \
  --set resources.memory.redpanda.reserveMemory=64Mi
redpanda:
- resources.cpu.cores: Invalid type. Expected: integer, given: string
```

charts/redpanda/templates/_helpers.tpl
```diff
@@ -305,6 +305,38 @@ Generate configuration needed for rpk
...
+  {{- if lt $coresInMillies 1000 }}
+    {{- $_ := set $.Values.resources.cpu "overprovisioned" true -}}
+  {{- end -}}
+  {{- if $coresInMillies -}}
+    {{- floor (divf $coresInMillies 1000) | int -}}
+  {{- else -}}
+    {{- int "1" -}}
+  {{- end -}}
+{{- end -}}
+
```
joejulian, line 339: nit: style `{{- ... -}}`

no need to `int "0"` as named templates always return strings you can just use a literal:
```
0
```

Does the change to $.Values persist outside of the named template? I've not tried that before, but I was pretty sure it didn't.

charts/redpanda/templates/_helpers.tpl
```diff
@@ -505,7 +537,7 @@ advertised-host returns a json sring with the data neded for configuring the adv
 */}}
 {{- define "warnings" -}}
   {{- $result := list -}}
-  {{- $warnings := list "redpanda-memory-warning" -}}
+  {{- $warnings := list "redpanda-memory-warning" "redpanda-cpu-warning" -}}
```
joejulian, line 540: thank you

charts/redpanda/templates/_helpers.tpl
anapsix: > no need to int "0"

Sure thing, will update when I have a chance

> Does the change to `$.Values` persist outside of the named templat

Yes. It's magic ✨ (tested with `helm template`). The key is using the `$` context, afaik.

anapsix: > nit: style `{{- ... -}}`

Do you mean around returned value?

joejulian: For the nit: in named templates, unless there's a specific need for newlines, we always use the `-` both front and back.

charts/redpanda/templates/_helpers.tpl
anapsix: weird.. I do see it added in https://github.com/redpanda-data/helm-charts/blob/2bf7ceebcd3e827c87d5a3918346a8217b44efbb/charts/redpanda/templates/_helpers.tpl#L540

charts/redpanda/templates/_helpers.tpl
anapsix: done, I think..
also, found some issues with defaulting logic

charts/redpanda/templates/_helpers.tpl
anapsix: going to mark it as resolved - let me know if you are still seeing missing "redpanda-cpu-warning" in the `$warnings ` list

anapsix: ah.. derp.. nevermind :)

charts/redpanda/templates/_helpers.tpl
```diff
@@ -305,6 +305,39 @@ Generate configuration needed for rpk
...
+{{/*
+Returns the SMP CPU count in whole cores, with minimum of 1, and sets
+"resources.cpu.overprovisioned: true" when the "resources.cpu.cores" is less
+than 1 core.
+*/}}
+{{- define "redpanda-smp" -}}
+  {{- $coresInMillies := include "redpanda-cores-in-millis" . | int -}}
+  {{- if lt $coresInMillies 1000 -}}
+    {{- $_ := set $.Values.resources.cpu "overprovisioned" true -}}
+    {{- 1 -}}
```
joejulian, line 335: This could just be
```suggestion
1
```
if you wanted

joejulian (approved):

charts/redpanda/templates/_helpers.tpl
anapsix: :) I was debating on it, but wanted to keep the leading space padding

anapsix:
@joejulian , I take it the failing version bump check will be ignored by whomever merges the PR? And I should not be bumping it myself?

joejulian:
Feel free to bump it, but if you don't, we'll help.

anapsix:
> Feel free to bump it, but if you don't, we'll help.

@joejulian, is `2.10.0` alright (minor version bump)? Or do you prefer `2.9.2` (patch version bump)?

joejulian:
Good point. New feature so minor bump.

PR #336, opened 2023-02-13
chore: additional nits 

alejandroEsc:


charts/redpanda/templates/secrets.yaml
```diff
@@ -47,13 +47,22 @@ stringData:
...
+{{- if and .Values.auth.sasl.enabled (not (empty .Values.auth.sasl.secretRef )) }}
+    set -e
+    
+    # Setup and export SASL bootstrap-user
+    IFS=":" read -r USER_NAME PASSWORD MECHANISM < /etc/secrets/users/users.txt
+    MECHANISM=${MECHANISM:-{{- include "sasl-mechanism" . }}}
+    rpk acl user create ${USER_NAME} --password=${PASSWORD} --mechanism ${MECHANISM} {{ template "rpk-flags-no-sasl" $ }} || true
+{{- end }}
+
+{{- if gt ( .Values.statefulset.replicas | int64 ) 2 }}
```
joejulian, line 59: Doesn't the need to stay below the `until` loop so the node is ready to accept the command?

alejandroEsc: great point, do not remember why i moved it up. Will revert this part.

alejandroEsc: oh right, it was to avoid the -x in set, let me check on a few things.

joejulian (approved):

PR #338, opened 2023-02-13
feat: adding podLabels option

frgaudet:
- Adding podLabels support

Signed-off-by: Frédéric Gaudet <frederic.gaudet@blablacar.com>

### DIFF
--- a/charts/console/templates/deployment.yaml
+++ b/charts/console/templates/deployment.yaml
@@ -20,6 +20,7 @@ spec:
       {{- end }}
       labels:
         {{- include "console.selectorLabels" . | nindent 8 }}
+        {{- with .Values.podLabels }}{{ toYaml . | nindent 8 }}{{ end }}
     spec:
       {{- with .Values.imagePullSecrets }}
       imagePullSecrets:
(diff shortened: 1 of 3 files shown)


charts/console/templates/deployment.yaml
```diff
@@ -20,6 +20,7 @@ spec:
       {{- end }}
       labels:
         {{- include "console.selectorLabels" . | nindent 8 }}
+        {{- with .Values.podLabels }}{{ toYaml . | nindent 8 }}{{ end }}
```
joejulian, line 23: No need for `with` here.
```suggestion
        {{ toYaml .Values.podLabels | nindent 8 }}
```

joejulian (commented):
Looks good. Just one style change and it's good to merge.

charts/console/templates/deployment.yaml
frgaudet: Fine with me ;)

charts/console/templates/deployment.yaml
```diff
@@ -20,6 +20,7 @@ spec:
       {{- end }}
       labels:
         {{- include "console.selectorLabels" . | nindent 8 }}
+        {{ toYaml .Values.podLabels | nindent 8 }}
```
joejulian, line 23: Ah, my bad, sorry. I was looking at the `with` version thinking _list_. :man_facepalming:

You were right the first time.

frgaudet: My bad as well I didn't really pay attention to your suggestion :)

joejulian (approved):

PR #343, opened 2023-02-15
chore: fix issue where registry may be empty

### LINKED ISSUE #335: Console Deployment - Image registry.
Just Curious!! is it mandatory to pass the image registry variable in values.yaml for the console setup. 

While setting up the console via helm charts, the default image registry is not recognized while performing helm deploy.

image

if i pass image.registry as empty from the values file and the deployment ignores the default image registry configured on k8s.

image

i suggest to remove the image registry from the deployment file or it can be merged into a single variable for the image from the values file..

alejandroEsc:
Fixes #335

### DIFF
--- a/charts/console/templates/deployment.yaml
+++ b/charts/console/templates/deployment.yaml
@@ -60,7 +60,7 @@ spec:
         {{- end }}
           securityContext:
             {{- toYaml .Values.securityContext | nindent 12 }}
-          image: "{{ include "console.imageRegistry" . }}/{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+          image: {{ include "console.container.image" . | quote }}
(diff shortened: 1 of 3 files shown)


charts/console/templates/deployment.yaml
```diff
@@ -60,7 +60,7 @@ spec:
         {{- end }}
           securityContext:
             {{- toYaml .Values.securityContext | nindent 12 }}
-          image: "{{ include "console.imageRegistry" . }}/{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+          image: {{ include "console.container.image" . | quote }}
```
joejulian, line 63: nit: it can only be a string. quotes are unneeded. not a blocker.
```suggestion
          image: {{ include "console.container.image" . }}
```

joejulian (approved):

PR #345, opened 2023-02-16
Rename post install/upgrade job

RafalKorepta:
The post-install suffix is misleading for end user. The job runs not only after install, but after upgrade too. That job can still error out before Admin API is able to serve any request.

### DIFF
--- a/charts/redpanda/templates/post-install-upgrade-job.yaml
+++ b/charts/redpanda/templates/post-install-upgrade-job.yaml
@@ -21,7 +21,7 @@ limitations under the License.
 apiVersion: batch/v1
 kind: Job
 metadata:
-  name: {{ template "redpanda.fullname" . }}-post-install
+  name: {{ template "redpanda.fullname" . }}-configuration
   namespace: {{ .Release.Namespace | quote }}
   labels:
 {{- with include "full.labels" . }}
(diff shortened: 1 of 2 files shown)


charts/redpanda/templates/post-install-upgrade-job.yaml
```diff
@@ -21,7 +21,7 @@ limitations under the License.
 apiVersion: batch/v1
 kind: Job
 metadata:
-  name: {{ template "redpanda.fullname" . }}-post-install
+  name: {{ template "redpanda.fullname" . }}-configurator
```
alejandroEsc, line 24: Does this do more bootstrap configuration? How about `bootstrap-configurator`?

joejulian: It re-configures things, too, which is why it's no longer just post-install.

If we're going to be pedantic, let's use real words: configuration, cluster-configuration, something like that?

alejandroEsc: Alright i think configuration makes more sense. Do we expect SASL superusers list to be ever changed/updated/appended in the future? Just making a mental note.

RafalKorepta: > Do we expect SASL superusers list to be ever changed/updated/appended in the future?

I think we should allow that. User should be able to change list of superusers including:
* changing the password
* removing or replacing superusers

alejandroEsc (approved):
Lgtm

PR #346, opened 2023-02-16
redpanda: allow the ability to skip the chown initContainer

joejulian:
When a user is in an environment that restricts root containers, allow them to skip the chown initContainer as it won't work without being run as root.


alejandroEsc (approved):
lgtm

PR #347, opened 2023-02-16
redpanda: fix console bug where deployment and configmap are erroneously created

joejulian:
When console.enabled is false, the console deployment and configmap would still be created.


alejandroEsc (approved):
Nice nits

PR #351, opened 2023-02-17
Fix release redpanda chart version

RafalKorepta:
By merging 2 PRs quickly CI can not release second change as the Chart version is the same.

REF:
https://github.com/redpanda-data/helm-charts/pull/345
https://github.com/redpanda-data/helm-charts/pull/272


alejandroEsc (approved):
lgtm, yes this is a problem i noticed yesterday and this was the solution i was thinking of, thank you

PR #356, opened 2023-02-23
redpanda: fix license_key regex

joejulian:
The regex for the license_key should match `<base64>.<base64>`


RafalKorepta (approved):

alejandroEsc (approved):
lgtm

PR #360, opened 2023-02-24
feat: change podAntiAffinity to allow for customer anti-affinity rules

### LINKED ISSUE #294: Not possible to set custom podAntiAffinity rules for redpanda
It is mentioned here that it is possible to set custom set of podAntiAffinity rules:
https://github.com/redpanda-data/helm-charts/blob/a7f9e4324bfbcc9e665916a7360cd99a3aab4c76/charts/redpanda/values.yaml#L328-L336

And according to the statefulset template file it is supported when "type" param is not provided:
https://github.com/redpanda-data/helm-charts/blob/a7f9e4324bfbcc9e665916a7360cd99a3aab4c76/charts/redpanda/templates/statefulset.yaml#L384

However, type has "soft" as default value and so it ignores the provided rules (they do not end up in the sts resource) even when type is not set explicitly. 
When type is set to empty string or null validation fails:

```
helm upgrade --install --debug -f helm/redpanda.yaml redpanda redpanda/redpanda
history.go:56: [debug] getting history for release redpanda
upgrade.go:142: [debug] preparing upgrade for redpanda
Error: UPGRADE FAILED: values don't meet the specifications of the schema(s) in the following chart(s):
redpanda:
- statefulset.podAntiAffinity.type: Does not match pattern '^(hard|soft)$'

helm.go:84: [debug] values don't meet the specifications of the schema(s) in the following chart(s):
redpanda:
- statefulset.podAntiAffinity.type: Does not match pattern '^(hard|soft)$'
```

So an easy solution would be to fix the `values.schema.json` so that it allows empty string and/or null:
https://github.com/redpanda-data/helm-charts/blob/main/charts/redpanda/values.schema.json#L519-L522

alejandroEsc:
Allow for custom podAntiAffinity rules. Currently the logic in place makes little sense. Nor is it clear where and how  your own rules can be placed. Here we allow for a mechanism that allows for removal of the rules complete, allows for the two defaults, and allows for a custom options. 

Fixes #294

### DIFF
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 2.12.0
+version: 2.12.1
 
 # The app version is the default version of Redpanda to install.
 appVersion: v22.3.13
--- a/charts/redpanda/ci/07-multiple-listeners-values.yaml
+++ b/charts/redpanda/ci/07-multiple-listeners-values.yaml
@@ -1,3 +1,18 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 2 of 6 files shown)


joejulian (approved):

joejulian (changes requested):
need a test

joejulian (approved):

PR #361, opened 2023-02-24
redpanda: add retries to all the tests

joejulian:
Convert the Job tests back to pods so we can get the logs. Add retries and timeouts for every test so they can succeed without Jobs.

### DIFF
--- a/charts/redpanda/templates/tests/test-pandaproxy-internal-tls-status.yaml
+++ b/charts/redpanda/templates/tests/test-pandaproxy-internal-tls-status.yaml
@@ -14,21 +14,21 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */}}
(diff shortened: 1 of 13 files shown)


charts/redpanda/templates/tests/test-pandaproxy-internal-tls-status.yaml
```diff
@@ -14,32 +14,38 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
     runAsUser: 65535
     runAsGroup: 65535
   containers:
     - name: {{ template "redpanda.name" . }}
-      image: {{ .Values.image.repository}}:{{ template "redpanda.tag" . }}
+      image: {{ .Values.image.repository }}:{{ template "redpanda.tag" . }}
       command:
         - curl
         - -svm3
+        - --fail
```
alejandroEsc, line 43: nice!

alejandroEsc (approved):
lgtm, thank you for doing this, its basically makes the tests a single style.

PR #363, opened 2023-02-25
Update console app version

weeco:
Change app version to v2.2.0

Maybe there's a way to auto-update this based on the latest docker tags longer-term.


RafalKorepta (commented):
We should bump the console dependency in redpanda chart.

RafalKorepta:
Turns out we have fancy regexp `">=0.5 <1.0"`
https://github.com/redpanda-data/helm-charts/blob/f8f8533ae68f57473b093eb1bc2b1101cb72387d/charts/redpanda/Chart.yaml#L35-L39

RafalKorepta (approved):

PR #374, opened 2023-03-03
[experimental] autotune changes to run on postStart

alejandroEsc:
This is just a test PR to share a possible way we could run tunning. 

This can eliminate the priv initContainer and run several tunners on the pod that runs redpanda. 

We avoid having to exit with non-zero postStart process for now by silently failing, we also allow to setup the mode in the values file to get a list of possible tuners. 

Note that we are able to create the ballast file here and the tunners were found to run as intended. Not all tuners make sense to run and there is questions about how to handle tune_disk_irq at the moment. 

There may be some specific container settings (which maybe we can change in our image creation tools) like swappiness that expects to be able to write on a file that it does not have permissions to write.

### DIFF
--- a/charts/redpanda/templates/secrets.yaml
+++ b/charts/redpanda/templates/secrets.yaml
@@ -47,6 +47,7 @@ stringData:
     # path below should match the path defined on the statefulset
     source /var/lifecycle/common.sh
 
+    # Clearing out of maintenance mode
 {{- if gt ( .Values.statefulset.replicas | int64 ) 2 }}
     set -ex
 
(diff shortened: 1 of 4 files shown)


joejulian:
The only disadvantage that comes to mind is that the postStart logging in Kubernetes is garbage. This would make diagnostics harder.

alejandroEsc:
> The only disadvantage that comes to mind is that the postStart logging in Kubernetes is garbage. This would make diagnostics harder.

Yep. It may be worth to try to store results to files. Not sure though.

RafalKorepta (commented):
Is PostStart running as privileged process? I don't understand how it could run

alejandroEsc:
> Is PostStart running as privileged process? I don't understand how it could run

No its not, but I am only running those items that landed green during my testing, including the ballast file. I think I would prefer this type of approach to tunning only things we can reasonably expect to tune. Anything else we should have special additional software to run with priv containers. I also think these should be jobs not daemonsets, but we can talk about that later today.

alejandroEsc:
I do think we should keep some things running in the main pod, but up for discussion in the future. Closing atm.

PR #378, opened 2023-03-03
Add annotations to external services

vuldin:
- add external.annotations to schema
- add external.annotations to LoadBalancer service
- add external.annotations to NodePort service


alejandroEsc (approved):
lgtm

joejulian:
I think I'd prefer these be configurable by service.

In part because I just realized that we're probably going to want a service-per-listener<dot external name>. If they define multiple external listeners, and have some of those exposed through load balancers, they probably will want different listener groups with different annotations.

alejandroEsc:
> I think I'd prefer these be configurable by service.
> 
> In part because I just realized that we're probably going to want a service-per-listener. If they define multiple external listeners, and have some of those exposed through load balancers, they probably will want different listener groups with different annotations.

Now that's a very good idea. Wouldnt that require a substantial refactor at this point?

joejulian:
> > I think I'd prefer these be configurable by service.
> > In part because I just realized that we're probably going to want a service-per-listener. If they define multiple external listeners, and have some of those exposed through load balancers, they probably will want different listener groups with different annotations.
> 
> Now that's a very good idea. Wouldnt that require a substantial refactor at this point?

It would, but for now just allowing annotations to be set anywhere we allow the service type to be set would satisfy.

joejulian (approved):

PR #380, opened 2023-03-06
chore: remove chown initContainer 

### LINKED ISSUE #370: Remove set-datadir-ownership as it appears to no longer be necessary
### What happened?

Our use of chown to fixup datadir directory ownership throws errors in Kubernetes environments where containers aren't allowed to run privileged. This is happening at least with Rancher/rke2.

```CreateContainerConfigError (container has runAsNonRoot and image will run as root (pod: "redpanada-0_redpanda(04835735-5d80-4d48-bad5-fa2cfbe6c39c)", container: set-datadir-ownership))

Which comes from this section 

        - name: set-datadir-ownership
          image: {{ .Values.statefulset.initContainerImage.repository }}:{{ .Values.statefulset.initContainerImage.tag }}
          command: ["/bin/sh", "-c", "chown {{ $uid }}:{{ $gid }} -R /var/lib/redpanda/data"]
          volumeMounts:
            - name: datadir
```

It appears that part of this was fixed up in https://github.com/redpanda-data/helm-charts/issues/173 but the chown was left around.

Don't throw an error. This appears to be unnecessary given the runAsUser/runAsGroup/fsGroup support that got added previously.

N/A. Requires a restricted K8s environment to replicate.

_No response_

Redpanda

### Chart Version(s)

2.12.0

### Cloud provider

N/A.

On-prem Rancher/rke2.

alejandroEsc:
and add fsGroupchangePolicy on mismatch by default. 

Fixes #370

### DIFF
--- a/charts/redpanda/values.yaml
+++ b/charts/redpanda/values.yaml
@@ -533,10 +533,9 @@ statefulset:
   securityContext:
     fsGroup: 101
     runAsUser: 101
+    fsGroupChangePolicy: OnRootMismatch
   initContainers:
     tuning:
-      resources: {}
-    setDatadirOwnership:
       resources: {}
     setTieredStorageCacheDirOwnership:
       resources: {}
@@ -545,9 +544,6 @@ statefulset:
   initContainerImage:
     repository: busybox
     tag: latest
-  # in environments where root is not allowed, you cannot change the ownership of files and directories
-  # set this to skip this step
-  skipChown: false
 
 # Service account management
 serviceAccount:
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 2.13.0
+version: 2.13.1
 
 # The app version is the default version of Redpanda to install.
 appVersion: v22.3.13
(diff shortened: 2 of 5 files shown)


charts/redpanda/values.yaml
```diff
@@ -533,21 +533,17 @@ statefulset:
...
     setTieredStorageCacheDirOwnership:
       resources: {}
     configurator:
       resources: {}
   initContainerImage:
     repository: busybox
     tag: latest
-  # in environments where root is not allowed, you cannot change the ownership of files and directories
-  # set this to skip this step
-  skipChown: false
```
RafalKorepta, line 550: Is this `skipChown` part of `values.schema.json`? The helm upgrade with —reuse-values will brake or will just left the unused properties?

alejandroEsc: Thank you for bringing that up! I meant to delete this from the schema, so thank you for bringing that up.

 So using --reuse-values will not cause an issue as the value and checking the values file after upgrading to this chart

```
  securityContext:
    fsGroup: 101
    fsGroupChangePolicy: OnRootMismatch
    runAsUser: 101
  skipChown: false
```

```
redpanda    	redpanda    	2       	2023-03-07 10:13:54.993363145 -0500 EST	deployed	redpanda-2.13.1     	v22.3.13   
```

RafalKorepta (approved):

PR #384, opened 2023-03-08
test: add rpk debug bundle test

r-vasquez:
This test needs to peak the cluster data-dir
that's why we mount to the redpanda pvc.

We don't test every single file, instead we check
for existence of some important files that were
added to the k8s-rpk debug bundle.


alejandroEsc:
@r-vasquez feel free to rebase, the new image is now in place.

RafalKorepta (approved):

PR #385, opened 2023-03-08
fix: if ca is not enabled, do not pass the cacert curl flag

alejandroEsc:
Small bug fix were we should not add cacert flag in the situation where ca is disabled. 

This should be the last small fix that we can commit without causing breaking changes. I will then close PR #289

### DIFF
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 2.13.2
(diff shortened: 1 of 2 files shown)


joejulian:
In my mind, the internal tls must always have the CA flag. There's no reason to ever have an internal address provided by an external entity that I can see.

alejandroEsc:
> In my mind, the internal tls must always have the CA flag. There's no reason to ever have an internal address provided by an external entity that I can see.

That is correct, but it will be an issue on the external ca which will not always be the case that an external ca exists.

alejandroEsc:
Not that important right now. I think we are in the process or will solve this where it will not be a concern.

PR #386, opened 2023-03-08
fix: should check initContainer field exists first to support upgrades

alejandroEsc:
Bug introduced for not checking from previous revisions if the initContainer.container field exists before checking if the subfield exists.

In addition we should also not require a new field as this will cause problems on upgrades with --reuse-values flag set.

### DIFF
--- a/charts/redpanda/templates/statefulset.yaml
+++ b/charts/redpanda/templates/statefulset.yaml
@@ -71,7 +71,7 @@ spec:
           volumeMounts:
             - name: {{ template "redpanda.fullname" . }}
               mountPath: /etc/redpanda
-        {{- if .Values.statefulset.initContainers.tuning.resources }}
+        {{- if get .Values.statefulset.initContainers.tuning "resources" }}
           resources: {{- toYaml .Values.statefulset.initContainers.tuning.resources | nindent 12 }}
         {{- end }}
 {{- end }}
@@ -82,7 +82,7 @@ spec:
           volumeMounts:
             - name: tiered-storage-dir
               mountPath: {{ template "tieredStorage.cacheDirectory" . }}
-        {{- if .Values.statefulset.initContainers.setTieredStorageCacheDirOwnership.resources }}
+        {{- if get .Values.statefulset.initContainers.setTieredStorageCacheDirOwnership "resources" }}
           resources: {{- toYaml .Values.statefulset.initContainers.setTieredStorageCacheDirOwnership.resources | nindent 12 }}
         {{- end }}
 {{- end }}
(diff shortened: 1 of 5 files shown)


charts/redpanda/templates/statefulset.yaml
```diff
@@ -71,7 +71,7 @@ spec:
           volumeMounts:
             - name: {{ template "redpanda.fullname" . }}
               mountPath: /etc/redpanda
-        {{- if .Values.statefulset.initContainers.tuning.resources }}
+        {{- if and .Values.statefulset.initContainers.tuning .Values.statefulset.initContainers.tuning.resources }}
```
joejulian, line 74:
```suggestion
        {{- if get .Values.statefulset.initContainers.tuning "resources" }
```

RafalKorepta (approved):

charts/redpanda/templates/statefulset.yaml
```diff
@@ -82,7 +82,7 @@ spec:
           volumeMounts:
             - name: tiered-storage-dir
               mountPath: {{ template "tieredStorage.cacheDirectory" . }}
-        {{- if .Values.statefulset.initContainers.setTieredStorageCacheDirOwnership.resources }}
+        {{- if and .Values.statefulset.initContainers.setTieredStorageCacheDirOwnership .Values.statefulset.initContainers.setTieredStorageCacheDirOwnership.resources }}
```
joejulian, line 85:
```suggestion
        {{- if get .Values.statefulset.initContainers.setTieredStorageCacheDirOwnership "resources" }}
```

charts/redpanda/templates/statefulset.yaml
```diff
@@ -176,7 +176,7 @@ spec:
               mountPath: /tmp/base-config
             - name: config
               mountPath: /etc/redpanda
-        {{- if .Values.statefulset.initContainers.configurator.resources }}
+        {{- if and .Values.statefulset.initContainers.configurator .Values.statefulset.initContainers.configurator.resources }}
```
joejulian, line 179:
```suggestion
        {{- if get .Values.statefulset.initContainers.configurator "resources" }}
```

joejulian (commented):
Can we use `get` instead of `and`? Using `and` won't get the result you're looking for with helm versions that were built with go before go1.19

alejandroEsc:
> Can we use `get` instead of `and`? Using `and` won't get the result you're looking for with helm versions that were built with go before go1.19

Of course, ill ping you later about the `and` and versiosn before go1.19, that is very interesting.

RafalKorepta (approved):

Issue #387, opened 2023-03-10
Fatal error during checker "Data directory is writable", permission denied


vuldin:
### What happened?

I'm trying to deploy with the following command on minikube:
```
helm upgrade --install redpanda redpanda/redpanda -n redpanda --create-namespace --wait
```
And get the following issue:
```
> kubectl logs -n redpanda redpanda-0 -c redpanda
Error: fatal error during checker "Data directory is writable" execution: open /var/lib/redpanda/data/test_file: permission denied
```

### What did you expect to happen?

The chart should have successfully deployed Redpanda.

### How can we reproduce it (as minimally and precisely as possible)?. Please include values file.

<details>

```console
$ helm get values <redpanda-release-name> -n <redpanda-release-namespace> --all
COMPUTED VALUES:              
auth:                             
  sasl:                              
    enabled: false             
    mechanism: SCRAM-SHA-512           
    secretRef: redpanda-users     
    users:                                   
    - mechanism: SCRAM-SHA-512               
      name: admin         
      password: change-me              
clusterDomain: cluster.local             
commonLabels: {}                                 
config:                                   
  cluster: {}                               
  node: {}                            
  tunable:                       
    compacted_log_segment_size: 67108864
    group_topic_partitions: 16  
    kafka_batch_max_bytes: 1048576
    kafka_connection_rate_limit: 1000
    log_segment_size: 134217728
    log_segment_size_max: 268435456    
    log_segment_size_min: 16777216
    max_compacted_log_segment_size: 536870912
    topic_partitions_per_shard: 1000         
console:                  
  affinity: {}                         
  autoscaling:                           
    enabled: false                               
    maxReplicas: 100                      
    minReplicas: 1
    targetCPUUtilizationPercentage: 80
  config: {}
  configmap:
    create: false
  console:
    config: {}
  deployment:
    create: false
  enabled: true
  extraContainers: {}
  extraEnv: []
  extraEnvFrom: []
  extraVolumeMounts: []
  extraVolumes: []
  fullnameOverride: ""
  global: {}
  image:
    pullPolicy: IfNotPresent
    registry: docker.redpanda.com
    repository: vectorized/console
    tag: ""
  imagePullSecrets: []
  ingress:
    annotations: {}
    className: ""
    enabled: false
    hosts:
    - host: chart-example.local
      paths:
      - path: /
        pathType: ImplementationSpecific
    tls: []
  livenessProbe:
    failureThreshold: 3
[...]

PR #391, opened 2023-03-16
chore: add task file for common operations

alejandroEsc:
Sorry it took a while, hope this helps, it is only the first pass. 

```
$ task --list-all
task: Available tasks for this project:
* install-redpanda-chart:         
* kind-create:                    
* kind-delete:                    
* minikube-delete:                
* minikube-start:                 
* uninstall-redpanda-chart:       
* upgrade-redpanda-chart:  
```

Partially solves #387

### DIFF
--- a/Taskfile.yaml
+++ b/Taskfile.yaml
@@ -24,7 +24,7 @@ tasks:
 
   install-redpanda-chart:
     cmds:
-      - helm install redpanda ./charts/redpanda --namespace redpanda --create-namespace --debug
+      - helm install rp ./charts/redpanda --namespace rn --create-namespace --debug
 
   upgrade-redpanda-chart:
     cmds:
--- a/Taskfile.yaml
+++ b/Taskfile.yaml
@@ -15,12 +15,12 @@ run: once
 tasks:
   kind-create:
     cmds:
-      - kind create cluster --config ./.github/kind.yaml
+      - kind create cluster --name rp-helm --config ./.github/kind.yaml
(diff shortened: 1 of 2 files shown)


Taskfile.yaml
```diff
@@ -0,0 +1,39 @@
...
+# file, or in a task's `env:` attribute.
+
+
+# if a task is referenced multiple times, only run it once
+run: once
+
+tasks:
+  kind-create:
+    cmds:
+      - kind create cluster --config ./.github/kind.yaml
```
joejulian, line 18: Should we add a name flag here instead of using the default?

Taskfile.yaml
```diff
@@ -0,0 +1,39 @@
...
+      - helm install cert-manager jetstack/cert-manager --set installCRDs=true --namespace cert-manager --create-namespace
+
+  kind-delete:
+    cmds:
+      - kind delete cluster
+
+  install-redpanda-chart:
+    cmds:
+      - helm install redpanda ./charts/redpanda --namespace redpanda --create-namespace --debug
```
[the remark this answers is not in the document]
alejandroEsc: This is a task... its up to the person who is running it, but yeah we can take those in. Next iteration was going to make those env vars but I wanted to verify defaulting etc.. works as expected with tasks.

Taskfile.yaml
alejandroEsc: Not sure, dow we? usually i just use the name kind, but sure.

RafalKorepta (approved):

joejulian (approved):

PR #392, opened 2023-03-17
Fix: Duplicated securityContext fails on helm template/ kustomize build

Alegrowin:
kustomize build app --enable-helm 
Error: map[string]interface {}(nil): yaml: unmarshal errors:
  line 62: mapping key "securityContext" already defined at line 27
error: no objects passed to apply


RafalKorepta:
Hi, Thank you for the contribution. Could you sign CLA?

alejandroEsc:
We also need to bump the chart version:

in charts/redpanda/Chart.yaml, to: 
```
version: 3.0.4
```

alejandroEsc (approved):
lgtm!

PR #393, opened 2023-03-17
charts/redpanda: Fix CAP_SYS_RESOURCE name

c4milo:
All linux capabilities in k8s security context drop the CAP prefix.

Ref: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-capabilities-for-a-container

> Note: Linux capability constants have the form CAP_XXX. But when you list capabilities in your container manifest, you must omit the CAP_ portion of the constant. For example, to add CAP_SYS_TIME, include SYS_TIME in your list of capabilities.

### DIFF
--- a/charts/redpanda/templates/statefulset.yaml
+++ b/charts/redpanda/templates/statefulset.yaml
@@ -68,7 +68,7 @@ spec:
               rpk redpanda tune all
           securityContext:
             capabilities:
-              add: ["CAP_SYS_RESOURCE"]
+              add: ["SYS_RESOURCE"]
             privileged: true
             runAsUser: 0
             runAsGroup: 0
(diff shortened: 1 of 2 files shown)


charts/redpanda/templates/statefulset.yaml
```diff
@@ -64,7 +64,7 @@ spec:
               rpk redpanda tune all
           securityContext:
             capabilities:
-              add: ["CAP_SYS_RESOURCE"]
```
alejandroEsc, line 67: Can you add a comment here along with the link you provided. This would help cover future questions about this. Thank you.

alejandroEsc (commented):
Please Bump chart version, i believe the next chart version in

`/charts/redpanda/Chart.yaml` is:

```
version: 3.0.5
```

RafalKorepta (approved):

PR #395, opened 2023-03-21
redpanda: do not set azure keys to empty strings

joejulian:
Redpanda does not interpret empty strings as being unset. Must use null.


joejulian:
@rtluckie

alejandroEsc (approved):
Read over the inc. Lgtm

RafalKorepta (approved):

joejulian:
No. We have an issue in the backlog for testing on the cloud providers but right now all our tests are in kind.

PR #397, opened 2023-03-21
fix: revert to tls enabled by default

alejandroEsc:
@vuldin found that this was set to false, tls should be enabled by default.


vuldin (approved):
LGTM

joejulian (approved):

PR #401, opened 2023-03-21
set up test matrix to provider better test coverage in less time

joejulian:


.github/workflows/pull_requests.yaml
```diff
@@ -55,13 +55,30 @@ jobs:
...
     steps:
       - name: Checkout
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
 
+      - name: Set up for matrix
+        run: |
+          echo bash -O extglob -c "rm -v charts/redpanda/ci/!(${{ matrix.testvaluepattern }})"
+          bash -O extglob -c "rm -v charts/redpanda/ci/!(${{ matrix.testvaluepattern }})"
```
RafalKorepta, line 80: Maybe thanks to -x like `bash -O extglob -x -c "rm -v charts/redpanda/ci/!(${{ matrix.testvaluepattern }})"` you could drop echo, right?

RafalKorepta (approved):

PR #402, opened 2023-03-22
Remove examples folder

vuldin:
This PR removes the examples folder. This is done because we have made many improvements and additions to our [official docs](https://docs.redpanda.com/docs/home), including to much of the content that was included (or planned for inclusion) here. It also creates a `contrib` folder for any example contributions that relate to the helm charts but may not make sense to live in our official docs.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1204218351694990


joejulian (approved):

joejulian:
We should check with @JakeSCahill before merging in case there are any docs that point to the examples.

JakeSCahill (approved):

PR #410, opened 2023-03-28
Use tpl render on external domain name for providing values dynamically

abhilashjoseph:
## Motivation
External access might require dynamic domain names to be used as the ADVERTISED DOMAIN names

## Approach
Add template rendering using tpl functions for external domain value

## Usecase
- When external access is enabled for the Redpanda service, the domain names are advertised by the brokers based on the Values.external.domain configuration. 
- For internal clients the cluster local endpoints work just fine, but
- Currently we do not have an option to dynamically set the external domains as we might have multiple environments or stacks were Redpanda is deployed (dev/qa/prod etc). 
- Having the external domain name configurable through values file and tpl rendered, we could use helm library functions to override the domain names appropriately for the stacks that we deploy Redpanda.

### DIFF
--- a/charts/redpanda/templates/_helpers.tpl
+++ b/charts/redpanda/templates/_helpers.tpl
@@ -534,7 +534,7 @@ advertised-host returns a json sring with the data neded for configuring the adv
 {{- define "advertised-host" -}}
   {{- $host := dict "name" .externalName "address" .externalAdvertiseAddress "port" .port -}}
   {{- if .values.external.addresses -}}
-    {{- if .values.external.domain -}}
+    {{- if (tpl ($values.external.domain | default "") $) }}
       {{- $host = dict "name" .externalName "address" (printf "%s.%s" (index .values.external.addresses .replicaIndex) (tpl .values.external.domain $)) "port" .port -}}
     {{- else -}}
       {{- $host = dict "name" .externalName  "address" (index .values.external.addresses .replicaIndex) "port" .port -}}
--- a/charts/redpanda/templates/_helpers.tpl
+++ b/charts/redpanda/templates/_helpers.tpl
@@ -534,7 +534,7 @@ advertised-host returns a json sring with the data neded for configuring the adv
 {{- define "advertised-host" -}}
   {{- $host := dict "name" .externalName "address" .externalAdvertiseAddress "port" .port -}}
   {{- if .values.external.addresses -}}
-    {{- if (tpl ($values.external.domain | default "") $) }}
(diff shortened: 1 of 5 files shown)


joejulian:
Would you please describe a use case, either in the description or an associated issue so it can be properly documented?

abhilashjoseph:
> Would you please describe a use case, either in the description or an associated issue so it can be properly documented?

Thank you @joejulian, for looking into this, I've updated the description with the use case for this capability.

joejulian:
The Chart.yaml needs a patch version bump and the upgrade test is failing:
```
 Error: UPGRADE FAILED: template: redpanda/templates/certs.yaml:52:19: executing "redpanda/templates/certs.yaml" at <$values.external.domain>: wrong type for value; expected string; got interface {}
```

abhilashjoseph:
> The Chart.yaml needs a patch version bump and the upgrade test is failing:
> 
> ```
>  Error: UPGRADE FAILED: template: redpanda/templates/certs.yaml:52:19: executing "redpanda/templates/certs.yaml" at <$values.external.domain>: wrong type for value; expected string; got interface {}
> ```

@joejulian bumped up the chart patch version, also fixed the issue with the default value test, all tests have passed now, thanks!

charts/redpanda/templates/statefulset.yaml
```diff
@@ -19,7 +19,7 @@ limitations under the License.
 {{- $internalAdvertiseAddress := printf "%s.%s" "$(SERVICE_NAME)" (include "redpanda.internal.domain" .) -}}
 {{- $externalAdvertiseAddress := printf "$(SERVICE_NAME)" -}}
 {{- if $values.external.domain -}}
```
joejulian, line 21:
```suggestion
{{- if (tpl ($values.external.domain | default "") $) }}
```

charts/redpanda/templates/_helpers.tpl
```diff
@@ -535,7 +535,7 @@ advertised-host returns a json sring with the data neded for configuring the adv
   {{- $host := dict "name" .externalName "address" .externalAdvertiseAddress "port" .port -}}
   {{- if .values.external.addresses -}}
     {{- if .values.external.domain -}}
```
joejulian, line 537:
```suggestion
    {{- if (tpl ($values.external.domain | default "") $) }}
```

abhilashjoseph:
@joejulian committed suggestions as well fixed a var typo, please verify, thank you!

joejulian (approved):
Thanks!

PR #417, opened 2023-03-29
Decrease cloud_storage_cache_size default value

### LINKED ISSUE #416: Dynamic tiered storage cache sizing
### What happened?

`cloud_storage_cache_size` is set to the same default size of the persistent disks. This means cache can possibly fill up the entire disk.

This value should be calculated as a % of available local disk space, taking guidance from our new cloud tiers (including high throughput tiers). The cache should optionally also be created as its own persistent volume to isolate this storage from others.

<details>

```console
$ helm get values <redpanda-release-name> -n <redpanda-release-namespace> --all
# paste output here
```

</details>

_No response_

Redpanda

### Chart Version(s)

<details>

```console
$ helm -n <redpanda-release-namespace> list 
# paste output here
```

</details>

### Cloud provider

<details>

</details>

vuldin:
Closes #416 

This value at it's current default of 20Gi is set to the same size as the default value for a persistent volume's size. This means that if consumers start consuming data from offset 0 and there is enough cloud storage data (over 20Gi), then the disk would be entirely consumed by cloud storage alone and crash the broker.

This PR sets the cloud storage cache storage value to a better default of 5Gi.

### DIFF
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 4.0.26
+version: 4.0.27
 
 # The app version is the default version of Redpanda to install.
 appVersion: v23.1.10
--- a/charts/redpanda/values.yaml
+++ b/charts/redpanda/values.yaml
@@ -411,7 +411,7 @@ storage:
     # -- Maximum size of the disk cache used by Tiered Storage.
     # Default is 20 GiB.
     # See the [property reference documentation](https://docs.redpanda.com/docs/reference/cluster-properties/#cloud_storage_cache_size).
-    cloud_storage_cache_size: 21474836480
+    cloud_storage_cache_size: 5368709120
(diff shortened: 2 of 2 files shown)


joejulian:
There was a discussion that suggested that this should also be accompanied by a change to tiered storage such that this value is computed rather than just set. Did we want to do that here, or should it be a separate PR?

vuldin:
> suggested that this should also be accompanied by a change to tiered storage such that this value is computed rather than just set. Did we want to do that here
I think it makes sense to include that change here. Another suggestion was to make the separate cloud storage PV enabled by default. Would we also want to include that related change here?

joejulian:
> > suggested that this should also be accompanied by a change to tiered storage such that this value is computed rather than just set. Did we want to do that here

>  I think it makes sense to include that change here. Another suggestion was to make the separate cloud storage PV enabled by default. Would we also want to include that related change here?

Seems like a good idea to me.

joejulian:
We're just going to accept this as-is and add an issue for the more sophisticated improvement: https://github.com/redpanda-data/helm-charts/issues/530

joejulian (approved):

PR #423, opened 2023-04-04
Added missing namespace to StatefulSet chart template.

pfcurtis:
Added missing namespace to Stateful set chart template.


pfcurtis:
Closed to correct user & email to complete the agreements properly.

PR #424, opened 2023-04-06
use org bot credentials

joejulian:
redpanda-data already has credentials that can be used for this. This uses those credentials rather than adding new ones.


alejandroEsc (approved):
lgtm

PR #425, opened 2023-04-07
Templating should occur external to a helper function

alejandroEsc:
We have seen an issue trying to set the external.addresses in the values file:

```
redpanda/templates/statefulset.yaml:120:23: executing "redpanda/templates/statefulset.yaml" at <include "advertised-host" (mustMerge $tmplVals (dict "port" $port))>: error calling include: template: redpanda/templates/_helpers.tpl:537:12: executing "advertised-host" at <tpl (.values.external.domain | default "") $>: error calling tpl: cannot retrieve Template.Basepath from values inside tpl function: : "BasePath" is not a value
```

This seems to be because 

https://github.com/helm/helm/issues/4166#issuecomment-497430905

It is not possible to get the BasePath scope into the template file. To fix that we move the templating above to the statefulset. I also added a test to help catch issues in the future on this particular code path.

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -51,6 +51,7 @@ jobs:
           - '0[1-3]*'
           - '0[4-6]*'
           - '0[7-9]*'
+          - '1[0-2]*'
       fail-fast: true
     runs-on: ubuntu-22.04
     steps:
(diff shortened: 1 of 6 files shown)


joejulian (approved):
Good catch!

RafalKorepta (approved):

PR #427, opened 2023-04-07
use vars for consistent naming in tasks

joejulian:
I noticed that the chart install and upgrade used different release names. Moved several things to vars to allow command-line overrides if desired, ie:

```
task kind-create KIND_CLUSTERNAME=foo
```


alejandroEsc (approved):
lgtm! beat me to it.

Issue #433, opened 2023-04-11
Redpanda does not receive SIGTERM on pod termination


travisdowns:
### What happened?

Redpanda process does not received SIGTERM on pod termination. This prevents a graceful shutdown as it will eventually receive only a SIGKILL.

### What did you expect to happen?

SIGTERM to Redpanda.

### How can we reproduce it (as minimally and precisely as possible)?. Please include values file.

https://redpandadata.slack.com/archives/C01H6JRQX1S/p1681230094226129

### Anything else we need to know?

Underlying issue is that `bash -c` is used to set up a pipeline which means `redpanda` isn't PID 1, so doesn't get the SIGTERM.

### Which are the affected charts?

redpanda

### Chart Version(s)

as of now

### Cloud provider

<details>
any
</details>

PR #435, opened 2023-04-12
bash does not pass sigterm to child processes

joejulian:
this causes Redpanda to be sigterm'd, which might be causing corrupted
segments.

bash was in use because we were filtering out secret data from the
Redpanda logs. All currently supported versions of Redpanda no longer
leak that data.

This PR adds a test for Redpanda versions that have the secret leak problem and will fail if an insecure version is being used.

Fixes: #433

### DIFF
--- a/charts/redpanda/templates/statefulset.yaml
+++ b/charts/redpanda/templates/statefulset.yaml
@@ -295,16 +295,14 @@ spec:
             periodSeconds: {{ .Values.statefulset.readinessProbe.periodSeconds }}
             successThreshold: {{ .Values.statefulset.readinessProbe.initialDelaySeconds }}
           command:
-            - bash
-            - -c
-            - |
-              rpk redpanda start \
-                --smp={{ include "redpanda-smp" . }} \
-                --memory={{ template "redpanda-memory" . }}M \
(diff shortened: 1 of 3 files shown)


charts/redpanda/templates/statefulset.yaml
```diff
@@ -295,16 +297,14 @@ spec:
...
-            - -c
-            - |
-              rpk redpanda start \
-                --smp={{ include "redpanda-smp" . }} \
-                --memory={{ template "redpanda-memory" . }}M \
-                --reserve-memory={{ template "redpanda-reserve-memory" . }}M \
-                --default-log-level={{ .Values.logging.logLevel }} \
-                --advertise-rpc-addr={{ $internalAdvertiseAddress }}:{{ .Values.listeners.rpc.port }} |
-              sed 's@RP_BOOTSTRAP_USER[^ ]\+@<censored>@g'
+            - rpk
```
alejandroEsc, line 300: wow, i keep forgetting this is a thing. Glad we can move away from this. There are other places we are still using bash, we may want to consider reviewing those also.

alejandroEsc (approved):
lgtm

PR #440, opened 2023-04-14
feat: allowing secretRef for tls

### LINKED ISSUE #352: Allow enabling TLS and using certificates without cert-manager

Some users are not using cert-manager, but would still like to enable TLS and use certificates.

### Why is this needed?

This will allow the helm chart to be used for TLS-enabled deployments in environments where cert-manager is not being used.

### mutual TLS in Admin API

The readiness Probe doesn't work with the following values:
```yaml
listeners:
  admin:
    tls:
      requireClientAuth: true
      enabled: true
```

The Admin API needs client certs in order to respond to the request. The `rpk` cli support them via `--admin-api-tls-truststore`, `--admin-api-tls-cert` and `--admin-api-tls-key`.

alejandroEsc:
Allow for secretRefs in tls definitions for certs to work without using cert-manager. Must create proper CA and tls certs to work. 

Fixes #352

### DIFF
--- a/charts/redpanda/templates/certs.yaml
+++ b/charts/redpanda/templates/certs.yaml
@@ -22,7 +22,9 @@ limitations under the License.
   {{- $listeners := .Values.listeners -}}
   {{- $values := .Values }}
   {{- range $name, $data := .Values.tls.certs }}
+  {{- if (empty $data.secretRef ) }}
     {{- $d := $data.duration }}
+
 ---
 apiVersion: cert-manager.io/v1
 kind: Certificate
@@ -68,6 +70,7 @@ spec:
     name: {{ template "redpanda.fullname" $ }}-{{ $name }}-root-issuer
     kind: Issuer
     group: cert-manager.io
+    {{- end }}
     {{- end }}
   {{- end }}
 {{- end }}
--- a/charts/redpanda/templates/post-install-upgrade-job.yaml
+++ b/charts/redpanda/templates/post-install-upgrade-job.yaml
@@ -17,6 +17,8 @@ limitations under the License.
 {{- if .Values.post_install_job.enabled }}
 {{- $values := .Values }}
 {{- $sasl := $values.auth.sasl }}
+{{- $root := deepCopy . }}
+
 ---
 apiVersion: batch/v1
 kind: Job
@@ -136,8 +138,9 @@ spec:
             name: {{ template "redpanda.fullname" . }}
         - name: config
           emptyDir: {}
-      {{- if (include "tls-enabled" . | fromJson).bool }}
-        {{- range $name, $cert := .Values.tls.certs }}
+{{- if (include "tls-enabled" . | fromJson).bool }}
+  {{- range $name, $cert := .Values.tls.certs }}
+      {{- $r :=  set $root "tempCert" ( dict "name" $name "cert" $cert ) }}
         - name: redpanda-{{ $name }}-cert
           secret:
             defaultMode: 420
@@ -146,13 +149,13 @@ spec:
               path: tls.key
             - key: tls.crt
               path: tls.crt
-          {{- if $cert.caEnabled }}
+      {{- if $cert.caEnabled }}
             - key: ca.crt
               path: ca.crt
-          {{- end }}
-            secretName: {{ template "redpanda.fullname" $ }}-{{ $name }}-cert
-        {{- end }}
-      {{- end -}}
+      {{- end }}
+            secretName: {{ template "cert-secret-name" $r }}
+  {{- end }}
+{{- end -}}
       {{- if and $sasl.enabled (not (empty $sasl.secretRef )) }}
         - name: {{ $sasl.secretRef }}
           secret:
--- a/charts/redpanda/values.yaml
+++ b/charts/redpanda/values.yaml
@@ -139,9 +139,28 @@ tls:
       # issuerRef:
       #   name: redpanda-default-root-issuer
       #   kind: Issuer   # Can be Issuer or ClusterIssuer
+      # -- To use a secret with custom tls files,
+      # secretRef:
+      #  name: my-tls-secret
       # -- Set the `caEnabled` flag to `true` only for Certificates
       # that are not authenticated using public authorities.
       caEnabled: true
+      # duration: 43800h
+    # -- Example external tls configuration
+    # uncomment and set the right key to the listeners that require them
+    # also enable the tls setting for those listeners.
+    # external:
+      # -- To use a custom pre-installed Issuer,
+      # add its name and kind to the `issuerRef` object.
+      # issuerRef:
+      #   name: redpanda-default-root-issuer
+      #   kind: Issuer   # Can be Issuer or ClusterIssuer
+      # -- To use a secret with custom tls files,
+      # secretRef:
+      #   name: my-tls-secret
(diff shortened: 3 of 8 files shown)


charts/redpanda/values.yaml
```diff
@@ -149,7 +149,7 @@ tls:
     # -- Example external tls configuration
     # uncomment and set the right key to the listeners that require them
     # also enable the tls setting for those listeners.
-    #external:
+    # external:
```
RafalKorepta, line 152: NIT: squash with previous commit?

charts/redpanda/templates/certs.yaml
```diff
@@ -23,6 +23,8 @@ limitations under the License.
   {{- $values := .Values }}
   {{- range $name, $data := .Values.tls.certs }}
     {{- $d := $data.duration }}
+    {{- if (empty $data.secretRef ) }}
+
```
RafalKorepta, line 27:
```suggestion
    {{- if (empty $data.secretRef ) }}
    {{- $d := $data.duration }}

```

charts/redpanda/templates/post-install-upgrade-job.yaml
```diff
@@ -146,13 +149,13 @@ spec:
...
-          {{- if $cert.caEnabled }}
+      {{- if $cert.caEnabled }}
             - key: ca.crt
               path: ca.crt
-          {{- end }}
-            secretName: {{ template "redpanda.fullname" $ }}-{{ $name }}-cert
-        {{- end }}
-      {{- end -}}
+      {{- end }}
+            secretName: {{ template "cert-secret-name" $r }}
```
RafalKorepta, line 156: Why the $cert can not be send to the template like
```suggestion
            secretName: {{ template "cert-secret-name" $cert }}
```

The template would need to change slightly.

RafalKorepta (commented):
I'm not sure if existing test could be duplicated like `charts/redpanda/ci/03-one-node-cluster-tls-no-sasl-values.yaml` with k8s secret created from github action step before we start executing `ct`, wdyt?

charts/redpanda/values.yaml
```diff
@@ -139,10 +139,29 @@ tls:
...
+    # also enable the tls setting for those listeners.
+    # external:
+      # -- To use a custom pre-installed Issuer,
+      # add its name and kind to the `issuerRef` object.
+      # issuerRef:
+      #   name: redpanda-default-root-issuer
+      #   kind: Issuer   # Can be Issuer or ClusterIssuer
+      # -- To use a secret with custom tls files,
+      # secretRef:
+      #   name: my-tls-secret
```
RafalKorepta, line 160: Should we describe that we require `tls.key` and `tls.crt` keys within referenced secret? If `caEnabled` is set to `true` then `ca.crt` too.

charts/redpanda/templates/certs.yaml
joejulian: Another nit, for all `{{-` between here and `{{- end }}`, add an indent. It just makes figuring out the nesting a little bit easier when reading.

charts/redpanda/templates/post-install-upgrade-job.yaml
joejulian: The named template looks to see if this iteration's $cert has a secretRef. He adds a key to a copy of the root context (so as not to pollute the root) and passes the combination to "cert-secret-name". If secretRef is empty, that template includes a call to the "redpanda.fullname" template which needs the root context.

https://github.com/redpanda-data/helm-charts/pull/440/files#diff-8407f32f870c88088a3041f8a5a8394806973dabdef154b7fde83d5e77f70ed8R624-R630

charts/redpanda/values.yaml
```diff
@@ -640,25 +662,16 @@ listeners:
...
-      default:
-        # enabled: true
-        port: 8083
-        advertisedPorts:
-        - 30082
+        # -- Uncomment to define external tls
+        # tls: #
+        #   Optional flag to override the global TLS enabled flag.
+        #   enabled: true
+        #   cert: external
```
joejulian, line 669:
```suggestion
        # tls:
        #   # Optional flag to override the global TLS enabled flag.
        #   # enabled: true
        #   cert: external
```

This allows most editors to easily uncomment the block and still have the comment and optional parameter commented.

charts/redpanda/values.yaml
```diff
@@ -677,6 +691,32 @@ listeners:
         port: 8084
         advertisedPorts:
         - 30081
+        # -- Uncomment to define external tls
+        # tls: #
+        #   Optional flag to override the global TLS enabled flag.
+        #   enabled: true
+        #   cert: external
```
joejulian, line 698:
```suggestion
        # tls:
        #   # Optional flag to override the global TLS enabled flag.
        #   # enabled: true
        #   cert: external
```

charts/redpanda/values.yaml
```diff
@@ -677,6 +691,32 @@ listeners:
...
+      default:
+        # enabled: true
+        port: 8083
+        advertisedPorts:
+          - 30082
+        # -- Uncomment to define external tls
+        # tls: #
+        #   Optional flag to override the global TLS enabled flag.
+        #   enabled: true
+        #   cert: external
```
joejulian, line 719:
```suggestion
        # tls:
        # #  Optional flag to override the global TLS enabled flag.
        # #  enabled: true
        #   cert: external
```

charts/redpanda/values.yaml
alejandroEsc: I think we can do this in documentation.

RafalKorepta (approved):

carlosfwrk:
Hello Redpanda team,

I noticed that you recently updated your Helm chart to version 4.0.4, which allows for installing Redpanda without Certmanager and using a custom certificate signed by a company's own CA. However, the[ documentation on your website](https://docs.redpanda.com/docs/manage/kubernetes/security/kubernetes-tls/) has not been updated yet and still refers to configuring Redpanda with Certmanager.

I tried to install Redpanda with TLS enabled based on the values.yaml explanation, but I'm encountering an error that says` "unable to build kubernetes objects from release manifest: [unable to recognize "": no matches for kind "Certificate" in version "cert-manager.io/v1", unable to recognize "": no matches for kind "Issuer" in version "cert-manager.io/v1"]"`.
This is how I configure TLS:
```yaml
tls:
  enabled: true
  certs:
    default:
      secretRef:
        name: tls-redpanda
      caEnabled: true
```
This is how the secret tls-redpanda looks like:
```shell
Name:         tls-redpanda
Namespace:    redpanda

Type:  Opaque

Data
====
ca.crt:   2159 bytes
tls.crt:  2277 bytes
tls.key:  1704 bytes
```

Could you please update the documentation to reflect the changes made in version 4.0.4? Additionally, it would be great if you could provide more detailed instructions on how to configure Redpanda with a custom certificate signed by a company's own CA.

Thank you very much for your help!

Best regards,
Carlos.

alejandroEsc:
@carlosfwrk Thank you so much for your question. Yes this was a bug we noticed as we have been adding supporting test, and in part that is why documentation is not out for this feature. Please update to the latest version `helm repo update` (as we think this was fixed),  and let us know if this is still an issue. If you find you are still having trouble, would you mind opening an issue and the steps you take so we can triage? Thank you!

PR #443, opened 2023-04-17
redo redpanda container lifecycle probes

### LINKED ISSUE #421: awk for startupProbe and readinessProbe only look for controller_id in down nodes, not current broker id
It appears that when doing the startupProbe and readinessProbe for pod health, we aren't actually looking for the current pod's broker id. We're looking for the controller id in the down nodes list, which could trigger a cascading failure of actually healthy nodes. Additionally, we also may be missing cases where a broker _is_ in the list, but we don't fail out the probe. It doesn't appear that we have a good way to find the current broker id in the list based on the health_overview.

https://github.com/redpanda-data/helm-charts/blob/6e31833a0df989a76229a4f12522bbff3d107aff/charts/redpanda/templates/statefulset.yaml#L216-L247

See thread for more detail:

https://redpandadata.slack.com/archives/C01H6JRQX1S/p1680276099505199

From the thread:

Assume we have 3 brokers in a cluster. controller is on broker 0. If broker id 0 shows up in the down nodes list, all pods in the cluster will likely die off.

Additionally, if any other broker shows up in the down nodes list, the awk ends up with a false negative and doesn't seem to exit with a non-zero status to trigger a failure in the probes for that broker.

I instrumented the awk a bit to make this more obvious in testing it. Below we can see that broker0 being down in the first test would trigger the probe failure. But if broker1 was down, we don't trigger.

```
$ cat health1
{"is_healthy": true, "controller_id": 0, "all_nodes": [0 1 2], "nodes_down": [0], "leaderless_partitions": []}
$
$ bash awk health1
id start: {"is_healthy": true, "controller_id": 0, "all_nodes": [0 1 2], "nodes_down": [0], "leaderless_partitions": []}
id first gsub: 0, "all_nodes": [0 1 2], "nodes_down": [0], "leaderless_partitions": []}
id second gsub: 0
nd_str start: {"is_healthy": true, "controller_id": 0, "all_nodes": [0 1 2], "nodes_down": [0], "leaderless_partitions": []}
nd_str first gsub: 0], "leaderless_partitions": []}
nd_str second gsub: 0
nodes_down: nd_list 1 items - nd_list[i] 0 nodes_down[nd_list[i]]  end
(exit value) id in nodes_down:  1
$
$ cat health3
{"is_healthy": true, "controller_id": 0, "all_nodes": [0 1 2], "nodes_down": [1], "leaderless_partitions": []}
$
$ bash awk health3
id start: {"is_healthy": true, "controller_id": 0, "all_nodes": [0 1 2], "nodes_down": [1], "leaderless_partitions": []}
id first gsub: 0, "all_nodes": [0 1 2], "nodes_down": [1], "leaderless_partitions": []}
id second gsub: 0
nd_str start: {"is_healthy": true, "controller_id": 0, "all_nodes": [0 1 2], "nodes_down": [1], "leaderless_partitions": []}
nd_str first gsub: 1], "leaderless_partitions": []}
nd_str second gsub: 1
nodes_down: nd_list 1 items - nd_list[i] 1 nodes_down[nd_list[i]]  end
(exit value) id in nodes_down:  0
```

joejulian:
Fixes #421

### DIFF
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 4.0.0
+version: 4.0.1
 
 # The app version is the default version of Redpanda to install.
 appVersion: v23.1.7
@@ -50,6 +50,6 @@ annotations:
       url: https://helm.sh/docs/intro/install/
   artifacthub.io/images: |
     - name: redpanda
-      image: vectorized/redpanda:v23.1.1
+      image: vectorized/redpanda:v23.1.7
     - name: busybox
       image: busybox:latest
--- a/charts/redpanda/templates/statefulset.yaml
+++ b/charts/redpanda/templates/statefulset.yaml
@@ -215,83 +215,44 @@ spec:
             preStop:
               exec:
                 command: ["/bin/bash", "-c", "./var/lifecycle/preStop.sh"]
+          # the startupProbe checks to see that the admin api is listening and that the broker has a node_id assigned. This
+          # check is only used to delay the start of the liveness and readiness probes until it passes.
           startupProbe:
             exec:
               command:
                 - /bin/sh
                 - -c
-                {{- if (include "admin-internal-tls-enabled" . |fromJson).bool }}
-                - >
-                  curl https://localhost:{{ .Values.listeners.admin.port }}/v1/cluster/health_overview
-                  -svk --cacert /etc/tls/certs/{{ .Values.listeners.admin.tls.cert }}/ca.crt |
-                  awk '{
-                    id = $0; gsub(/.*"controller_id": /, "", id); gsub(/,.*/, "", id)
-                    nd_str = $0; gsub(/.*"nodes_down": \[/, "", nd_str); gsub(/\].*/, "", nd_str)
-                    FS=","
-                    split(nd_str, nd_list)
-                    for (i in nd_list) nodes_down[nd_list[i]]=""
-                    exit (id in nodes_down)
-                  }'
-                {{- else }}
-                - >
-                  curl -sv http://localhost:{{ .Values.listeners.admin.port }}/v1/cluster/health_overview |
-                  awk '{
-                    id = $0; gsub(/.*"controller_id": /, "", id); gsub(/,.*/, "", id)
-                    nd_str = $0; gsub(/.*"nodes_down": \[/, "", nd_str); gsub(/\].*/, "", nd_str)
-                    FS=","
-                    split(nd_str, nd_list)
-                    for (i in nd_list) nodes_down[nd_list[i]]=""
-                    exit (id in nodes_down)
-                  }'
-                {{- end }}
+                - |
+                  set -e
+                  RESULT=$(curl --silent --fail -k {{ include "admin-tls-curl-flags" . }} "{{ include "admin-http-protocol" . }}://localhost:{{ .Values.listeners.admin.port }}/v1/status/ready")
+                  echo $RESULT
+                  echo $RESULT | grep ready
             initialDelaySeconds: {{ .Values.statefulset.startupProbe.initialDelaySeconds }}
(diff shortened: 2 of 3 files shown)


charts/redpanda/Chart.yaml
```diff
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 4.0.0
+version: 4.0.1
```
mmaslankaprv, line 26: nit: this could be a separate commit

RafalKorepta (approved):

charts/redpanda/templates/statefulset.yaml
```diff
@@ -215,83 +215,38 @@ spec:
...
                 - -c
-                {{- if (include "admin-internal-tls-enabled" . |fromJson).bool }}
-                - >
-                  curl https://localhost:{{ .Values.listeners.admin.port }}/v1/cluster/health_overview
-                  -svk --cacert /etc/tls/certs/{{ .Values.listeners.admin.tls.cert }}/ca.crt
-                {{- else }}
-                - >
-                  curl -sv http://localhost:{{ .Values.listeners.admin.port }}/v1/cluster/health_overview
-                {{- end }}
+                - curl --silent --fail -k {{ include "admin-tls-curl-flags" . }} "{{ include "admin-http-protocol" . }}://localhost:{{ .Values.listeners.admin.port }}/v1/node_config"
```
jcsp, line 234: If you use `/v1/ready`, that's a bit cheaper for the server than serializing the node config each time.

You don't necessarily have to read the body of the response ("ready" or "booting"), if you just want a simple endpoint to check if the admin API is up.

charts/redpanda/templates/statefulset.yaml
```diff
@@ -215,83 +215,38 @@ spec:
...
-                  awk '{
-                    id = $0; gsub(/.*"controller_id": /, "", id); gsub(/,.*/, "", id)
-                    nd_str = $0; gsub(/.*"nodes_down": \[/, "", nd_str); gsub(/\].*/, "", nd_str)
-                    FS=","
-                    split(nd_str, nd_list)
-                    for (i in nd_list) nodes_down[nd_list[i]]=""
-                    exit (id in nodes_down)
-                  }'
-                {{- end }}
+                - curl --silent --fail -k {{ include "admin-tls-curl-flags" . }} "{{ include "admin-http-protocol" . }}://localhost:{{ .Values.listeners.admin.port }}/v1/node_config" | grep -o '"node_id":[0-9]\+'
```
jcsp, line 224: Perhaps this check should be followed by a call to `/v1/ready` to check the status is `ready`.

As-written, this startupProbe will complete slightly before the node is actually ready to serve Kafka requests.

The startup order is:
- Bootstrap (including getting a node ID if necessary)
- Start most other services (including the admin server, which will return "booting" from /v1/ready at this point)
- Start listening on the kafka port
- Finally, set the status to "ready" in the result of /v1/ready.
- Bootstrap (this includes getting a node ID if it doesn't already have one)

alejandroEsc: @jcsp would it be enough just to wait for the /v1/ready to both respond and give an http 200 response?

joejulian: Looks like a "no". It gives 200 for "booting" which is not "ready".

charts/redpanda/templates/statefulset.yaml
joejulian: Thanks!

charts/redpanda/Chart.yaml
joejulian: We like to keep them together. It's handy for `git log -S`.

charts/redpanda/templates/statefulset.yaml
joejulian: I just switched to the ready link. That seems to be sufficient for the startupProbe

joejulian: had to use `/v1/status/ready` as `/v1/ready` was giving 404s.

alejandroEsc (approved):
this is an awesome improvement.

PR #446, opened 2023-04-19
Remove check for unsupported Redpanda version

RafalKorepta:
22.1.1 Redpanda version is now unsupported, so all occurrences was removed.


alejandroEsc (approved):
lgtm, nice to remove these

alejandroEsc:
One question, Do we have a stop gap that if a user uses this image with the chart that we should "not deploy?"

RafalKorepta:
@mattschumpert What you think about explicitly not allow deployments for unsupported versions of Redpanda?

RafalKorepta:
I will merge as is and then address any follow ups.

mattschumpert:
Good question. Within a version of Helm/operator designed to be paired with certain versions of Redpanda, I would be happy to disallow deploying the 'wrong' redpanda version for that helm/operator version.  The question is what's the correct allowed range of versions.  We previously proposed locking the Operator/Helm chart to a particular Redpanda version and even auto-upgrading the cluster after Operator upgrade (so there is no chance to mix). That might be too extreme for Helm, but would really love to enforce conjoined upgrades (operator+helm chart is part of version X of the platform) to avoid supporting mixed versions as much as possible.  Having a pure 1:1 is by far cheapest thing to support long-term and I haven't seen a customer need to mix versions. This becomes even more messy once Operator+Helm is the stuff you deploy working together.  I am all for limiting it and forcing people to upgrade chart to a matched version, unless I missed something and that doesn't work for some reason

Issue #405, opened 2023-03-25
After upgrade, access_point_uri is empty


travisdowns:
### What happened?

After an upgrade of my helm chart to 3.0.6 and Redpanda from 23.1.1 to 23.1.3, S3 cloud storage was not working.

Underlying cause is that the redpanda config `cloud_storage_api_endpoint` is an empty string rather than null. Redpanda treats this as a "set" URL and then uses the empty string instead of the default `s3.amazonaws...` URI and will fail to connect.

I initially deployed this cluster with helm and had done several helm upgrades without this issue. In the most recent upgrade I updated the helm chart to 3.0.6 and also the Redpanda version to v23.1.3 (from .1) when this issue occurred.

```
redpanda        redpanda        13              2023-03-25 03:28:03.054703601 +0000 UTC deployed        redpanda-3.0.6          v23.1.1
```

### What did you expect to happen?

Tiered storage to keep working.

### How can we reproduce it (as minimally and precisely as possible)?. Please include values file.

Perhaps a helm upgrade with the newest helm chart will do it.

```console
$ helm get values <redpanda-release-name> -n <redpanda-release-namespace> --all
COMPUTED VALUES:
USER-SUPPLIED VALUES: null
auth:
  sasl:
    enabled: false
    mechanism: SCRAM-SHA-512
    secretRef: redpanda-users
    users:
    - mechanism: SCRAM-SHA-512
      name: admin
      password: change-me
clusterDomain: cluster.local
commonLabels: {}
config:
  cluster:
    retention_bytes: 461718750000
    retention_local_target_bytes_default: 39410000000
  node: {}
  tunable:
    compacted_log_segment_size: 67108864
    group_topic_partitions: 16
    kafka_batch_max_bytes: 1048576
    kafka_connection_rate_limit: 1000
    log_segment_size: 536900000
    log_segment_size_max: 536900000
    log_segment_size_min: 4096
    max_compacted_log_segment_size: 536870912
    topic_partitions_per_shard: 1000
console:
  affinity: {}
  autoscaling:
    enabled: false
    maxReplicas: 100
    minReplicas: 1
    targetCPUUtilizationPercentage: 80
  config: {}
  configmap:
    create: false
  console:
    config: {}
  deployment:
    create: false
  enabled: true
  extraContainers: {}
  extraEnv: []
  extraEnvFrom: []
  extraVolumeMounts: []
  extraVolumes: []
  fullnameOverride: ""
  global: {}
  image:
[...]

PR #449, opened 2023-04-19
Omit empty value in Redpanda configuration

RafalKorepta:
From go template documentation:
```
The empty values are false, 0, any nil pointer or interface value,
and any array, slice, map, or string of length zero.
```

Fixes https://github.com/redpanda-data/helm-charts/issues/405


joejulian (approved):
perfect!

PR #451, opened 2023-04-20
feat: make login-jwt-secret configurable

stefan-caraiman:
Required in GitOps tools such as ArgoCD https://argo-cd.readthedocs.io/en/stable/user-guide/helm/#random-data

The random value is regenerated every time the comparison is made, any application which makes use of the randAlphaNum function will always be in an OutOfSync state.

### DIFF
--- a/charts/console/Chart.yaml
+++ b/charts/console/Chart.yaml
@@ -27,7 +27,7 @@ type: application
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
 
 # Chart versions do not track appVersion
-version: 0.6.1
+version: 0.6.2
 
 # The app version is the version of the Chart application
 appVersion: v2.2.3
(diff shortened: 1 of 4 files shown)


RafalKorepta (changes requested):
Forgot to ask. Can you bump the chart version https://github.com/redpanda-data/helm-charts/blob/98b10ae0cced0c9a8be4f56750b94dc6ec88e9a1/charts/console/Chart.yaml#L30 ?

stefan-caraiman:
> Forgot to ask. Can you bump the chart version
> 
> https://github.com/redpanda-data/helm-charts/blob/98b10ae0cced0c9a8be4f56750b94dc6ec88e9a1/charts/console/Chart.yaml#L30
> 
> ?

done ✅

RafalKorepta (approved):

RafalKorepta (approved):

joejulian (approved):
We've talked about this internally a bit. Having a plain-text secret in the values.yaml isn't ideal, but the helm release will reside in a Secret and the template places the data in a Secret so it's not completely horrible. Plus, the rest of this chart has this same problem so I'll give my approval as well.

Sorry this required some patience.

joejulian:
rebased and bumped chart version

PR #453, opened 2023-04-20
allow ingress without host

acristu:
There are cases where explicitly setting the host is not desired, multiplexing to different services can be done via `path`.
The console can be deployed with a `basePath` and ingress without an explicit host,


alejandroEsc (commented):
Can you please update the chart.yaml file to the next patch version. Thank you!

joejulian:
Also, please rebase. I think that'll fix CI.

charts/console/templates/ingress.yaml
```diff
@@ -38,7 +38,9 @@ spec:
   {{- end }}
   rules:
     {{- range .Values.ingress.hosts }}
-    - host: {{ tpl (.host) $ | quote }}
+    - {{- if .host }}
+      host: {{ .host }}
+      {{- end}}
```
joejulian, line 43: This was previously allowing .host to be a template. This changes removes this feature.
```suggestion
    {{- $host := tpl .host $ }}
    - {{- if $host }}
      host: {{ $host }}
      {{- end}}
```

acristu: thanks for the review, I've re-added the `tpl` , I can add the `$host` also if you want, but for now I just put the previous code back

RafalKorepta (approved):

joejulian:
If you'll check that allow changes box, we can bump the chart version for you as needed to get this merged. For now, it needs rebased and the chart version needs bumped again.

charts/console/templates/ingress.yaml
joejulian: I was thinking that if `tpl .host $` produces an empty string, you may still want it to be skipped. That's why I suggested the `$host` idea.

joejulian:
closing as abandoned. Work moved to #537

PR #455, opened 2023-04-21
Testing external listeners with secret certs

### LINKED ISSUE #448: redpanda-chart: improving tls cert testing

We wish to expand tls cert testing to cover the following:

- Creating custom certs  for both internal and external listeners
- Creating custom certs for external listeners but use cert-manager for internal users
- Creating custom certs for internal listeners but use cert-manager for external users
- Creating cert-manager certs for internal and external listeners

We will need to create custom certs and apply them to a cluster, they should be available before deploying redpandas. 
We also need a way to test via nodeport to ensure that the right certs are being applied. 

### Why is this needed?

We need to have confidence that we are handling certificate creation properly and that the listeners are behaving as intended.

alejandroEsc:
Fixes #448 

Adds ability to test using certificates on listeners.

A CI entry that adds (and can be expanded to) external ~and internal tls~ backed by secrets. These are referenced in the modified listeners (changed in this custom values file). 

The secrets are created in a PR where a script is called, if there is a domain passed it is assumed that it is for an external listener. The script also adds annotations to the secrets so that they will be created before the launch of other resources in a helm hook. Finally, the PR code will move these into the template directory, including a service yaml file (discussed below).

The service file that is also added to the templates directory was necessary to test the listeners from the inside of the a pod. The service targets the listener endpoints and is named in the same way as the domain defined in the CI file. The naming is to take advantage of the kubernetes internal dns with respect to services, since internally we can simply call `<service-name>:<port>` which is then routed to the targetPort of the pods defined in the proper selectors. That helps in ensuring the certificates have the correct alt-names defined.

The new test added, allows us to use opessl `s_client` to verify that each listener is using the correct certificate (internally and externally) and that they have the correct certificate and the the right key works as intended.

Finally an outcome of this testing has been multiple fixes, one additional issue was discovered where we need to also have the tests pick up certificate files correctly if they are from secrets. We include those fixes here. 

This is a first step change. We should include additional tests soon but for now this covers a wide range of possible certificate cases. Though we are not really testing externally, the use of an internal service that points to the same external endpoints should work as a close analog and so we expect to capture any issues in the future if they arise.

### DIFF
--- a/.github/workflows/pull_requests.yaml
+++ b/.github/workflows/pull_requests.yaml
@@ -125,6 +125,17 @@ jobs:
         run: .github/annotate_kind_nodes.sh chart-testing
         if: steps.list-changed.outputs.changed == 'true'
 
+      - name: Create tls helm templates
+        run: |
+          .github/create_tls.sh "random-domain"
+        if: steps.list-changed.outputs.changed == 'true'
+
+      - name: Move tls files to template
+        run: |
+          mv external-tls-secret.yaml charts/redpanda/templates/
+          cp .github/external-service.yaml charts/redpanda/templates/
+        if: steps.list-changed.outputs.changed == 'true'
+
       - name: install cert-manager
         run: |
           helm repo add jetstack https://charts.jetstack.io &&
--- a/.github/create_tls.sh
+++ b/.github/create_tls.sh
@@ -1,0 +1,100 @@
+#!/usr/bin/env bash
+set -xeuo pipefail
+
+DOMAIN=${1-}
+
+rm tls.crt tls.csr tls.key \
+  ca.crt ca.key cert.conf csr.conf \
+  ca.srl || true
+
+# if domain was clean, we wanted to just clean everything.
+if [ "${DOMAIN}" = "clean" ]; then
+  rm external-tls-secret.yaml internal-tls-secret.yaml || true
+  exit 0
+fi
+
+# check here if we have openssl, exit immediately if not
+if ! command -v openssl &> /dev/null
+then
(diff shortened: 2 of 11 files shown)


RafalKorepta (commented):
Everything looks good. Now we need to use this script right?

.github/workflows/pull_requests.yaml
```diff
@@ -120,6 +120,17 @@ jobs:
...
+      - name: Create tls helm templates
+        run: |
+          .github/create_tls.sh
+          .github/create_tls.sh "random-domain"
+
+      - name: Move tls files to template
+        run: |
+          mv external-tls-secret.yaml ./charts/redpanda/templates/
+          mv internal-tls-secret.yaml ./charts/redpanda/templates/
+          cp .github/external-service.yaml ./charts/redpanda/templates/
```
RafalKorepta, line 132:
```suggestion
      - name: Create tls helm templates
        run: |
          .github/create_tls.sh
          .github/create_tls.sh "random-domain"
        if: steps.list-changed.outputs.changed == 'true'

      - name: Move tls files to template
        run: |
          mv external-tls-secret.yaml ./charts/redpanda/templates/
          mv internal-tls-secret.yaml ./charts/redpanda/templates/
          cp .github/external-service.yaml ./charts/redpanda/templates/
        if: steps.list-changed.outputs.changed == 'true'
```

alejandroEsc: Sounds good, out of curios what does this guarantee? I have tried searching but found little.

RafalKorepta (approved):

joejulian (approved):

PR #457, opened 2023-04-21
Create tests for cloud storage configurations

### LINKED ISSUE #396: Create testing process  for cloud storage settings [eks]
We have no tests for the cloud storage settings. The azure keys were added, defaulting to empty strings. This broke gcp (and probably aws) storage because the azure settings too precedence. Redpanda treats empty strings as being set so it broke the whole cluster.

We need tests in each cloud to protect against regressions.

joejulian:
This creates a process for testing cloud tiered storage settings for EKS.

Fixes #396


alejandroEsc (approved):
Not a blocker, but can we move the chart in to the charts repo? or is this something down the line?

PR #461, opened 2023-04-25
Change the default termination grace period

### LINKED ISSUE #376: Default terminationGracePeriodSeconds to at least 60 seconds
### What happened?

nothing yet!

We need to ensure enough time is given for maintenance mode to transfer leadership and for redpanda to shut down. According to core:

> It's a little hard to estimate exactly how long it should take: leadership movement is very fast if there is already a follower that is up to date + ready to take over, but I don't think we have good testing of exactly how long that takes when you've got a fully loaded system with 10ks of partitions.
>
>Very roughly, it might make sense to have something like a 30 second timeout for leadership draining (that's very generous, it will usually be much faster), and then another 30 second timeout for the node to shut down once it has drained (again, very generous compared with typical time).

So 60 seconds is considered generous. Let's make it a configurable default 120 seconds and edit the preStop and postStart hooks to timeout (no-fail) after half that time. This will allow redpanda to still gracefully shut down even if something else is wrong - and allow us to read the logs.

_No response_

Redpanda

RafalKorepta:
Fixes #376

### DIFF
--- a/charts/redpanda/templates/secrets.yaml
+++ b/charts/redpanda/templates/secrets.yaml
@@ -48,19 +48,25 @@ stringData:
     source /var/lifecycle/common.sh
 
 {{- if gt ( .Values.statefulset.replicas | int64 ) 2 }}
-    set -ex
+    postStartHook () {
+      set -x
 
-    until NODE_ID=$(${CURL_NODE_ID_CMD} | grep -o '\"node_id\":[^,}]*' | grep -o '[^: ]*$'); do
-        sleep 0.5
-    done
+      until NODE_ID=$(${CURL_NODE_ID_CMD} | grep -o '\"node_id\":[^,}]*' | grep -o '[^: ]*$'); do
+          sleep 0.5
+      done
 
-    echo "Clearing maintenance mode on node ${NODE_ID}"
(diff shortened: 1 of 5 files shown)


charts/redpanda/templates/secrets.yaml
```diff
@@ -48,19 +48,23 @@ stringData:
...
+      echo "Clearing maintenance mode on node ${NODE_ID}"
+      CURL_MAINTENANCE_DELETE_CMD="${CURL_MAINTENANCE_DELETE_CMD_PREFIX} {{ include "admin-tls-curl-flags" . }} ${CURL_URL}/v1/brokers/${NODE_ID}/maintenance"
+      # a 400 here would mean not in maintenance mode
+      until [ "${status:-}" = '"200"' ] || [ "${status:-}" = '"400"' ]; do
+          status=$(${CURL_MAINTENANCE_DELETE_CMD})
+          sleep 0.5
+      done
+    }
+
+    timeout {{ .Values.statefulset.terminationGracePeriodSeconds }} postStartHook
```
joejulian, line 67:
```suggestion
    timeout {{ div .Values.statefulset.terminationGracePeriodSeconds 2 }} postStartHook | true
```
Let's give Redpanda some time after this is terminated. Also never fail or the pod will terminate immediately.

alejandroEsc (approved):
lgtm!

joejulian:
rebased

joejulian (approved):

PR #463, opened 2023-04-26
Expose console startup times

### LINKED ISSUE #411: Allow console startup times to be configured

Allow console deployments to configure the startup maxRetries, retryInterval, maxRetryInterval, and backoffMultiplier through the chart values.

### Why is this needed?

When the console pods are starting, if kafka isn't listening yet the pod will die and get restarted. This is normal expected behavior but this can alerts for failing pods in some monitoring configurations. Users would like to have this pod not fail during installation.

RafalKorepta:
Expose kafka startup configuration options.

Fixes #411


alejandroEsc (approved):
lgtm

alejandroEsc:
The testing is failing because the console chart fails to upgrade.

alejandroEsc (approved):

joejulian (approved):

PR #464, opened 2023-04-26
PR action has wrong val to test for matrix

alejandroEsc:
PR fixes our actions PR tests. I have been wondering why the right values were not run, here is why. This should fix our matrix.


joejulian (approved):

alejandroEsc:
Merging, this should not affect any of the charts directly.

PR #469, opened 2023-04-27
make external tls overrides easier to use

joejulian:
This will simplify the use of external tls certificates in the common case, changing from:
```
tls:
  certs:
    external:
      secretRef:
        name: my-external-cert
      caEnabled: false
listeners:
  admin:
    external:
      default:
        tls:
          cert: external
  kafka:
    external:
      default:
        tls:
          cert: external
  schemaRegistry:
    external:
      default:
        tls:
          cert: external
  http:
    external:
      default:
        tls:
          cert: external
```      
to
```
tls:
  certs:
    external:
      secretRef:
        name: my-external-cert
      caEnabled: false
```

### DIFF
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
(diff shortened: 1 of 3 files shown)


alejandroEsc (approved):
Lgtm

PR #472, opened 2023-04-28
Pass boolean properties to Redpanda configuration

### LINKED ISSUE #470: Cannot set boolean cluster config settings false
### What happened?

`{{- if $value }}` in https://github.com/redpanda-data/helm-charts/blob/main/charts/redpanda/templates/post-upgrade.yaml#L68
`{{- if $element }}` in https://github.com/redpanda-data/helm-charts/blob/main/charts/redpanda/templates/configmap.yaml#L74 and https://github.com/redpanda-data/helm-charts/blob/main/charts/redpanda/templates/configmap.yaml#L118

Because if I am reading these correctly # enable_idempotence: false                                    # Enable idempotent producer in https://github.com/redpanda-data/helm-charts/blob/main/charts/redpanda/values.yaml#L744
would never be set and the default value for enable_idempotence is true https://github.com/redpanda-data/redpanda/blob/dev/src/v/config/configuration.cc#L560-L565

Don't allow setting any values to boolean `false`. 

Allow key/values where the value is `typeOf` boolean.

.

_No response_

Redpanda

### Chart Version(s)

.

### Cloud provider

.

RafalKorepta:
fixes #470

### DIFF
--- a/charts/redpanda/templates/configmap.yaml
+++ b/charts/redpanda/templates/configmap.yaml
@@ -244,10 +244,14 @@ data:
             port: {{ $values.listeners.rpc.port }}
 {{- end }}
 {{- if and (include "is-licensed" . | fromJson).bool .Values.storage.tieredConfig.cloud_storage_enabled }}
-  {{- if (include "redpanda-atleast-22-3-0" . | fromJson).bool }}
-    {{- toYaml .Values.storage.tieredConfig | nindent 6 }}
-  {{- else }}
-    {{- unset .Values.storage.tieredConfig "cloud_storage_credentials_source" | toYaml | nindent 6 }}
+  {{- $tieredStorageConfig := deepCopy .Values.storage.tieredConfig }}
+  {{- if not (include "redpanda-atleast-22-3-0" . | fromJson).bool }}
+    {{- $tieredStorageConfig = unset $tieredStorageConfig "cloud_storage_credentials_source"}}
+  {{- end }}
+  {{- range $key, $element := $tieredStorageConfig}}
+    {{- if or (eq (typeOf $element) "bool") $element }}
+    {{ $key }}: {{ $element | toYaml }}
+    {{- end }}
   {{- end }}
 {{- end }}
 {{- /* Schema Registry API */}}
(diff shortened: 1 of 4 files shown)


RafalKorepta:
Screenshot 2023-04-28 at 12 15 08

https://github.com/redpanda-data/helm-charts/actions/runs/4829528122/jobs/8604675668

Dedicated test that should be run only in origin not in fork is correctly detected by Github Action and it is cancelled.

charts/redpanda/templates/configmap.yaml
```diff
@@ -53,6 +53,37 @@ limitations under the License.
...
+"enable_transactions"
+"enable_usage"
+"features_auto_enable"
+"kafka_enable_authorization"
+"kafka_enable_partition_reassignment"
+"kafka_qdc_enable"
+"legacy_group_offset_retention_enabled"
+"memory_abort_on_alloc_failure"
+"release_cache_on_segment_roll"
+"storage_strict_data_init" -}}
```
RafalKorepta, line 86: The boolean properties was taken from running cluster v23.1.7 by the following command:

```
kubectl exec redpanda-0 -- curl -sk https://127.0.0.1:9644/v1/cluster_config/schema | jq -c  '.properties | to_entries | .[] | select(.value.type | contains("boolean")) | .key'
```

RafalKorepta: Updated as 23.1.8 added new boolean `use_fetch_scheduler_group`

charts/redpanda/templates/configmap.yaml
```diff
@@ -71,7 +107,10 @@ data:
   {{- end }}
   {{- with (dig "cluster" dict .Values.config) }}
       {{- range $key, $element := .}}
-        {{- if $element }}
```
alejandroEsc, line 74: We could do something like https://helm.sh/docs/chart_template_guide/function_list/#type-functions
how about `{{ if or (eq (typeOf $element) boolean) $element }}` that way we do not have to know an exhaustive list of boolean configs. We leave the responsibility of the user to know what they are doing and we offload maintenance.

RafalKorepta: Sure I can change that.

joejulian:
rebased

joejulian (approved):

PR #474, opened 2023-04-28
tests require linter to pass in order to run

alejandroEsc:
Simple check so that tests are not run if linter fails.


alejandroEsc:
verified in https://github.com/redpanda-data/helm-charts/pull/475

RafalKorepta (approved):

PR #475, opened 2023-04-28
testing that linter fails cancels tests

alejandroEsc:


alejandroEsc:
Chart tests were skipped because we failed linter. Proving the change works. Closing ticket.

PR #476, opened 2023-04-28
Allow boolean false Redpanda configuration properties

### LINKED ISSUE #470: Cannot set boolean cluster config settings false

RafalKorepta:
fix #470


alejandroEsc (approved):
please merge as soon as we are green. Thank you!

PR #479, opened 2023-04-28
chore: lint filename in ci to ensure they are run

alejandroEsc:
We have had cases where CI wont run, unnoticed test cases because of file names. Here we avoid this pitfall by failing fast.


alejandroEsc:
This does not change any chart. So I have stopped the CI process, but let the linter do its job.

RafalKorepta (approved):

PR #480, opened 2023-04-29
23.1.8 version bump

alejandroEsc:


charts/redpanda/Chart.yaml
```diff
@@ -23,10 +23,10 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 4.0.12
+version: 5.0.0
```
joejulian, line 26: We'll only do the major for the "minor" Redpanda version (23.2 next).
```suggestion
version: 4.1.0
```

joejulian (approved):

PR #492, opened 2023-05-05
fix the nightly CI

### LINKED ISSUE #483: Nightly test failure 22.2 one-node-cluster-no-tls-sasl fails to install
### What happened?

https://github.com/redpanda-data/helm-charts/actions/runs/4826056502/jobs/8597529597#step:10:5840
https://github.com/redpanda-data/helm-charts/actions/runs/4847503478/jobs/8637813041#step:10:5517

Tests are failing to install with Redpanda 22.2

test to pass

<details>

```console
$ helm get values <redpanda-release-name> -n <redpanda-release-namespace> --all
# paste output here
```

</details>

_No response_

Redpanda

### Chart Version(s)

<details>

```console
7f7f645
4cf51ec
```

</details>

### Cloud provider

<details>

</details>

joejulian:
Add the missing certificate generation to the nightly CI to allow it to pass the external cert test.

probably fixes #483

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -41,6 +41,7 @@ jobs:
         run: ct lint --all --config .github/ct.yaml
   test:
     name: Run ct tests for ci values matching ${{ matrix.testvaluespattern }} for Redpanda version ${{ matrix.version }}
+    needs: lint
     strategy:
       matrix:
         version:
@@ -60,11 +61,6 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Set up for matrix
-        run: |
-          echo bash -O extglob -c "rm -v charts/redpanda/ci/!(${{ matrix.testvaluespattern }})"
-          bash -O extglob -c "rm -v charts/redpanda/ci/!(${{ matrix.testvaluespattern }})"
-
       - name: Set up Helm
         uses: azure/setup-helm@v3
         with:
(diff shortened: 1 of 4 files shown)


.github/workflows/pull_requests.yaml
```diff
@@ -66,7 +66,6 @@ jobs:
       - name: compare console values with main
         run: ./dyff --color=off -k between -s <(git show 'origin/main:charts/console/values.yaml') charts/console/values.yaml
   test:
-    needs: lint
```
alejandroEsc, line 69: why are we removing this line? We should avoid proceeding if linter fails.

.github/workflows/nightly.yaml
```diff
@@ -75,6 +70,11 @@ jobs:
         with:
           version: v3.7.1-2
 
+      - name: Set up for matrix
```
alejandroEsc, line 73: Are we checking if changes occurred, or is nightly just checking current main?

.github/workflows/pull_requests.yaml
joejulian: Because they run in parallel. If the linter fails, everything else will get killed anyway. Why make us wait?

.github/workflows/nightly.yaml
joejulian: Nightly is checking the current main.

.github/workflows/pull_requests.yaml
alejandroEsc: Linter and check-values run in parallel, not the matrix tests? from past CI runs we had linter failed but the other tests continue. Did you change some setting to fail if any tests fail? Maybe i missed this.

joejulian: Damn, you're right. Jobs can't cancel each other. There are tools to do that, but I'll just put it back for now.

PR #495, opened 2023-05-08
feat: create mock kafka to allow console to start

alejandroEsc:
Using a mock-kafka server to test out console and allow it to deploy during testing. The goal is to use a redpanda broker.

TODO
- ~Create mock scaffolding~
- ~replace with redpanda broker~

I have updated the pull_request workflow and included one distinct flow for just testing console. This is open to debate but I think there is legitimate reason to want to split these workflows out.


joejulian:
rebased

PR #496, opened 2023-05-10
sasl user creation rework and allow updates

### LINKED ISSUE #459: Allow changing superuser SASL user password

We are able to create a SASL user and set their password within `values.yaml`, but we cannot update this password in `values.yaml` for it then update the user's password.

### Why is this needed?

This behavior is expected since the password is listed in `values.yaml`

alejandroEsc:
Fixes #459 

This PR is attempting to create and update sasl users by means of inotify watchers, which will be monitoring user secret for sals users when enabled, update these users if they exist already, and maintain the usual creation process. 

Changes to make:
- [x] Move sasl creation code to a secret with execution happening in a side-car
- [x] sasl creation code changed to perform updates
- [x] clean code and debug tools
- [x] create testing

### DIFF
--- a/.github/ct.yaml
+++ b/.github/ct.yaml
@@ -15,7 +15,7 @@
 debug: true
 remote: origin
 target-branch: main
-helm-extra-args: --timeout 600s
+helm-extra-args: --timeout 900s
 chart-repos:
   - redpanda=https://charts.redpanda.com
 
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 4.0.17
+version: 4.0.18
 
 # The app version is the default version of Redpanda to install.
 appVersion: v23.1.8
--- a/charts/redpanda/templates/secrets.yaml
+++ b/charts/redpanda/templates/secrets.yaml
@@ -249,6 +249,6 @@ stringData:
         process_users $USERS_DIR
       done
     {{- end }}
-    sleep 60
+      sleep infinity
     done
 {{- end }}
--- a/charts/redpanda/templates/secrets.yaml
+++ b/charts/redpanda/templates/secrets.yaml
@@ -136,3 +136,83 @@ stringData:
   {{- end }}
     # intentional empty line
 {{- end }}
+
+{{- if .Values.statefulset.sideCars.configWatcher.enabled }}
+  {{- $values := .Values }}
+  {{- $sasl := .Values.auth.sasl }}
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ template "redpanda.fullname" . }}-config-watcher
+  namespace: {{ .Release.Namespace | quote }}
+  labels:
+  {{- with include "full.labels" . }}
+  {{- . | nindent 4 }}
+  {{- end }}
+type: Opaque
+stringData:
+  sasl-user.sh: |-
+    #!/usr/bin/env bash
+    set -e
+    
+    while true; do
+    {{- if and $sasl.enabled (not (empty $sasl.secretRef )) }}
+      echo "RUNNING: Monitoring and Updating SASL users"
+      USERS_DIR="/etc/secrets/users"
+      
+      process_users() {
+        USERS_DIR=${1-"/etc/secrets/users"}
+        USERS_FILE=$(find ${USERS_DIR}/* -print)
+        while read p; do
+          IFS=":" read -r USER_NAME PASSWORD MECHANISM <<< $p
+          # Do not process empty lines
+          if [ -z "$USER_NAME" ]; then
+            continue
+          fi
+          echo "Creating user ${USER_NAME}..."
+          MECHANISM=${MECHANISM:-{{- include "sasl-mechanism" . }}}
+          creation_result=$(rpk acl user create ${USER_NAME} --password=${PASSWORD} --mechanism ${MECHANISM} {{ template "rpk-flags-no-sasl" $ }} 2>&1) && creation_result_exit_code=$? || creation_result_exit_code=$?  # On a non-success exit code
+          if [[ $creation_result_exit_code -ne 0 ]]; then
+            # Check if the stderr contains "User already exists"
+            # this error occurs when password has changed
+            if [[ $creation_result == *"User already exists"* ]]; then
+              # we will try to update by first deleting
+              deletion_result=$(rpk acl user delete ${USER_NAME} 2>&1) && deletion_result_exit_code=$? || deletion_result_exit_code=$?
+              if [[ $deletion_result_exit_code -ne 0 ]]; then
+                echo "deletion of user ${USER_NAME} failed: ${deletion_result}"
+                break
+              fi
+              # Now we update the user
+              update_result=$(rpk acl user create ${USER_NAME} --password=${PASSWORD} --mechanism ${MECHANISM} {{ template "rpk-flags-no-sasl" $ }} 2>&1) && update_result_exit_code=$? || update_result_exit_code=$?  # On a non-success exit code
+              if [[ $update_result_exit_code -ne 0 ]]; then
+                echo "updating of user ${USER_NAME} failed: ${update_result}"
+                break
+              else
+                echo "Updated user ${USER_NAME}." 
+              fi
+            else
+              # Another error occurred, so output the original message and exit code
+              echo "error creating user ${USER_NAME}: ${creation_result}"
+              break
+            fi
+          # On a success, the user was created so output that
+          else
+            echo "Created user ${USER_NAME}."
+          fi
+        done < $USERS_FILE
+      }
+  
+      # first time processing
+      process_users $USERS_DIR
+    
+      # subsequent changes detected here
+      # watching delete_self as documented in https://ahmet.im/blog/kubernetes-inotify/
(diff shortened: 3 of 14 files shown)


alejandroEsc:
Blocking progress as I may need help on rpk to continue progress here.

.github/ct.yaml
```diff
@@ -15,7 +15,7 @@
 debug: true
 remote: origin
 target-branch: main
-helm-extra-args: --timeout 600s
+helm-extra-args: --timeout 900s
```
joejulian, line 18: Should we try setting this back?

joejulian (approved):

joejulian:
:tada:

charts/redpanda/values.yaml
```diff
@@ -537,6 +537,10 @@ statefulset:
     fsGroup: 101
     runAsUser: 101
     fsGroupChangePolicy: OnRootMismatch
+  sideCars:
+    configWatcher:
+      enabled: true
+      resources: {}
```
joejulian, line 543: Oh! Don't we need to make sure this is in the json schema for the Redpanda controller to import?

alejandroEsc: Thank you! that was in my todos and i completely dropped the ball.

.github/ct.yaml
alejandroEsc: I will retry

charts/redpanda/templates/secrets.yaml
```diff
@@ -136,3 +136,114 @@ stringData:
...
+  {{- with include "full.labels" . }}
+  {{- . | nindent 4 }}
+  {{- end }}
+type: Opaque
+stringData:
+  sasl-user.sh: |-
+    #!/usr/bin/env bash
+    set -e
+    
+    sleep 10 # wait for admin api to start
```
RafalKorepta, line 159: https://github.com/redpanda-data/helm-charts/blob/9b20abf81e80b2f265a3e75335c002d2ea9cf5c7/charts/redpanda/templates/statefulset.yaml#L264 

This should be good check for Admin API readiness.

alejandroEsc: Awesome, yes I completely forgot about this, thank you!

charts/redpanda/values.yaml
alejandroEsc: @joejulian the error still fails with a lower timeout unfortunately 
```
2023-05-18T13:22:53.1639354Z >>> helm install redpanda-pznmh9x5ml ct_previous_revision860339983/charts/redpanda --namespace redpanda-pznmh9x5ml --wait --values ct_previous_revision860339983/charts/redpanda/ci/12-external-cert-secrets-values.yaml --timeout 600s --set=image.tag=v22.3.17
2023-05-18T13:32:54.5034755Z Error: INSTALLATION FAILED: timed out waiting for the condition
```
on this run with the reverted timeout: https://github.com/redpanda-data/helm-charts/actions/runs/5014336900/jobs/8988471288?pr=496

charts/redpanda/templates/secrets.yaml
```diff
@@ -136,3 +136,114 @@ stringData:
...
+      process_users $USERS_DIR
+    
+      # subsequent changes detected here
+      # watching delete_self as documented in https://ahmet.im/blog/kubernetes-inotify/
+      USERS_FILE=$(find ${USERS_DIR}/* -print)
+      while RES=$(inotifywait -q -e delete_self ${USERS_FILE}); do
+        process_users $USERS_DIR
+      done
+    {{- end }}
+    sleep 60
```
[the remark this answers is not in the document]
alejandroEsc: We are adding a new container to the pod, in the case that there is no sasl enabled we cannot shut off this container or the pod will fail and be restarted. This just ensures that if we are running this sidecar it remains running.

charts/redpanda/templates/secrets.yaml
```diff
@@ -136,3 +136,114 @@ stringData:
...
+        READ_LIST_SUCCESS=0
+        while read p; do
+          IFS=":" read -r USER_NAME PASSWORD MECHANISM <<< $p
+          # Do not process empty lines
+          if [ -z "$USER_NAME" ]; then
+            continue
+          fi
+          echo "Creating user ${USER_NAME}..."
+          MECHANISM=${MECHANISM:-{{- include "sasl-mechanism" . }}}
+          creation_result=$(rpk acl user create ${USER_NAME} --password=${PASSWORD} --mechanism ${MECHANISM} {{ template "rpk-flags-no-sasl" $ }} 2>&1) && creation_result_exit_code=$? || creation_result_exit_code=$?  # On a non-success exit code
```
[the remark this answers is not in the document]
alejandroEsc: Not a stupid question at all, its the style that was there originally and i found nothing necessarily wrong with it. If you prefer to convert it I could if you prefer.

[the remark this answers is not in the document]
RafalKorepta: It can stay as is. I personally can not get used to bash scripting.

charts/redpanda/templates/secrets.yaml
```diff
@@ -136,3 +136,119 @@ stringData:
...
+          # On a success, the user was created so output that
+          else
+            echo "Created user ${USER_NAME}..."
+            USERS_LIST=$(new_users_list "${USERS_LIST}" "${USER_NAME}")
+          fi
+        done < $USERS_FILE
+            
+        if [[ -n "${USERS_LIST}" && ${READ_LIST_SUCCESS} ]]; then
+          echo "Setting superusers configurations with users [${USERS_LIST}]"
+          superuser_result=$(rpk cluster config set superusers [${USERS_LIST}] {{ template "rpk-flags-no-sasl" $ }} 2>&1) && superuser_result_exit_code=$? || superuser_result_exit_code=$?
```
RafalKorepta, line 233: Isn't helm chart suppose to manage `redpanda.yaml` file with the super users list? 

https://github.com/redpanda-data/helm-charts/blob/9b20abf81e80b2f265a3e75335c002d2ea9cf5c7/charts/redpanda/templates/configmap.yaml#L70

https://github.com/redpanda-data/helm-charts/blob/9b20abf81e80b2f265a3e75335c002d2ea9cf5c7/charts/redpanda/templates/configmap.yaml#L114

charts/redpanda/templates/statefulset.yaml
```diff
@@ -321,6 +321,30 @@ spec:
             limits:
               cpu: {{ .Values.resources.cpu.cores }}
               memory: {{ .Values.resources.memory.container.max }}
+      {{- if .Values.statefulset.sideCars.configWatcher.enabled }}
+        - name: config-watcher
+          image: {{ .Values.image.repository }}:{{ template "redpanda.tag" . }}
+          command:
+            - /bin/bash # could be expanded for multiple scripts
+            - -c 
+            - /etc/secrets/config-watcher/scripts/sasl-user.sh
```
RafalKorepta, line 330: If we plan to have more operation could we change the name from `sasl-user.sh` to `reconcile`?

charts/redpanda/templates/statefulset.yaml
```diff
@@ -321,6 +321,30 @@ spec:
...
+          volumeMounts:
+            - name: {{ template "redpanda.fullname" . }}-config-watcher
+              mountPath: /etc/secrets/config-watcher/scripts
+              readOnly: true
+          {{- if and .Values.auth.sasl.enabled (not (empty .Values.auth.sasl.secretRef )) }}
+            - name: {{ .Values.auth.sasl.secretRef }}
+              mountPath: /etc/secrets/users
+              readOnly: true
+          {{- end }}
```
RafalKorepta, line 339: Will this be even allowed to start from sasl disabled to then move to sasl enabled? If I recall stateful set will not allow to change volume and volumeMounts definition.

charts/redpanda/templates/tests/test-sasl-updated.yaml
```diff
@@ -0,0 +1,88 @@
...
+      image: {{ .Values.image.repository }}:{{ template "redpanda.tag" . }}
+      command:
+        - /usr/bin/timeout
+        - "120"
+        - bash
+        - -c
+        - |
+          set -xe
+          sleep 30
+          {{ include "rpk-cluster-info" $rpk }}
```
RafalKorepta, line 51: What this rpk call tests?

charts/redpanda/Chart.yaml
```diff
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 4.0.17
+version: 4.0.18
```
RafalKorepta, line 26: I think you need to bump to 
```suggestion
version: 5.0.0
```

It is because you will not be able to pass upgrade tests as you changes Stateful Set definition by adding volumes and volumeMounts

charts/redpanda/templates/secrets.yaml
alejandroEsc: Only for bootstraping and only if the userlist is not empty. In the case you are providing your own user list this will not be accepted. In the case of updating list, this list will be out of date.

charts/redpanda/templates/statefulset.yaml
alejandroEsc: this isn't a change to volumes being mounted, it is an addition to the new container so it can mount the same volume. And yes the new changes alllow us to move from disabled to enabled which was a known issue before.

RafalKorepta (approved):

PR #498, opened 2023-05-15
add a ServiceMonitor to scrape RedPanda metrics

### LINKED ISSUE #149: support prometheus servicemonitors
it would be great if the chart would support servicemonitors for prometheus.
this would involve two things:

- create a service monitor template
- add back the `admin` port in the `service.internal.yaml` as the servicemonitor targets `/metrics` path at the admin port

prune998:
Fixes #149

### DIFF
--- a/charts/redpanda/templates/service.internal.yaml
+++ b/charts/redpanda/templates/service.internal.yaml
@@ -24,7 +24,7 @@ metadata:
   name: {{ include "redpanda.fullname" . }}
   namespace: {{ .Release.Namespace | quote }}
   labels:
-    prometheus/scrape: "true"
+    monitoring/enabled: "{{ .Values.monitoring.enabled }}"
 {{- with include "full.labels" . }}
   {{- . | nindent 4 }}
 {{- end }}
--- a/charts/redpanda/templates/service.internal.yaml
+++ b/charts/redpanda/templates/service.internal.yaml
@@ -24,7 +24,7 @@ metadata:
   name: {{ include "redpanda.fullname" . }}
   namespace: {{ .Release.Namespace | quote }}
   labels:
-    monitoring/enabled: "{{ .Values.monitoring.enabled }}"
+    monitoring.redpanda.com/enabled: {{ .Values.monitoring.enabled | quote }}
 {{- with include "full.labels" . }}
   {{- . | nindent 4 }}
 {{- end }}
--- a/charts/redpanda/templates/servicemonitor.yaml
+++ b/charts/redpanda/templates/servicemonitor.yaml
@@ -33,7 +33,7 @@ spec:
     targetPort: admin
   selector:
     matchLabels:
-      monitoring/enabled: "true"
+      monitoring.redpanda.com/enabled: "true"
       app.kubernetes.io/name: {{ template "redpanda.name" . }}
       app.kubernetes.io/instance: {{ .Release.Name | quote }}
 {{- end }}
--- a/charts/redpanda/templates/servicemonitor.yaml
+++ b/charts/redpanda/templates/servicemonitor.yaml
@@ -28,12 +28,12 @@ metadata:
 {{- end }}
 spec:
   endpoints:
-  - interval: 30s
+  - interval: {{ .Values.monitoring.scrape_interval }}
     path: /public_metrics
     targetPort: admin
   selector:
     matchLabels:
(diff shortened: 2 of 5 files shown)


charts/redpanda/templates/service.internal.yaml
```diff
@@ -16,14 +16,15 @@ limitations under the License.
...
-# application
+# the stateful set and allow the serviceMonitor to target the pods.
+# This service should not be used by any client application
 apiVersion: v1
 kind: Service
 metadata:
   name: {{ include "redpanda.fullname" . }}
   namespace: {{ .Release.Namespace | quote }}
   labels:
+    prometheus/scrape: "true"
```
joejulian, line 27: should this be gated on monitoring.enabled?

charts/redpanda/templates/service.internal.yaml
```diff
@@ -34,3 +35,8 @@ spec:
   selector:
     app.kubernetes.io/name: {{ template "redpanda.name" . }}
     app.kubernetes.io/instance: {{ .Release.Name | quote }}
+  ports:
+    - name: admin
+      protocol: TCP
+      targetPort: {{ .Values.listeners.admin.port }}
+      port: {{ .Values.listeners.admin.port }}
```
joejulian, line 42: Should this also be gated?

charts/redpanda/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,39 @@
...
+metadata:
+  name: {{ include "redpanda.fullname" . }}
+  namespace: {{ .Release.Namespace | quote }}
+  labels:
+{{- with include "full.labels" . }}
+  {{- . | nindent 4 }}
+{{- end }}
+spec:
+  endpoints:
+  - interval: 30s
```
joejulian, line 31: Should this use a value under monitoring?

charts/redpanda/templates/service.internal.yaml
joejulian: Is this even needed if you create a ServiceMonitor? iirc, you don't need both.

prune998: I actually set this specific label so we can differenciate between this service and the others services from the chart, but sure I can gate it

prune998: The `serviceMonitor` needs to select a `Service` based on labels. So I need to be able to pick the right one between the `service.internal`, `service.loadbalancer` and `service.nodeport` as they all show the same list of labels...

joejulian: Got it! Thanks.

joejulian (approved):

charts/redpanda/templates/service.internal.yaml
```diff
@@ -16,14 +16,15 @@ limitations under the License.
...
-# application
+# the stateful set and allow the serviceMonitor to target the pods.
+# This service should not be used by any client application
 apiVersion: v1
 kind: Service
 metadata:
   name: {{ include "redpanda.fullname" . }}
   namespace: {{ .Release.Namespace | quote }}
   labels:
+    monitoring/enabled: "{{ .Values.monitoring.enabled }}"
```
joejulian, line 27: To distinguish labels that are specific to our use case, it's good practice to add a domain.

```suggestion
    monitoring.redpanda.com/enabled: "{{ .Values.monitoring.enabled }}"
```

charts/redpanda/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,39 @@
...
+  {{- . | nindent 4 }}
+{{- end }}
+spec:
+  endpoints:
+  - interval: {{ .Values.monitoring.scrapeInterval }}
+    path: /public_metrics
+    targetPort: admin
+  selector:
+    matchLabels:
+      monitoring/enabled: "true"
```
joejulian, line 36:
```suggestion
      monitoring.redpanda.com/enabled: "true"
```

joejulian (commented):
Will need a minor version bump in Chart.yaml as this is a new feature. That will fix the linter.

prune998:
not sure what to do with the failed check... ?

joejulian:
It's not a required check. It fails if there are any changes to the values files to ensure no accidental changes are merged.

Looks good, thanks!

PR #519, opened 2023-05-25
add a test summary that we can mark as a blocker

joejulian:
This adds a "summary" job that will pass or fail depending on the pass/fail state of all the matrix tests. The summary test can then be added to the required status checks.

### DIFF
--- a/.github/workflows/pull_requests.yaml
+++ b/.github/workflows/pull_requests.yaml
@@ -162,3 +162,13 @@ jobs:
             --skip-missing-values \
             --excluded-charts console \
             --target-branch ${{ github.event.repository.default_branch }}
+  summary:
+    if: always()
+    needs:
+      - test
+    runs-on: ubuntu-latest
+    steps:
+      - name: Summarize the results of the test matrix pass/fail
+        uses: re-actors/alls-green@release/v1
+        with:
+          jobs: ${{ toJSON(needs)}}
--- a/.github/workflows/pull_requests.yaml
+++ b/.github/workflows/pull_requests.yaml
@@ -166,7 +166,7 @@ jobs:
     if: always()
     needs:
       - test
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/pull_requests.yaml
```diff
@@ -162,3 +162,13 @@ jobs:
             --skip-missing-values \
             --excluded-charts console \
             --target-branch ${{ github.event.repository.default_branch }}
+  summary:
+    if: always()
+    needs:
+      - test
```
RafalKorepta, line 168:
```suggestion
    needs:
      - test
      - lint
```

.github/workflows/pull_requests.yaml
```diff
@@ -162,3 +162,13 @@ jobs:
...
+    runs-on: ubuntu-latest
```
RafalKorepta, line 169:
```suggestion
    runs-on: ubuntu-22.04
```

.github/workflows/pull_requests.yaml
```diff
@@ -162,3 +162,13 @@ jobs:
...
+    steps:
+      - name: Summarize the results of the test matrix pass/fail
+        uses: re-actors/alls-green@release/v1
+        with:
+          jobs: ${{ toJSON(needs)}}
```
RafalKorepta, line 174:
```suggestion
        with:
          allowed-failures: check-values
          jobs: ${{ toJSON(needs)}}
```

.github/workflows/pull_requests.yaml
joejulian: No, those can be checked directly. Only the matrix tests need this.

.github/workflows/pull_requests.yaml
joejulian: again, this is only for the matrix tests.

RafalKorepta (approved):

PR #521, opened 2023-05-26
test chart-testing github groups

joejulian:


RafalKorepta (approved):
I personally download CI logs to search anything offline using my IDE. Never the less this is good improvement if I would review some failure on my mobile phone :)

PR #534, opened 2023-05-31
Do not attempt to check TLS state for dry-runs

phooijenga:
When running `helm template` or passing `--dry-run`, the built-in `lookup` function always returns an empty map. This results in the check thinking that TLS is currently not enabled for RPCs, generating an error.

To reproduce:

```
$ helm install redpanda redpanda/redpanda
...
$ helm upgrade redpanda redpanda/redpanda --dry-run
Error: UPGRADE FAILED: execution error at (redpanda/templates/statefulset.yaml:60:28): 

Error: Cannot do a rolling restart to enable or disable tls at the RPC layer: changing listeners.rpc.tls.enabled (redpanda.yaml:repdanda.rpc_server_tls.enabled) from false to true
***WARNING The following instructions will result in a short period of downtime.
To accept this risk, run the upgrade again adding `--set force=true` and do the following:

While helm is upgrading the release, manually delete ALL the pods:
    kubectl -n infra delete pod -l app.kubernetes.io/component=redpanda-statefulset

If you got here thinking rpc tls was already enabled, see technical service bulletin 2023-01.

$ helm upgrade redpanda redpanda/redpanda
Release "redpanda" has been upgraded. Happy Helming!
...
```

### DIFF
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 4.0.29
+version: 4.0.30
 
 # The app version is the default version of Redpanda to install.
 appVersion: v23.1.10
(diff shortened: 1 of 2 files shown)


phooijenga:
I noticed the lint action failed because the branch did not include a version bump, so I rebased and added one.

RafalKorepta (approved):

PR #544, opened 2023-06-14
Exclude schema registry and HTTP proxy config from 22.2.X deployment

RafalKorepta:
As version 22.2.X does not provide basic auth configuration option in both schema registry and panda proxy the client configuration for both services needs SASL credentails to be explicitly set in Redpanda configuration. That Redpanda configuration is generated in helm and sotred as config map which should not hold any credentails.

Newer version can take HTTP basic auth header to impersonate user and authenticate against Redpanda Kafka listener.

Additionally HTTP proxy and schema registry tests now are run when SASL is enabled

### DIFF
--- a/charts/redpanda/templates/_helpers.tpl
+++ b/charts/redpanda/templates/_helpers.tpl
@@ -538,6 +538,14 @@ than 1 core.
 {{- toJson (dict "bool" (or (not (eq .Values.image.repository "docker.redpanda.com/redpandadata/redpanda")) (include "redpanda.semver" . | semverCompare ">=22.2.10-0,<22.3"))) -}}
 {{- end -}}
 
+{{- define "redpanda-22-2-x-without-sasl" -}}
+{{- $result :=  (include "redpanda-atleast-22-3-0" . | fromJson).bool -}}
+{{- if or (include "sasl-enabled" . | fromJson).bool .Values.listeners.kafka.authenticationMethod -}}
+{{-   $result := false -}}
+{{- end -}}
+{{- toJson (dict "bool" $result) -}}
+{{- end -}}
+
 # manage backward compatibility with renaming podSecurityContext to securityContext
 {{- define "pod-security-context" -}}
 fsGroup: {{ dig "podSecurityContext" "fsGroup" .Values.statefulset.securityContext.fsGroup .Values.statefulset }}
--- a/charts/redpanda/templates/tests/test-schemaregistry-internal-tls-status.yaml
+++ b/charts/redpanda/templates/tests/test-schemaregistry-internal-tls-status.yaml
@@ -14,10 +14,11 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */}}
-{{- if and (include "schemaRegistry-internal-tls-enabled" . | fromJson).bool (not (include "sasl-enabled" .|fromJson).bool) -}}
+{{- if and (include "schemaRegistry-internal-tls-enabled" . | fromJson).bool .Values.listeners.schemaRegistry.enabled (include "redpanda-22-2-x-without-sasl" . | fromJson).bool -}}
   {{- $service := .Values.listeners.schemaRegistry -}}
   {{- $cert := get .Values.tls.certs $service.tls.cert -}}
   {{- $root := deepCopy . }}
+  {{- $sasl := .Values.auth.sasl }}
 apiVersion: v1
 kind: Pod
 metadata:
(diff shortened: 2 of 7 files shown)


RafalKorepta:
Tests are blocked by https://github.com/redpanda-data/helm-charts/pull/547

RafalKorepta:
Test 05 was adjusted to use not default settings https://github.com/redpanda-data/helm-charts/pull/550. I hope upgrade tests will pass now.

charts/redpanda/templates/_helpers.tpl
```diff
@@ -538,6 +538,10 @@ than 1 core.
 {{- toJson (dict "bool" (or (not (eq .Values.image.repository "docker.redpanda.com/redpandadata/redpanda")) (include "redpanda.semver" . | semverCompare ">=22.2.10-0,<22.3"))) -}}
 {{- end -}}
 
+{{- define "redpanda-22-2-x-without-sasl" -}}
+{{- toJson (dict "bool" (or (not (or (include "sasl-enabled" . | fromJson).bool .Values.listeners.kafka.authenticationMethod)) (include "redpanda-atleast-22-3-0" . | fromJson).bool)) -}}
```
joejulian, line 542: If this is accurate, I'd suggest this. The nested logic is way to hard to parse in my head.

```suggestion
{{- $result :=  (include "redpanda-atleast-22-3-0" . | fromJson).bool -}}
{{- if or (include "sasl-enabled" . | fromJson).bool .Values.listeners.kafka.authenticationMethod)) -}}
{{-   $result := false -}}
{{- end -}}
{{- toJson (dict "bool" $result -}}
```

charts/redpanda/templates/tests/test-schemaregistry-internal-tls-status.yaml
```diff
@@ -41,40 +42,94 @@ spec:
...
+
+          curl -svm3 --fail --retry "120" --retry-max-time "120" --retry-all-errors \
+          -X POST -H "Content-Type: application/vnd.schemaregistry.v1+json" \
+          -d '{"schema": "{\"type\":\"record\",\"name\":\"sensor_sample\",\"fields\":[{\"name\":\"timestamp\",\"type\":\"long\",\"logicalType\":\"timestamp-millis\"},{\"name\":\"identifier\",\"type\":\"string\",\"logicalType\":\"uuid\"},{\"name\":\"value\",\"type\":\"long\"}]}"}' \
+          {{- if or (include "sasl-enabled" .|fromJson).bool .Values.listeners.schemaRegistry.authenticationMethod }}
+          -u $USERNAME:$PASSWORD \
+          {{- end }}
+          {{- if $cert.caEnabled }}
+          --cacert /etc/tls/certs/{{ $service.tls.cert }}/ca.crt \
+          {{- end }}
```
joejulian, line 70: This is repeated for each command. What do you think about putting it in a function?

```
schemaCurl () {
          {{- if or (include "sasl-enabled" .|fromJson).bool .Values.listeners.schemaRegistry.authenticationMethod }}
          -u $USERNAME:$PASSWORD \
          {{- end }}
          {{- if $cert.caEnabled }}
          --cacert /etc/tls/certs/{{ $service.tls.cert }}/ca.crt \
          {{- end }} $1
}
```

```
schemaCurl https://{{ include "redpanda.internal.domain" . }}:{{ .Values.listeners.schemaRegistry.port }}/subjects/sensor-value/versions
```

RafalKorepta: Great suggestion

charts/redpanda/templates/tests/test-schemaregistry-status.yaml
```diff
@@ -38,14 +39,64 @@ spec:
...
+          {{- end }}
+          http://{{ include "redpanda.internal.domain" . }}:{{ .Values.listeners.schemaRegistry.port }}/schemas/types
+
+          curl -svm3 --fail --retry "120" --retry-max-time "120" --retry-all-errors \
+          -X POST -H "Content-Type: application/vnd.schemaregistry.v1+json" \
+          -d '{"schema": "{\"type\":\"record\",\"name\":\"sensor_sample\",\"fields\":[{\"name\":\"timestamp\",\"type\":\"long\",\"logicalType\":\"timestamp-millis\"},{\"name\":\"identifier\",\"type\":\"string\",\"logicalType\":\"uuid\"},{\"name\":\"value\",\"type\":\"long\"}]}"}' \
+          {{- if or (include "sasl-enabled" .|fromJson).bool .Values.listeners.schemaRegistry.authenticationMethod }}
+          -u $USERNAME:$PASSWORD \
+          {{- end }}
+          http://{{ include "redpanda.internal.domain" . }}:{{ .Values.listeners.schemaRegistry.port }}/subjects/sensor-value/versions
```
joejulian, line 62: similar to above, could this be made a little more DRY?

charts/redpanda/templates/_helpers.tpl
RafalKorepta: I'm not that proficient in templating langue. If I could code this in go :)

Comment included into next push

joejulian (commented):
The only change I'd like to see for sure is the breaking up of the compound logic. The other suggestions are just nice-to-haves.

charts/redpanda/templates/_helpers.tpl
joejulian: It IS go! :grin:

joejulian (approved):

PR #550, opened 2023-06-15
Adjust super user SASL mechanism in test case 05

RafalKorepta:
Last changes did not change SASL mechanism in test `05-one-node-cluster-tls-sasl-values.yaml`

REF: https://github.com/redpanda-data/helm-charts/actions/runs/5278404502/jobs/9547707008?pr=544


alejandroEsc (approved):
lgtm,

PR #555, opened 2023-06-19
Retry openssl commands

RafalKorepta:
Due to openssl command failure retries was added.

REF:
https://github.com/redpanda-data/helm-charts/actions/runs/5302957882/jobs/9598259166#step:14:18369
https://github.com/redpanda-data/helm-charts/actions/runs/5306814731/jobs/9604764858#step:14:18928

### DIFF
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 4.0.41
+version: 4.0.42
 
 # The app version is the default version of Redpanda to install.
 appVersion: v23.1.12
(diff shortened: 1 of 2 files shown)


RafalKorepta:
I used 

https://github.com/redpanda-data/helm-charts/pull/555/commits/c9eccab0bd1142b6ab7dfa473bece587919b62cd 
if it is not reachable see https://github.com/RafalKorepta/helm-charts-1/commit/c9eccab0bd1142b6ab7dfa473bece587919b62cd

to validate that nightly pass. 

The CI execution https://github.com/redpanda-data/helm-charts/actions/runs/5307618681 as a proof.

The TODO commit will be removed.

alejandroEsc (approved):
I am ok to approve this, but is causing the need here, is it the time to come up? install?

RafalKorepta:
I'm not following your comment.

> I am ok to approve this, but is causing the need here

The problem is test is failing in CI not in dev environment. I can only think that cause is somewhere along DNS handling. I'm not sure if there is another solution to wait for external domain to be available. 

>  is it the time to come up? install?

I don't understand those questions.

Issue #394, opened 2023-03-20
Allow configuration of additional_start_flags in rpk section of the redpanda chart configmap


hcoyote:
### What would you like to be added?

allow setting of `additional_start_flags` in the the `rpk:` section of the redpanda chart config map.

### Why is this needed?

we occasionally need to set additional Redpanda startup flags that do not exist with equivalent configuration settings. In non-k8s installs, we would insert something like the following into the redpanda.yaml. The example entries are Seastar configuration flags that don't have a direct config file equivalent.

```
rpk:
    additional_start_flags:
        - "--idle-poll-time-us=0"
        - "--poll-aio=0"
```

We don't appear to be able to insert additional things into the helm-generated version of the redpanda.yaml. There's an rpk section, but it's fairly controlled on what we allow to go into that part of the configmap.

RafalKorepta:
I'm questioning this need as we support `additionalRedpandaCmdFlags`
https://github.com/redpanda-data/helm-charts/blob/756adfc58272e70c102c092344d3bf0b3c8e521c/charts/redpanda/templates/statefulset.yaml#L281-L283
https://github.com/redpanda-data/helm-charts/blob/756adfc58272e70c102c092344d3bf0b3c8e521c/charts/redpanda/values.yaml#L575-L576

If `rpk` does not support passing those additional Redpanda flags then it should be reported as bug in `rpk` (Redpanda mono repo).

hcoyote:
This support didn't exist when the issue was created. Looks like it was added a few weeks after by @alejandroEsc ? so if this commit solves the problem, awesome.

https://github.com/redpanda-data/helm-charts/commit/f0013eb53bc847237765ab7cc2ade535fba9d424

alejandroEsc:
@hcoyote I am a bit skeptical about just dropping support for this. There are reasons to have certain things added to the config file and other options to just flags. If you look at the current config we add some tuning items there now and not on the startup flags. I think there is room to make this change. That being said, is this option supported by flags today? If so then i can just lower the priority of this issue.

PR #562, opened 2023-06-22
chore:  rpk config field

alejandroEsc:
Fixes: #394 

Please check if the location of the new config makes sense. If it doesnt any suggestions welcome.

### DIFF
--- a/charts/redpanda/templates/configmap.yaml
+++ b/charts/redpanda/templates/configmap.yaml
@@ -118,7 +118,7 @@ data:
       superusers: {{ toJson $users }}
   {{- end }}
   {{- with (dig "cluster" dict .Values.config) }}
-      {{- range $key, $element := .}}
+      {{- range $key, $element := . }}
         {{- if or (eq (typeOf $element) "bool") $element }}
       {{ $key }}: {{ $element | toYaml }}
         {{- end }}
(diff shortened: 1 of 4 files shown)


charts/redpanda/templates/configmap.yaml
```diff
@@ -429,6 +429,23 @@ data:
 {{- /* END LISTENERS */}}
 
     rpk:
+{{- with (dig "rpk" dict .Values.config) }}
```
alejandroEsc, line 432: This seems overly complicated, we could simplify if we were not concerned with configs being empty/null etc.. let me know if rpk configs are not as concerned as we are with redpanda.

RafalKorepta: They should not be as much concerned as Redpanda config fileds.

RafalKorepta:
https://github.com/redpanda-data/helm-charts/issues/394#issuecomment-1604129160 I'm not sure if this should be even merged.

alejandroEsc:
from a values file entry of:
```
config:
  rpk:
    additional_start_flags:
      - "--idle-poll-time-us=0"
      - "--foo=bar"
    something: else
    blahblah: true
    additionally:
      blahblha: true
      something: no
```

we get in the configmap:

```
    rpk:
      additional_start_flags:
      - --idle-poll-time-us=0
      - --foo=bar
      additionally:
        blahblha: true
        something: false
      blahblah: true
      something: else
```

RafalKorepta (approved):

PR #572, opened 2023-06-30
Do not test node_config in v22.2.11

RafalKorepta:
In v22.2.11 the node config does not return enable rack awareness property.

https://github.com/redpanda-data/helm-charts/actions/runs/5418326297/jobs/9850304789#step:16:17462


alejandroEsc (approved):
lgtm

PR #579, opened 2023-07-07
use pod names for load-balancer addresses if address list is unset

joejulian:
If `external.addresses` is empty, use the pod name for the hostname when creating external dns entries for load-balancer services.


joejulian:
merge this after #578

alejandroEsc (approved):
lgtm

PR #586, opened 2023-07-18
chore: allow schema for kafkaEndpoint to allow dashes

alejandroEsc:
The schema for kafkaEndpoints should also allow dashes, here we add them.


RafalKorepta (approved):
Did I mention that I hate regex

joejulian (approved):

PR #597, opened 2023-07-22
bump redpanda appVersion to 23.2.2

joejulian:


RafalKorepta (approved):
I'm not sure why bumping major version, but it doesn't bother me that much.

joejulian:
Core has said that the quarterly release is a major version.

RafalKorepta:
> Core has said that the quarterly release is a major version.

Agree, but App version shouldn't be reflected with helm chart version. The chart does not have any braking change.

PR #600, opened 2023-07-24
Retry resource group deletion as it not always delete resource group

RafalKorepta:
Few resource was left behind even when deletion operation finished. The same command is called to remove the rest leftovers.

REF:

https://buildkite.com/redpanda/helm-charts/builds/470#018985aa-71a4-4331-8f8e-f62c1bb9c054


alejandroEsc (approved):
lgtm

RafalKorepta:
Trying to re-trigger Github action pipelines

gousteris:
discussed with @RafalKorepta and I force merged the PR without meeting the gh status checks requirements, as nothing was triggered/related to this PR

PR #608, opened 2023-07-25
Remove unused chart releaser definition

RafalKorepta:
```
Notice that if no config file is specified, cr.yaml (or any of the supported formats) is loaded from the current directory, $HOME/.cr, or /etc/cr, in that order, if found.
```


alejandroEsc (approved):
lgtm

PR #611, opened 2023-07-26
Fix operator CI tests

RafalKorepta:


charts/operator/templates/deployment.yaml
```diff
@@ -47,6 +47,8 @@ spec:
         - --health-probe-bind-address=:8081
         - --metrics-bind-address=127.0.0.1:8080
         - --leader-elect
+        - --configurator-tag={{ include "operator.tag" . }}
+        - --configurator-base-image=redpandadata/configurator
```
joejulian, line 51:
```suggestion
        - --configurator-base-image=docker.redpanda.com/redpandadata/configurator
```

joejulian (approved):
One minor suggestion but LGTM.

charts/operator/templates/deployment.yaml
RafalKorepta: Even better I replaced the hardcoded values with the values.yaml.

PR #612, opened 2023-07-26
Rename Github pipeline to indicate correctly which chart is checked

RafalKorepta:


joejulian (approved):
I was going to say this wouldn't work since the required test name wouldn't match, but it did. :exploding_head: I can't look at branch protections anymore to see why it worked, but good enough for me.

PR #614, opened 2023-07-26
feat: add prometheus for testing

alejandroEsc:
- Adding prometheus as part of our testing
- Adding service-monitor tests
- Added a test to ensure that no targets our down for our deployment
- Added prometheus deployment to nightly, nightly_tip, pull_requests, and pull_from_origin

### DIFF
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 5.1.0
+version: 5.0.7
 
 # The app version is the default version of Redpanda to install.
 appVersion: v23.2.3
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 5.0.6
+version: 5.1.0
 
 # The app version is the default version of Redpanda to install.
 appVersion: v23.2.3
--- a/charts/redpanda/ci/14-prometheus-no-tls-values.yaml
+++ b/charts/redpanda/ci/14-prometheus-no-tls-values.yaml
@@ -1,0 +1,20 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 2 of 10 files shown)


charts/redpanda/Chart.yaml
```diff
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 5.0.4
+version: 5.0.7
```
RafalKorepta, line 26: NIT: Why such bump?

From 5.0.4 to 5.0.7 seems 3 patch version jump.

charts/redpanda/ci/14-prometheus-no-tls-values.yaml
```diff
@@ -0,0 +1,25 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+---
+tls:
+  enabled: false
+
+statefulset:
+  initContainers:
+    setDataDirOwnership:
+      enabled: true
```
RafalKorepta, line 22: Why this needs to be enabled?

charts/redpanda/templates/tests/test-prometheus-targets.yaml
```diff
@@ -0,0 +1,50 @@
...
+      image: mintel/docker-alpine-bash-curl-jq:latest
+      command: [ "/bin/bash", "-c" ]
+      args:
+        - |
+          set -xe
+          RESULT=$( curl --fail http://prometheus-operated.default.svc.cluster.local:9090/api/v1/targets?status=active | jq . | grep '"health": "up"' | wc -l  )
+          if [ $RESULT != {{ .Values.statefulset.replicas }} ]; then 
+              echo "targets are down, only have:\n ${RESULT} targets 'up'"
+              exit 1
+          fi
```
RafalKorepta, line 49: From the test output
```
  + RESULT=21
  + '[' 21 '!=' 3 ']'
  + echo 'targets are down, only have:\n 21 targets '\''up'\'''
  + exit 1
  targets are down, only have:\n 21 targets 'up'
```
I wonder if targets have namespace label, so that you can compare correct targets. 21 number seems odd to me.

alejandroEsc: been trying to get this right. ill check again. The prometheus api does not make this easy, so parsing is required

charts/redpanda/Chart.yaml
alejandroEsc: I was counting on a few of your commits to merge at the time i made that change. Ill update in a few.

charts/redpanda/ci/14-prometheus-no-tls-values.yaml
alejandroEsc: Naw, ill revert, this was me using the metrics example given to me for testing service monitory.

charts/redpanda/templates/tests/test-prometheus-targets.yaml
alejandroEsc: this is working locally, strangely enough, given the same setup...

RafalKorepta (approved):

PR #615, opened 2023-07-26
Create first version of connectors helm chart

RafalKorepta:
Next step would be to create tests and Github action pipeline.


charts/connectors/templates/deployment.yaml
```diff
@@ -0,0 +1,208 @@
...
+            matchLabels:  {{ include "connectors-pod-labels" $root | nindent 14 }}
+          maxSkew: {{ $spread.maxSkew }}
+          topologyKey: {{ $spread.topologyKey }}
+          whenUnsatisfiable: {{ $spread.whenUnsatisfiable }}
+    {{- end }}
+      volumes:
+      {{- if and .Values.auth.sasl.userName .Values.auth.sasl.mechanism .Values.auth.sasl.secretRef }}
+        - name: rc-credentials
+          secret:
+            defaultMode: 292
```
joejulian, line 202:
```suggestion
            defaultMode: 0o444
```

charts/connectors/templates/deployment.yaml
```diff
@@ -0,0 +1,208 @@
...
+            secretName: {{ .Values.auth.sasl.secretRef }}
+      {{- end }}
+        - emptyDir:
+            medium: Memory
+            sizeLimit: {{ .Values.deployment.connectTmpSizeLimit }}
+          name: rp-connect-tmp
```
joejulian, line 208: Should this be templated? I can imagine users that don't want ram disks when they have nvme disks.

charts/connectors/values.yaml
```diff
@@ -0,0 +1,247 @@
...
+      status: _internal_connectors_status
+  groupID: connectors-cluster
+
+# -- Authentication settings.
+# For details,
+# see the [SASL documentation](https://docs.redpanda.com/docs/manage/kubernetes/security/sasl-kubernetes/).
+auth:
+  sasl:
+    # -- The authentication mechanism to use for the superuser. Options are `scram-sha-256` and `scram-sha-512`.
+    mechanism: scram-sha-256
```
joejulian, line 97: we default to 512 in the redpanda chart, should we do the same here?

joejulian (approved):
Few nits, but otherwise LGTM.

charts/connectors/templates/deployment.yaml
RafalKorepta: ram disks is quicker than nvme disks. I can templated it sure.

charts/connectors/values.yaml
RafalKorepta: sure I will switch to 512

charts/connectors/templates/deployment.yaml
```diff
@@ -0,0 +1,204 @@
...
+                status.storage.redpanda.remote.write={{ .Values.connectors.storage.remote.write.status }}
+                offset.storage.replication.factor={{ .Values.connectors.storage.replicationFactor.offset }}
+                config.storage.replication.factor={{ .Values.connectors.storage.replicationFactor.config }}
+                status.storage.replication.factor={{ .Values.connectors.storage.replicationFactor.status }}
+                producer.linger.ms={{ .Values.connectors.producerLingerMS }}
+                producer.batch.size={{ .Values.connectors.producerBatchSize }}
+                config.providers=file,secretsManager
+                config.providers.file.class=org.apache.kafka.common.config.provider.FileConfigProvider
+                {{- if .Values.connectors.secretManager.enabled }}
+                config.providers.secretsManager.class=com.github.jcustenborder.kafka.config.aws.SecretsManagerConfigProvider
```
tomasz-sadura, line 109: this is for AWS only, for GCP config is different. Please also note the providers were forked and the classes changed:

- `config.providers.secretsManager.class=com.redpanda.connectors.config.provider.aws.SecretsManagerConfigProvider` for AWS

- `config.providers.secretsManager.class=com.redpanda.connectors.config.provider.gcp.SecretManagerConfigProvider` for GCP

PR #623, opened 2023-07-28
Fix values diff for operator

RafalKorepta:


alejandroEsc (approved):
correct, sorry i missed this.

PR #624, opened 2023-07-28
Connectors chart fixes and e2e test

RafalKorepta:
Github action workflow that executes first connectors chart test.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,5 +25,6 @@ charts/redpanda/templates/external-service.yaml
 charts/redpanda/templates/external-tls-secret.yaml
 charts/redpanda/templates/some-users-updated.yaml
 charts/kminion/templates/redpanda-tls.yaml
+charts/connectors/templates/hidden-only-for-ci/
 
 ct_previous_*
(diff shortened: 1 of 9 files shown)


.gitignore
```diff
@@ -25,5 +25,6 @@ charts/redpanda/templates/external-service.yaml
 charts/redpanda/templates/external-tls-secret.yaml
 charts/redpanda/templates/some-users-updated.yaml
 charts/kminion/templates/redpanda-tls.yaml
+charts/connectors/templates/redpanda-tls.yaml
```
joejulian, line 28: Could we just move all these temporary files into a subdirectory and just exclude that directory?

charts/connectors/templates/deployment.yaml
```diff
@@ -195,6 +225,24 @@ spec:
           whenUnsatisfiable: {{ $spread.whenUnsatisfiable }}
     {{- end }}
       volumes:
+      {{- if .Values.connectors.brokerTLS.ca.secretRef }}
+        - name: truststore
+          secret:
+            defaultMode: 0444
```
joejulian, line 231: The modern yaml standard is `0o444`

joejulian (approved):

.gitignore
RafalKorepta: Agree. This PR will do it for connectors only. Rest will be done in different PR.

PR #328, opened 2023-02-09
console: add changes to allow this to be used as a subchart

joejulian:
These changes are needed to use the console chart as a subchart.

PR #625, opened 2023-07-29
connectors: add changes to allow this to be used as a subchart

RafalKorepta:
Similar how console deployment and configmap is rendered now connectors can be used as subchart

REF:
https://github.com/redpanda-data/helm-charts/pull/328


charts/connectors/templates/deployment.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Values.deployment.create -}}
```
alejandroEsc, line 1: nit: can we move this to line 18

charts/connectors/templates/tests/01-mm2-values.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Values.test.create -}}
```
alejandroEsc, line 1: nit: same here.

alejandroEsc (approved):
lgtm, this is just setting up for allowing you to subchart while overriding at the top level for the redpanda chart, correct?

charts/connectors/templates/tests/01-mm2-values.yaml
RafalKorepta: Done

charts/connectors/templates/deployment.yaml
RafalKorepta: Done

PR #630, opened 2023-08-01
Fix nightly build

RafalKorepta:
```
ct lint --all --config .github/ct-redpanda.yaml
Linting charts...
Error: failed loading configuration: specifying both, '--all' and '--charts', is not allowed
failed loading configuration: specifying both, '--all' and '--charts', is not allowed
Error: Process completed with exit code 1.
```

https://github.com/redpanda-data/helm-charts/actions/runs/5687162122/job/15415137752#step:6:11


.github/workflows/nightly.yaml
```diff
@@ -43,7 +43,7 @@ jobs:
           version: v3.8.0-1
 
       - name: Run chart-testing (lint)
-        run: ct lint --all --config .github/ct-redpanda.yaml
+        run: ct lint --config .github/ct.yaml
```
joejulian, line 46: Don't we still need
```suggestion
        run: ct lint --config .github/ct-redpanda.yaml
```

RafalKorepta: I was thinking that nightly should lint all charts. That's why I bring back ct.yaml configuration. WDYT?

alejandroEsc: What is the intent of the nightly now? I thought it was to run tests only for the redpanda chart, after all we have a matrix. Should we have a nightly for all charts that just checks linting?

RafalKorepta: reverted

RafalKorepta:
I pushed one correction to this fix as none of the Github action workflow was triggered. Our PR requires `summary` and `lint` jobs.

PR #636, opened 2023-08-03
Add operator RBAC role to update pod status

RafalKorepta:
This is follow up form https://github.com/redpanda-data/redpanda/pull/12167


alejandroEsc (approved):
lgtm

alejandroEsc (approved):
nice, we should review these roles soon.

PR #641, opened 2023-08-04
chores: set default log level to debug

alejandroEsc:
The helm operators from flux require that we have at least a debug log level in order to see the progress and errors happening when the helmrelease controller is trying to reconcile. Having this set by default will help with us with customer issues in the near future.


joejulian:
Can we only set debug logging for the flux controllers?

alejandroEsc:
> Can we only set debug logging for the flux controllers

I will check again, but it was my impression that this was challenging with some side affects.

PR #645, opened 2023-08-07
chore: allow extra init containers, volumes, and volumemounts

### LINKED ISSUE #533: Request template for initContainers
### What happened?

In case if we have self-managed kafka cluster signed by self-signed certificate, we need to add this certificate to trusted.

For it we have the next folder `/etc/ssl/certs/ca-certificates.crt ` where locates all the trusted certificate.
In this case we don't have any initContainer for adding this cert to trusted. It will be fine to implement it.

Adding new configs and possibilities to helm chart

Something like this can be possible:

```
{{- if .Values.initContainers }}
initContainers:
{{- range .Values.initContainers }}
- name: {{ .name }}
  image: "{{ .image }}:{{ .tag }}" -> can be possible to set the same image as running under `redpandadata/console`
  imagePullPolicy: {{ .imagePullPolicy }}
  command: {{ .command }}
  args:
  {{ toYaml .args | indent 10 }}

  {{- if .configmap }}
  envFrom:
  - configMapRef:
      name: {{ .name }}
  {{- end }}

  {{- if .volumeMounts }}
  volumeMounts:
  {{- range .volumeMounts }}
  - name: {{ .name }}
    mountPath: {{ .mountPath }}
  {{- if .subPath }}
    subPath: {{ .subPath }}
  {{- end }}
  {{- if .readOnly }}
    readOnly: {{ .readOnly }}
  {{- end }}
  {{- end }}
  {{- end }}
{{- end }}
{{- end }}
```

```
values:
  initContainers: []
```

Console

### Chart Version(s)

Using the latest one: `0.6.4`

### Cloud provider

<details>

- AWS

</details>

alejandroEsc:
Fixes #533

### DIFF
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 5.0.9
+version: 5.0.10
 
 # The app version is the default version of Redpanda to install.
 appVersion: v23.2.4
--- a/charts/redpanda/templates/statefulset.yaml
+++ b/charts/redpanda/templates/statefulset.yaml
@@ -85,6 +85,9 @@ spec:
             runAsUser: 0
             runAsGroup: 0
           volumeMounts:
+{{- if dig "initContainers" "tuning" "extraVolumeMounts" false .Values.statefulset -}}
+{{ tpl .Values.statefulset.initContainers.tuning.extraVolumeMounts . | nindent 12 }}
+{{- end }}
             - name: {{ template "redpanda.fullname" . }}
               mountPath: /etc/redpanda
         {{- if get .Values.statefulset.initContainers.tuning "resources" }}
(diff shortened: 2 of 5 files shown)


charts/redpanda/Chart.yaml
```diff
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 5.0.9
+version: 5.0.10
```
joejulian, line 26: This new feature seems worthy of a minor release, imo.
```suggestion
version: 5.1.0
```

charts/redpanda/templates/statefulset.yaml
```diff
@@ -85,6 +85,9 @@ spec:
             runAsUser: 0
             runAsGroup: 0
           volumeMounts:
+{{- if get .Values.statefulset.initContainers.tuning "extraVolumeMounts" -}}
```
joejulian, line 88: initContainers on down are not required parameters in the schema. We should either require them or use dig for each of these:
```suggestion
{{- if dig  "initContainers" "tuning" "extraVolumeMounts" false .Values.statefulset -}}
```

alejandroEsc: oh i see what you mean, alright.

charts/redpanda/templates/statefulset.yaml
```diff
@@ -199,13 +208,19 @@ spec:
...
+{{ tpl .Values.statefulset.initContainers.configurator.extraVolumeMounts . | nindent 12 }}
+{{- end }}
             - name: {{ template "redpanda.fullname" . }}
               mountPath: /tmp/base-config
             - name: config
               mountPath: /etc/redpanda
         {{- if get .Values.statefulset.initContainers.configurator "resources" }}
           resources: {{- toYaml .Values.statefulset.initContainers.configurator.resources | nindent 12 }}
         {{- end }}
+{{- if get .Values.statefulset.initContainers "extraInitContainers" -}}
```
joejulian, line 221: missed one

alejandroEsc: too many

joejulian (approved):

PR #650, opened 2023-08-08
Chore: update appVersion for operator

alejandroEsc:
Also added operator to nightly version checks


RafalKorepta:
I don't know if it is worth separated PR but connectors should be included too.

PR #659, opened 2023-08-12
rename test file and make test name match it

joejulian:


charts/connectors/templates/tests/01-connectors-mm2-test.yaml
```diff
@@ -18,7 +18,7 @@ limitations under the License.
 apiVersion: v1
 kind: Pod
 metadata:
-  name: {{ include "connectors.fullname" . }}-mm2-test-{{ randNumeric 3 }}
```
RafalKorepta, line 21: This random numbers is important as there are cases where helm install can cause problems with tests after helm upgrade. That's why I randomized names so that CR will print everything.

joejulian: With the `before-hook-creation` annotation, the pod with the same name will be deleted first.

alejandroEsc (approved):
lgtm, but is this a problem with upgrades being performed or tested? @RafalKorepta can you describe the problems with tests? Will CI capture this?

RafalKorepta:
> can you describe the problems with tests?

After the helm upgrade and second helm test invocation you will no longer be able to see in the `ct` output what mirror maker 2 test generated in the output. The curl commands might pass successfully, but it would be good to verify if everything was as expected. I remember there was some bug I found that thanks to running commands manually by hand (without ct) and examine the test Pods output.

joejulian:
> After the helm upgrade and second helm test invocation you will no longer be able to see in the `ct` output what mirror maker 2 test generated in the output. The curl commands might pass successfully, but it would be good to verify if everything was as expected. I remember there was some bug I found that thanks to running commands manually by hand (without ct) and examine the test Pods output.

Aren't there logs prior to the upgrade?

RafalKorepta:
No, the test Pod is removed with it's logs.

joejulian:
Right, I mean doesn't CT run tests after the install but before the upgrade that output logs?

RafalKorepta:
> Right, I mean doesn't CT run tests after the install but before the upgrade that output logs?

No. The order:

- helm instal
- helm test
- helm upgrade
- helm test
- fail
- output

PR #663, opened 2023-08-15
Add a kustomize base for operator

joejulian:
This will create a `kustomize` branch for releases of templates that could be rendered with `kustomize`. Initially, this will include the `operator` chart.


alejandroEsc (approved):
wow, awesome idea. Thank you!

joejulian:
bypassed checks because they're not relevant to this change.

PR #664, opened 2023-08-15
use the correct named template for the service account name in the rolebinding

joejulian:
Also use `include` instead of `template` for consistency.

Associated with redpanda-data/cloudv2/issues/7740


alejandroEsc (approved):
nice catches overall, lgtm

PR #666, opened 2023-08-16
Probes now target their own pod

### LINKED ISSUE #584: Cluster-wide unreadiness makes operations harder
### What happened?

After 40s of downtime for a single broker, all pods in the replicaset go unready.

I would expect a health check that indicates health per pod, to aid alerting and troubleshooting. Furthermore given such a probe, a service _without_ `publishNotReadyAddresses` could be used for admin API traffic and possibly client bootstrap to avoid hitting a less healthy replica or a replica on a temporarily troubled node.

I suppose this is a matter of opinion. However, if the only reason for ["It's ok that this cluster-wide check affects all the pods"](https://github.com/redpanda-data/helm-charts/blob/redpanda-4.0.53/charts/redpanda/templates/statefulset.yaml#L258) is the PodDisruptionBudget, I'd like to point out that with max unavailable 1 it will prevent "roll any pods" just as well if only the unhealthy pod went unready.

In operations seeing this result from `get pods` is quite frightening, but everything might actually be fine. Maybe redpanda-2 started up slowly for example because it triggered node scale-up.
image

values:

```
statefulset:
  replicas: 3
```

`kubectl scale --replicas=2 statefulset redpanda && sleep 40 && kubectl scale --replicas=3 statefulset redpanda`

_No response_

Redpanda

### Chart Version(s)

4.0.53

### Cloud provider

<details>

</details>

solsson:
This means that unresponsiveness is reflected in individual pods' readiness. Anecdotally I see more predictable unreadiness behavior at rolling upgrade.

Closes #584 while it's by design that all pods can go unready based on rpk cluster health.

### DIFF
--- a/charts/redpanda/templates/statefulset.yaml
+++ b/charts/redpanda/templates/statefulset.yaml
@@ -264,7 +264,7 @@ spec:
                 - -c
                 - |
                   set -e
-                  RESULT=$(curl --silent --fail -k {{ include "admin-tls-curl-flags" . }} "{{ include "admin-http-protocol" . }}://{{ include "admin-api-urls" . }}/v1/status/ready")
+                  RESULT=$(curl --silent --fail -k {{ include "admin-tls-curl-flags" . }} "{{ include "admin-http-protocol" . }}://localhost:{{.Values.listeners.admin.port }}/v1/status/ready")
                   echo $RESULT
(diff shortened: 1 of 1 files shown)


solsson:
See https://github.com/redpanda-data/helm-charts/issues/584#issuecomment-1680972014

Issue #644, opened 2023-08-07
Support customize redpanda service name


sniperking1234:
### What would you like to be added?

I would like the internal access address within Kubernetes to be configured as `kafka-0.kafka-headless.common.svc.cluster.local:9092`, but currently, it can only be configured as `kafka-0.kafka.common.svc.cluster.local:9092`.

### Why is this needed?

I previously installed Kafka using the chart package from https://github.com/bitnami/charts/tree/main/bitnami/kafka. The access address after installation is as mentioned above. I would like other applications to seamlessly migrate to Redpanda without needing to modify the access address.

joejulian:
Seems reasonable. Do you think this is something you might be interested in contributing?

sniperking1234:
I will give it a try.

RafalKorepta:
achieved in https://github.com/redpanda-data/helm-charts/pull/674

Thank you @sniperking1234

PR #674, opened 2023-08-21
Support customize redpanda service name

sniperking1234:
https://github.com/redpanda-data/helm-charts/issues/644

### DIFF
--- a/charts/redpanda/templates/_helpers.tpl
+++ b/charts/redpanda/templates/_helpers.tpl
@@ -34,6 +34,17 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this 
 {{- end -}}
 
 {{/*
+Create a default service name
+*/}}
+{{- define "redpanda.servicename" -}}
+{{- if dig "service" "name" false .Values.AsMap -}}
+{{- .Values.service.name | trunc 63 | trimSuffix "-" -}}
+{{- else -}}
+{{ include "redpanda.fullname" . | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
 full helm labels + common labels
 */}}
 {{- define "full.labels" -}}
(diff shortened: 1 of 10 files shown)


joejulian (approved):
Just needs a minor version bump of the Chart.yaml

sniperking1234:
@joejulian done.

charts/redpanda/templates/_helpers.tpl
```diff
@@ -33,6 +33,17 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
...
+{{/*
+Create a default service name
+*/}}
+{{- define "redpanda.servicename"}}
+{{- if .Values.service.name}}
+{{- .Values.service.name | trunc 63 | trimSuffix "-" -}}
+{{- else }}
+{{ include "redpanda.fullname" . | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+{{- end}}
```
joejulian, line 45:
```suggestion
{{- define "redpanda.servicename" -}}
{{- if dig "service" "name" false .Values -}}
{{- .Values.service.name | trunc 63 | trimSuffix "-" -}}
{{- else -}}
{{ include "redpanda.fullname" . | trunc 63 | trimSuffix "-" -}}
{{- end -}}
{{- end -}}
```
If we use dig here, we won't need to require that the service key exists.

Also, added trailing `-` to match our style for named templates.

charts/redpanda/values.yaml
```diff
@@ -56,6 +56,10 @@ image:
   # If `image.tag` is 'latest', the default is `Always`.
   pullPolicy: IfNotPresent
 
+# -- Redpanda Service settings.
+service:
+  name: redpanda
+
```
joejulian, line 62:
```suggestion
# -- Redpanda Service settings.
# service:
#   -- set service.name to override the default service name
#   name: redpanda

```

joejulian (changes requested):
CI is failing because you included a default service.name that changes the default behavior.

On upgrades, this fails because the existing external service that was named for the release and the new one named redpanda have conflicting ports.

I've added a couple suggestions that should solve that.

charts/redpanda/templates/_helpers.tpl
sniperking1234: Thank you for your suggestion.

joejulian:
Sorry about the bad suggestion. I pushed a fix.

One last thing, could you squash this PR please?

sniperking1234:
@joejulian It's already done.

joejulian (approved):
LGTM!

PR #684, opened 2023-08-23
Redpanda decommissioning

alejandroEsc:
This PR will be split into two, however this is a preview of what changes are needed to allow for decommissioning to work.


alejandroEsc:
test will not pass as a sidecar is passing flags that do not exist in the operator, this will depend on the redpanda PR to merge: https://github.com/redpanda-data/redpanda/pull/12847

alejandroEsc:
Will be splitting this up into two PRs

charts/redpanda/templates/rbac.yaml
```diff
@@ -33,9 +33,31 @@ rules:
     - ""
     resources:
       - nodes
+      - secrets
```
RafalKorepta, line 36: why secret is added here?

charts/redpanda/templates/statefulset.yaml
```diff
@@ -389,6 +389,27 @@ spec:
...
+      {{- if and .Values.rbac.enabled .Values.statefulset.sideCars.controllers.enabled }}
+        - name: redpanda-controllers
+          image: {{ .Values.statefulset.sideCars.controllers.image.repository }}:{{ .Values.statefulset.sideCars.controllers.image.tag }}
+          command:
+            - /manager
+          args:
+            - --operator-mode=false
+            - --namespace={{ .Release.Namespace }}
+            - --health-probe-bind-address=:8085
+            - --metrics-bind-address=:9082
```
RafalKorepta, line 401: Can you move those ports to values.yaml?

They should not collide with Redpanda ports.

charts/redpanda/values.yaml
```diff
@@ -584,6 +584,15 @@ statefulset:
...
       securityContext: {}
       extraVolumeMounts: |-
+    controllers:
+      image:
+        tag: v23.2.5
+        repository: docker.redpanda.com/redpandadata/redpanda-operator
+      enabled: false # experimental, must opt in
+      resources: {}
+      securityContext: {}
+      run:
```
RafalKorepta, line 594: I wouldn't name it `run`, but rather `controllerNames`.
```suggestion
      controllerNames:
```

charts/operator/templates/role.yaml
```diff
@@ -184,6 +186,14 @@ rules:
...
       - update
       - watch
+  - apiGroups:
+      - ""
+    resources:
+      - pods
+    verbs:
+      - get
+      - list
+      - watch
```
RafalKorepta, line 196: Why this is added? I was thinking that only sidecar not the operator need Kubernetes Role definition change. Was there missing permission and this is bugfix?

charts/operator/templates/role.yaml
```diff
@@ -316,5 +326,16 @@ rules:
...
+      - ""
+    resources:
+      - persistentvolumeclaims
+    verbs:
+      - delete
+      - get
+      - list
+      - patch
+      - update
+      - watch
```
RafalKorepta, line 339: ditto

charts/redpanda/templates/statefulset.yaml
alejandroEsc: This is why they were chosen like this, yeah will move it to values.

charts/operator/templates/role.yaml
alejandroEsc: Both need permissions, not just the sidecar. Remember you can run this on the operator and on the sidecar

charts/redpanda/templates/rbac.yaml
alejandroEsc: In order to get values file we need to get and watch for secrets.

alejandroEsc:
> Will be splitting this up into two PRs

done.

charts/operator/templates/role.yaml
```diff
@@ -391,11 +381,6 @@ rules:
       - topics
     verbs:
       - create
-  - apiGroups:
-      - ""
-    resources:
-      - persistentvolumeclaims
-    verbs:
```
RafalKorepta, line 398: Why this is removed? Decommission capabilities will always be available as a side car? I thought it will be an option THIS or THAT.

charts/operator/templates/role.yaml
```diff
@@ -94,8 +94,6 @@ rules:
       - get
       - patch
       - update
-      - list
-      - watch
```
RafalKorepta, line 98: This seems suspicious. Doesn't flux controller use list or watch capabilities from Kubernetes client-go library?

charts/operator/templates/clusterrole.yaml
```diff
@@ -262,32 +262,3 @@ rules:
...
-    resources:
-      - persistentvolumes
-    verbs:
-      - delete
-      - get
-      - list
-      - patch
-      - update
-      - watch
-{{- end -}}
```
RafalKorepta, line 293: Maybe I'm missing, but will this removal break operator v1 capabilities to decommission Redpandas?

charts/redpanda/templates/rbac.yaml
```diff
@@ -57,16 +79,29 @@ rules:
...
+      - nodes
+    verbs:
+      - get
+      - list
+      - watch
+  - apiGroups:
+      - ""
+    resources:
+      - persistentvolumes
+      - persistentvolumeclaims
```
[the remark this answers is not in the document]
alejandroEsc: I think i had at some point this set to clusterrole, but yeah you are right it is not needed here.

charts/operator/templates/role.yaml
alejandroEsc: suspicious? Interesting choice of words. Yeah ill review the file.

charts/redpanda/templates/rbac.yaml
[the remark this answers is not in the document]
alejandroEsc: Oh i see, if you check line 60, you will notice that it was set to cluster scope already, I moved it to line 97 because i needed to add patch, update, watch to this.

charts/operator/templates/role.yaml
alejandroEsc: i have no idea what you mean. ill revert this file since there is no harm and I cannot remember why.

charts/operator/templates/clusterrole.yaml
alejandroEsc: This was not supposed to be here, removing it from the commits.

charts/operator/templates/role.yaml
alejandroEsc: This was not supposed to be here, only redpanda chart changes were supposed to here, removing.

charts/operator/templates/role.yaml
alejandroEsc: This was not supposed to be here, only RP changes are supposed to be here.

charts/redpanda/templates/rbac.yaml
RafalKorepta: I don't understand why Redpanda chart needs to have access to secrets at the Cluster level. 

In the new 2 controllers (nor decommission nor nodeWatcher) https://github.com/redpanda-data/redpanda/pull/12847 I don't see any of them to get nor watch for secrets. 

Maybe I'm missing something with the side car approach. Could you help me understand that?

charts/redpanda/templates/rbac.yaml
```diff
@@ -33,9 +33,31 @@ rules:
...
+  - apiGroups:
+      - cluster.redpanda.com
+    resources:
+      - redpandas
+    verbs:
+      - get
+      - list
+      - patch
+      - update
+      - watch
```
RafalKorepta, line 50: Maybe I don't understand the side car deployment model. I see that you disables operator mode with `--operator-mode=false`

https://github.com/redpanda-data/helm-charts/pull/684/files#diff-6f494fcb2c4edfa7256b33ba89ef8e7c83d92c7f0a6bf6bea53341881e224620R327

Could you help me understand why Redpandas Role rules are added here?

charts/redpanda/templates/rbac.yaml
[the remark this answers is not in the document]
RafalKorepta: I understand that side car need to have some new rules, but this particular place is used to define `rpk-bundle` ClusterRole rules. That's why I would like to propose to separate rpk-bundle specific rules that could be applied in helm chart after enabling some flag. 

My proposal would be to create new `Role` when the `statefulset.controllers.enabled` is set to `true`. What do you think about my proposal?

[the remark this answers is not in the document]
alejandroEsc: Makes sense. I was trying to keep things a bit more tidy to avoid having to create more template files. sure thing.

charts/redpanda/templates/rbac.yaml
alejandroEsc: A requirement that was moved over from the roles since I do need to list out the redpandas. However i believe i removed that requirement for the non-operator mode, so it should be safe to remove now.

charts/redpanda/templates/rbac.yaml
alejandroEsc: The need does not need to be explicit. If you look closely I use the helm library to get the values file. Helm itself stores stateful information regarding the releases in a secret. This is the chain that explains the need for it.

RafalKorepta (approved):

RafalKorepta (approved):

PR #687, opened 2023-08-27
provide client certs to servicemonitor CRDs

gugu:


charts/redpanda/templates/servicemonitor.yaml
```diff
@@ -38,10 +38,22 @@ spec:
     scheme: https
     tlsConfig:
       insecureSkipVerify: true
```
alejandroEsc, line 40: Should we skip this if we are applying the certs?

charts/redpanda/templates/servicemonitor.yaml
```diff
@@ -38,10 +38,22 @@ spec:
...
+      ca:
```
alejandroEsc, line 41: Only if CA is enabled right? That is an option in the certs stanza of the values file.

charts/redpanda/templates/servicemonitor.yaml
```diff
@@ -38,10 +38,22 @@ spec:
...
+        secret:
+          name: {{ template "redpanda.fullname" $ }}-default-cert
```
alejandroEsc, line 43: This is the default name in most cases, but not necessarily, we should check to see if a secret is being provided instead also.

alejandroEsc (changes requested):
Thank you for your contribution, and i have to say great suggestion for an improvement!

charts/redpanda/templates/servicemonitor.yaml
gugu: It makes HTTPS request to IP address (without hostname) and check SANs for the certificate (and it does not match)

david-yu:

PR #689, opened 2023-08-28
Support custom domain when using connectors and console

Lavaburn:
Replace hard-coded "cluster.local" by clusterDomain value

### DIFF
--- a/charts/redpanda/templates/console/configmap-and-deployment.yaml
+++ b/charts/redpanda/templates/console/configmap-and-deployment.yaml
@@ -61,7 +61,7 @@ limitations under the License.
   "enabled" $values.connectors.enabled
   "clusters" (list
     (dict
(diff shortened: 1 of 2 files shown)


charts/redpanda/templates/console/configmap-and-deployment.yaml
```diff
@@ -1,4 +1,4 @@
-{{/*
```
alejandroEsc, line 1: was this removed accident? If not then we need to do the same on the other end of the boilerplate

alejandroEsc (changes requested):
thank you for the contribution, this is looking great! just one ask

charts/redpanda/templates/console/configmap-and-deployment.yaml
Lavaburn: Sorry, that was accidental. I have fixed the commit now.

alejandroEsc (approved):
lgtm, thank you again!

alejandroEsc:
Updating branch, will merge once I am given the go ahead, there is another PR we are waiting for first,

RafalKorepta (approved):

PR #692, opened 2023-08-29
chore: calculate sha256 without rolling when replicas change

### LINKED ISSUE #690: Broker pod removal should should not trigger a restart due to updating Redpanda.yaml

Broker pod removal may update the seed servers conf in node conf ( redpanda.yaml ) if the pod being removed is one of those seed servers.

This blindly triggers a whole cluster rolling restart when its not needed.  The config map should be updated so fresh brokers get the updated list when they start, but there's no reason to proactively restart

### Why is this needed?

 this causes unneeded disruption to the cluster when downscaling (worse if downscaling multiple nodes)

alejandroEsc:
Fixes #690 

New approach allows configmap to be created and defined once. This should satisfy our needs, test below still apply.

default sha256
```
 helm template redpanda -n redpanda charts/redpanda  | grep "checksum/config"
        checksum/config: 01ba4719c80b6fe911b091a7c05124b64eeece964e09c058ef8f9805daca546b
        checksum/config: 9c28e0f53ca7c99cb3c5d84f4c68e5ac86e07c03d4ee7044a6ac0f88b896c341
```

then change replicas
```
helm template redpanda -n redpanda charts/redpanda --set statefulset.replicas=6  | grep "checksum/config"
        checksum/config: 01ba4719c80b6fe911b091a7c05124b64eeece964e09c058ef8f9805daca546b
        checksum/config: 9c28e0f53ca7c99cb3c5d84f4c68e5ac86e07c03d4ee7044a6ac0f88b896c341
```

or 

```
 helm template redpanda -n redpanda charts/redpanda --set statefulset.replicas=4  | grep "checksum/config"
        checksum/config: 01ba4719c80b6fe911b091a7c05124b64eeece964e09c058ef8f9805daca546b
        checksum/config: 9c28e0f53ca7c99cb3c5d84f4c68e5ac86e07c03d4ee7044a6ac0f88b896c341
```

In main branch we see

defaults
```
helm template redpanda -n redpanda charts/redpanda  | grep "checksum/config"
        checksum/config: 01ba4719c80b6fe911b091a7c05124b64eeece964e09c058ef8f9805daca546b
        checksum/config: 2f77333c60e715dbd7f1b2540a7da54f3e417ee4330752b1202d6c5cd8318409
```

changing reps
```
helm template redpanda -n redpanda charts/redpanda --set statefulset.replicas=6  | grep "checksum/config"
        checksum/config: 01ba4719c80b6fe911b091a7c05124b64eeece964e09c058ef8f9805daca546b
        checksum/config: f9bc461bf447c7b81cfbe5e79e3b2d7f8c4d35030a806413591ebb428b6a4699
```


charts/redpanda/templates/_configmap.tpl
```diff
@@ -0,0 +1,463 @@
...
+
+{{- define "configmap-content-no-seed" -}}
+{{- /*
+  configmap content without seed list.
+*/ -}}
+{{- $root := . }}
+{{- $values := .Values }}
+
+{{- /*
+  It's impossible to do a rolling upgrade from not-tls-enabled rpc to tls-enabled rpc.
```
RafalKorepta, line 26: NIT: Interesting comment, is there work around to enable TLS on rpc?

charts/redpanda/templates/_configmap.tpl
```diff
@@ -0,0 +1,463 @@
...
+*/ -}}
+{{- $check := list
```
RafalKorepta, line 28: NIT

The `check` variable isn't meaningful. I don't even understand why we need at least 23.1.2 or 22.3.13 or 22.2.10? Could we name the feature or bugfix that is included in that version? 

```suggestion
{{- $check := list
```

RafalKorepta (approved):

charts/redpanda/templates/_configmap.tpl
alejandroEsc: This is just me copy pasting the configmap details here. Do you want me to take a look at this?

charts/redpanda/templates/_configmap.tpl
alejandroEsc: Will review, this is a great time to make these kind of changes.

charts/redpanda/templates/_configmap.tpl
joejulian: not as part of this PR

RafalKorepta: > not as part of this PR

@joejulian my question still stand as I would like to know if there is a work around. 

P.S. The PR is already approved. 

P.S. 2 I moved to convo to the slack.

joejulian (approved):

RafalKorepta (approved):

PR #698, opened 2023-09-02
feature: do not roll on replica changes

alejandroEsc:
The configuration script depended on the state of the replica count, which meant the init container was changed each time on each pod whenever the replicas was changed. We move the script into its own secret that gets mounted by the init container. 

That script will be run anew for each new broker that is created and so will update the configuration as needed. 

TODO

When adding a new broker, scaling up, this does not run the right rpk command.


alejandroEsc:
the tests passed, just updating from main and changing a variable name.

charts/redpanda/templates/secrets.yaml
```diff
@@ -257,3 +263,95 @@ stringData:
...
+  labels:
+  {{- with include "full.labels" . }}
+  {{- . | nindent 4 }}
+  {{- end }}
+type: Opaque
+stringData:
+  configurator.sh: |-
+    # TODO you have to bring back the range so that the external address list is ammended properly
+    # once that is done, by bringing back the range here, you then must fine a way to only pick up the listeners that pertain
+    # to the ordinal we are working with, cutting should be the right approach here
```
joejulian, line 281: Is this a note to self, or is this something that still needs done. If the latter, what do we get or miss by not doing it now?

alejandroEsc: will remove

charts/redpanda/templates/statefulset.yaml
```diff
@@ -73,10 +73,9 @@ spec:
         - name: tuning
           image: {{ .Values.image.repository }}:{{ template "redpanda.tag" . }}
           command:
-            - bash
+            - /bin/bash
```
joejulian, line 76:
```suggestion
            - rpk redpanda tune all
```

charts/redpanda/templates/statefulset.yaml
```diff
@@ -73,10 +73,9 @@ spec:
...
             - -c
-          args:
-            - |
+            - | 
               rpk redpanda tune all
```
joejulian, line 79:
```suggestion
```

charts/redpanda/templates/statefulset.yaml
```diff
@@ -124,88 +123,20 @@ spec:
 {{- end }}
         - name: {{ (include "redpanda.name" .) | trunc 51 }}-configurator
           image: {{ .Values.image.repository }}:{{ template "redpanda.tag" . }}
-          command: ["/bin/bash", "-c"]
+          command:
+            - /bin/bash
+            - -c
+            - | 
+              /etc/secrets/configurator/scripts/configurator.sh "${SERVICE_NAME}" "${KUBERNETES_NODE_NAME}"
```
joejulian, line 130: Perhaps we should do this for every container that requires a shell script to ensure sigterm is caught and used:
```
         env:
           - name: SCRIPT
           - value: /etc/secrets/configurator/scripts/configurator.sh
          command:
            - /bin/sh
            - -c
            - 'trap "exit 0" TERM; exec $SCRIPT & wait $!'
```

alejandroEsc: this is an init container, so should run only once right?

charts/redpanda/templates/statefulset.yaml
joejulian: This one didn't take.

charts/redpanda/templates/statefulset.yaml
joejulian: Yes, but I'm just thinking we should do this as best practice. If a pod is terminated during init, there's no reason to be forced to wait for the script to finish.

charts/redpanda/templates/statefulset.yaml
alejandroEsc: it did not work, the container would not start.

RafalKorepta: I can not believe that previously it wouldn't start. At least one test should catch that.

charts/redpanda/templates/secrets.yaml
```diff
@@ -14,6 +14,12 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */}}
+{{- $values := .Values -}}
+{{- $internalAdvertiseAddress := printf "%s.%s" "${SERVICE_NAME}" (include "redpanda.internal.domain" .) -}}
+{{- $externalAdvertiseAddress := printf "${SERVICE_NAME}" -}}
+{{- if (tpl ($values.external.domain | default "") $) -}}
+  {{- $externalAdvertiseAddress = printf "${SERVICE_NAME}.%s" (tpl $values.external.domain $) -}}
+{{- end -}}
```
RafalKorepta, line 22: This comes from statefulset.yaml, so why it was not removed from there?

RafalKorepta (approved):
The change LGTM.

I would like to understand what exactly in rpk command was wrong? I don't see this in diff nor in commit message.

In the TODO list you mentioned test 10 relay on replicaIndex, but nothing has changed in test folder.

I believe using chart testing this is impossible to simulate scale up/down. Maybe you can do it in Operator kuttle testing and verify that Pods are not roll over?

charts/redpanda/templates/secrets.yaml
alejandroEsc: GOOD POINT! I wasnt sure if it was being used for something else, ill check and remove if nothing else is using it.

alejandroEsc:
> The change LGTM.
> 
> I would like to understand what exactly in rpk command was wrong? I don't see this in diff nor in commit message.
> 
> In the TODO list you mentioned test 10 relay on replicaIndex, but nothing has changed in test folder.
> 
> I believe using chart testing this is impossible to simulate scale up/down. Maybe you can do it in Operator kuttle testing and verify that Pods are not roll over?



> The change LGTM.
> 
> I would like to understand what exactly in rpk command was wrong? I don't see this in diff nor in commit message.
> 
> In the TODO list you mentioned test 10 relay on replicaIndex, but nothing has changed in test folder.
> 
> I believe using chart testing this is impossible to simulate scale up/down. Maybe you can do it in Operator kuttle testing and verify that Pods are not roll over?

Those were not for the command in the code, that was for my changes. I was trying to simplify the process but then realized that the helpers being used did not really play nice with the pod ordinals due to type expectations. It can be done but it would require some work to make this happen correctly. So i decided to rever those changes for now and just move things here.

PR #701, opened 2023-09-05
feat: allow ability to disable external service creation

alejandroEsc:
Allow us to disable external service creation but while defining external services so we can create our own 

This will require a change to the api, ill add a follow PR to do this. 

Partially Fixes [#13271](https://github.com/redpanda-data/redpanda/issues/13271)

### DIFF
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 5.3.1
+version: 5.3.3
 
 # The app version is the default version of Redpanda to install.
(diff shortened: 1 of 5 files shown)


charts/redpanda/Chart.yaml
```diff
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 5.3.1
+version: 5.3.3
```
RafalKorepta, line 26: Why bump is +2? Should we have
```suggestion
version: 5.3.2
```

RafalKorepta (approved):

charts/redpanda/Chart.yaml
alejandroEsc: Yeah! but I am waiting for another PR to go in first ;)

PR #714, opened 2023-09-11
chore: additional rbac rules needed for 12847

alejandroEsc:
Second set of roles needed. These are statefulsets/status, i also reformatted these to prepare for automation later. 

This is for PR [redpanda-data/redpanda 12847](https://github.com/redpanda-data/redpanda/pull/12847)

### DIFF
--- a/charts/operator/templates/clusterrole.yaml
+++ b/charts/operator/templates/clusterrole.yaml
@@ -53,13 +53,6 @@ rules:
     - update
     - watch
 - apiGroups:
-    - apps
-  resources:
-    - statefulsets/status
-  verbs:
-    - patch
-    - update
-- apiGroups:
     - cert-manager.io
   resources:
     - certificates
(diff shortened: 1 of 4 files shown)


charts/operator/templates/clusterrole.yaml
```diff
@@ -17,250 +17,257 @@ metadata:
...
-    verbs:
-      - get
-      - patch
-      - update
+- apiGroups:
+    - ""
+  resources:
+    - persistentvolumes
+  verbs:
+    - delete
```
RafalKorepta, line 25: I can not find delete operation https://github.com/redpanda-data/redpanda/pull/12847 PR

charts/operator/templates/clusterrole.yaml
```diff
@@ -17,250 +17,257 @@ metadata:
...
+    - patch
+    - update
+    - watch
+- apiGroups:
+    - apps
+  resources:
+    - statefulsets/status
+  verbs:
+    - patch
+    - update
```
RafalKorepta, line 61: The `statefulsets/status` role rules are defined both in ClusterRole and Role. My assumption was that we stick to namespaced operator in v2. With introduction of decommission in helm chart I don't understand why we would need cluster scope permission for `statefulse/status`. Could you help me understand that?

RafalKorepta (approved):

PR #715, opened 2023-09-11
redpanda/feat: Allow to set affinity for jobs

koooge:
Hi there, there is no way to set `affinity` for redpanda-post-upgrade and redpanda-configuration Jobs. Allow affinity to be set as well as nodeSelector. statefulset is not changed in PR.

refs #305

### DIFF
--- a/charts/redpanda/values.yaml
+++ b/charts/redpanda/values.yaml
@@ -41,6 +41,10 @@ nodeSelector: {}
 # For details,
 # see the [Kubernetes documentation](https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/).
 tolerations: []
+# -- Allows to specify affinity for Jobs.
+# For details,
+# see the [Kubernetes documentation](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity).
+affinity: {}
 
 # -- Redpanda Docker image settings.
 image:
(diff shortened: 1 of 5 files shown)


charts/redpanda/values.yaml
```diff
@@ -41,6 +41,10 @@ nodeSelector: {}
 # For details,
 # see the [Kubernetes documentation](https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/).
 tolerations: []
+# -- Allows to specify affinity for Jobs.
```
alejandroEsc, line 44: Could you move this and place one in each of the following fields

```
post_upgrade_job:
```
and
```
post_install_job:
```
So they can take affinity rules to target those jobs only. The statefulsets have their own pod affinities also. We can later do something like this and then make the individual ones as overrides, i think we have used that pattern in multiple places at this point.

alejandroEsc (commented):
Thank you for this contributions, great catch on something we completely missed.

charts/redpanda/values.yaml
koooge: I moved it to each field in this MR. Yes, I want to override it globally but it would be also convenient for fine-tuning.

alejandroEsc (approved):
lgtm, thank you!

alejandroEsc:
waiting on CI to pass, do not worry about check-values, the error there will be expected.

joejulian (approved):
Thanks!

PR #718, opened 2023-09-12
Add default crash loop limit node property

RafalKorepta:
When cluster is experiencing allocation failures then it can start to accumulate very small segments that makes bootstraping of a particular node harder. The amount of memory that is requiured due to high amount of small segments on disk with possibility of managing tiered storage cache in paraller is greater every time Redpanda stops/crash. This crash loop limit prevents from accumulating those small segments.

User can disable this crash loop limit check by the following action:
* One hour elapses since the last crash
* The node configuration file, redpanda.yaml, is updated via config.cluster or config.node or config.tunable objects
* The startup_log file in the node’s data_directory is manually deleted

REF:
https://github.com/redpanda-data/redpanda/pull/8122
https://docs.redpanda.com/current/reference/node-properties/#crash_loop_limit

Fixes https://github.com/redpanda-data/helm-charts/issues/717

### DIFF
--- a/charts/redpanda/values.yaml
+++ b/charts/redpanda/values.yaml
@@ -940,7 +940,19 @@ config:
     # wait_for_leader_timeout_ms: 5000ms                           # Timeout (ms) to wait for leadership in metadata cache
   # -- Node (broker) properties.
   # See the [property reference documentation](https://docs.redpanda.com/docs/reference/node-properties/).
-  node: {}
+  node:
+    # -- Crash loop limit
+    # A limit on the number of consecutive times a broker can crash within one hour before its crash-tracking logic is reset.
+    # This limit prevents a broker from getting stuck in an infinite cycle of crashes.
+    # User can disable this crash loop limit check by the following action:
+    #
+    # * One hour elapses since the last crash
+    # * The node configuration file, redpanda.yaml, is updated via config.cluster or config.node or config.tunable objects
+    # * The startup_log file in the node’s data_directory is manually deleted
+    #
+    # Default to 5
+    # REF: https://docs.redpanda.com/current/reference/node-properties/#crash_loop_limit
+    crash_loop_limit: 5
     # node_id:                                                     # Unique ID identifying a node in the cluster
     # data_directory:                                              # Place where redpanda will keep the data
     # admin_api_doc_dir: /usr/share/redpanda/admin-api-doc         # Admin API doc directory
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 5.4.0
+version: 5.4.1
 
 # The app version is the default version of Redpanda to install.
 # ** NOTE for maintainers: please ensure the artifacthub image annotation is updated before merging
--- a/charts/redpanda/templates/_configmap.tpl
+++ b/charts/redpanda/templates/_configmap.tpl
@@ -129,7 +129,7 @@ limitations under the License.
   {{- end }}
       {{- with dig "node" dict .Values.config }}
       {{- range $key, $element := .}}
-        {{- if or (eq (typeOf $element) "bool") $element }}
+        {{- if and (or (eq (typeOf $element) "bool") $element) (and (eq $key "crash_loop_limit") (include "redpanda-atleast-23-1-1" $root | fromJson).bool) }}
       {{ $key }}: {{ $element | toYaml }}
         {{- end }}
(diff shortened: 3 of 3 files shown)


alejandroEsc (approved):
I am ok approving this, @JakeSCahill we should document this, its a change of behavior right? So it will impact a lot of customer, should we bump this to 5.4.0?

JakeSCahill:
This should go into release notes for the chart.

This new default will also appear in our reference doc here: https://docs.redpanda.com/current/reference/redpanda-helm-spec/#config-node

I'm not sure when a user should edit this value or disable it. Maybe there's a troubleshooting scenario where we can discuss the use of this property?

charts/redpanda/values.yaml
```diff
@@ -925,7 +925,8 @@ config:
     # wait_for_leader_timeout_ms: 5000ms                           # Timeout (ms) to wait for leadership in metadata cache
   # -- Node (broker) properties.
   # See the [property reference documentation](https://docs.redpanda.com/docs/reference/node-properties/).
-  node: {}
+  node:
+    crash_loop_limit: 5
```
joejulian, line 929: Can you please add a comment explaining where 5 comes from?

RafalKorepta: Great question. I asked @piyushredpanda for core guidance. I found only slack conversation about the default 

https://github.com/redpanda-data/helm-charts/issues/717#issuecomment-1715218308

P.S. Found yet another slack message https://redpandadata.slack.com/archives/C04B5T58QQ1/p1694554270861379?thread_ts=1694549537.193709&cid=C04B5T58QQ1

RafalKorepta: I would like to see good description in core documentation that I could copy-paste it here. At the moment I need to invent the reasoning.

piyushredpanda: This is more art than science. 5 "feels" enough unclean restarts. Not sure if there can be more to it. Agreed that it is better to have a comment on this somewhere and we should document (unsure if it was), @bharathv ?

joejulian: I'd be satisfied with, "5 is chosen with the guidance of the Redpanda core team"

bharathv: > This is more art than science. 5 "feels" enough unclean restarts. Not sure if there can be more to it. Agreed that it is better to have a comment on this somewhere and we should document (unsure if it was), @bharathv ?

Agree, the thought process here is if a broker crashed or shutdown uncleanly for 5 times back to back, something is off and needs a manual intervention.

RafalKorepta:
@bharathv I would like again ask the same question as @JakeSCahill asked already.

> I'm not sure when a user should edit this value or disable it. Maybe there's a troubleshooting scenario where we can discuss the use of this property?

My naive understanding is we should never disable this property. As much as possible stick to the default value of 5. In the troubleshooting scenario when Redpanda is in crash loop backoff it would mean that due to internal state Redpanda can not start the initialization process. User would need to find previous log, which might be order of magnitude further than 5 restarts before, and those logs should be send to Support Team. Normal `rpk support bundle` will not help here as `kubectl logs --previous --namespace $NS_PLACEHOLDER $REDPANDA_POD_NAME` does not gives us answers. Only central logging system would preserve logs when actual error happen. 

@bharathv What do you thing? Do you have different view based on my attempt to answer @JakeSCahill question?

RafalKorepta:
> So we need a temporary way of deactivating this config, right? You think deleting $DATA_DIR/startup_log is a feasible approach? (that disables backoff temporarily)

In current state Redpanda stores only  
```c++
    struct crash_tracker_metadata {
        uint32_t _crash_count{0};
        uint64_t _config_checksum{0};
        model::timestamp _last_start_ts;
    };
```

which doesn't gives us causes of the previous crashes to the STDOUT when Redpanda executes the check [application::check_for_crash_loop()](https://github.com/redpanda-data/redpanda/blob/7f512fb2e7f0a682e967196b68bc7c0bc7ce995e/src/v/redpanda/application.cc#L789-L799).

The Kubernetes (container runtime) will provide only last 2 logs. I agree that changing config option would be easier option just from usability perspective. In Kubernetes everything should be declared, Redpanda config included. 

@JakeSCahill Should I change the wording inside values.yaml helm docs? Or you would like to buble up this information to official Redpanda documentation page?

https://github.com/redpanda-data/helm-charts/pull/718/files#diff-df34075ea48d6120c348021a3727de5f513afd338dfa2e8378993b5b1f374d39R944-R951
```
User can disable this crash loop limit check by the following action:

* One hour elapses since the last crash
* The node configuration file, redpanda.yaml, is updated via config.cluster or config.node or config.tunable objects
* The startup_log file in the node’s data_directory is manually deleted
```

Issue #703, opened 2023-09-06
Redpanda Enterprise license not configured for Console


voutilad:
### What happened?

When configuring a license via `license_key`, the license is installed for the Redpanda cluster, but not for the Redpanda Console.

### What did you expect to happen?

Console should have the license key applied per: https://docs.redpanda.com/current/get-started/licenses/#apply-a-license-key-to-redpanda-console

### How can we reproduce it (as minimally and precisely as possible)?. Please include values file.

1. Generate a license and configure it in your `values.yaml` or via `--set` arguments.
2. Connect to Console. You'll see it detects the cluster has a license, but the Console enterprise features are locked out.

### Anything else we need to know?

See internal Slack conversation: https://redpandadata.slack.com/archives/C01H6JRQX1S/p1693947465843879

### Which are the affected charts?

Console

### Chart Version(s)

```
NAME                    NAMESPACE       REVISION        UPDATED                                 STATUS          CHART           APP VERSION
redpanda                redpanda        5               2023-09-05 21:19:31.690029619 +0000 UTC deployed        redpanda-5.3.1  v23.2.8
```

### Cloud provider

n/a

PR #719, opened 2023-09-12
feat: console - allow to define secret for license

alejandroEsc:
Partially fixes [#703](https://github.com/redpanda-data/helm-charts/issues/703)

### DIFF
--- a/charts/console/values.yaml
+++ b/charts/console/values.yaml
@@ -234,6 +234,13 @@ secret:
       # tlsCert:
       # tlsKey:
 
+# -- Settings for license key, as an alternative to secret.enterprise when
+# a license secret is available
+enterprise:
+  licenseSecretRef:
+    name: ""
+    key: ""
+
 # -- Settings for liveness and readiness probes.
 # For details,
 # see the [Kubernetes documentation](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#configure-probes).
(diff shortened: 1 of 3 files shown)


joejulian (approved):

charts/console/values.yaml
```diff
@@ -234,6 +234,13 @@ secret:
       # tlsCert:
       # tlsKey:
 
+# -- Settings for license key, as an alternative to secret.enterprise when
+# a license secret is available
```
RafalKorepta, line 238: NIT:

```suggestion
# -- Settings for Redpanda license, as an alternative to secret.enterprise because
# the secret.enterprise would be stored in helm values as plain text.
```

RafalKorepta, line 238: Maybe I don't understand previous intention, but for me secret.enterprise.license is a way to define plain text Redpanda license in values.yaml that is then propagated to secret configuration file 

https://github.com/redpanda-data/helm-charts/blob/eae186334a49c651fbc731dbc0baaa08ecc6d400/charts/console/templates/secret.yaml#L51

Because of that I would suggest to change the comment/documentation to reflect that `secret.enterprise.license` is a plain text representation and will be stored in helm release values. That precise note could be applied in both new `enterprise.licenseSecretRef` value and old `secret.enterprise.license` value.

alejandroEsc: Sure, we can start to make those kind of doc changes that are definitely needed.

RafalKorepta: So it will be addressed in the next PR or this conversation will not have follow up?

alejandroEsc: We can have a followup, this is just the first PR.

PR #728, opened 2023-09-17
Enable new topic custom resource RBAC rules

RafalKorepta:
As the topic controller will be enabled by default the RBAC rules for new topic custom resource needs to be added to cluster scoped type of deployment. The Kubernetes controller runtime would shutdown whole operator Pod due to missing RBAC permissions.

REF:
https://github.com/redpanda-data/redpanda/pull/11208

### DIFF
--- a/charts/operator/Chart.yaml
+++ b/charts/operator/Chart.yaml
@@ -6,7 +6,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 0.3.17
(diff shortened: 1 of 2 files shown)


RafalKorepta:
We are doing patch not delete, right?

RafalKorepta:
Like in this

https://github.com/redpanda-data/redpanda/blob/a07019c22d30c19f9de0a44c1146a75acf0a8515/src/go/k8s/controllers/redpanda/redpanda_controller.go#L327

The client Update executes patch, right?

PR #730, opened 2023-09-18
redpanda: Explain terminationGracePeriodSeconds

RafalKorepta:
Additional explanation for customers that reads defaults value in
artifacthub.io or Redpanda docs.

REF:
https://artifacthub.io/packages/helm/redpanda-data/redpanda?modal=values
https://docs.redpanda.com/current/reference/redpanda-helm-spec/#statefulset-terminationgraceperiodseconds

### DIFF
--- a/charts/redpanda/values.yaml
+++ b/charts/redpanda/values.yaml
@@ -640,6 +640,19 @@ statefulset:
   # -- Additional flags to pass to redpanda,
   additionalRedpandaCmdFlags: []
 #    - --unsafe-bypass-fsync
+  # -- Termination grace period in seconds is time required to execute preStop hook
+  # which puts particular Redpanda Pod (process/container) into maintenance mode.
+  # Before settle down on particular value please put Redpanda under load and perform
+  # rolling upgrade or rolling restart. That value needs to accommodate two processes:
+  # * preStop hook needs to put Redpanda into maintenance mode
+  # * after preStop hook Redpanda needs to handle gracefully SIGTERM signal
+  #
+  # Both processes are executed sequentially where preStop hook has hard deadline in the
+  # middle of terminationGracePeriodSeconds.
+  #
+  # REF:
+  # https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/#hook-handler-execution
+  # https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination
   terminationGracePeriodSeconds: 90
   ## Additional Volumes that you mount
   extraVolumes: |-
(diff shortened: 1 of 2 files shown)


charts/redpanda/values.yaml
```diff
@@ -640,6 +640,19 @@ statefulset:
...
 #    - --unsafe-bypass-fsync
+  # -- Termination grace period in seconds is time required to execute preStop hook
+  # which puts particular Redpanda Pod (process/container) into maintenance mode.
+  # Before settle down on particular value please put Redpanda under load and perform
+  # rolling upgrade or rolling restart. That value needs to accommodate two processes:
+  # * preStop hook needs to put Redpanda into maintenance mode
+  # * after preStop hook Redpanda needs to handle gracefully SIGTERM signal
+  #
+  # Both processes are executed sequentially where preStop hook has hard deadline in the
+  # middle of terminationGracePeriodSeconds.
```
JakeSCahill, line 651: Does this mean that the `preStop` hook only has 45 seconds by default and the rest is allocated to the `SIGTERM`?

RafalKorepta: Great question. It was working like that, but latest changes https://github.com/redpanda-data/helm-charts/commit/2b87951931b00d5c19cf062f7113238860c99b1e#diff-defacebd1c27c71d9807d81685dd0838ddd8392ca6084491962f85dca6bb2307L68 changed that behavior.

@joejulian Why we don't cap the preStop hook?

RafalKorepta: We where using timeout command 

REF: 
https://man7.org/linux/man-pages/man1/timeout.1.html

RafalKorepta: @joejulian send me the exact place where we are using timeout function:
* https://github.com/redpanda-data/helm-charts/blob/4f599dcc17d941b3e83f84c292faf51f0740e122/charts/redpanda/templates/statefulset.yaml#L177
* https://github.com/redpanda-data/helm-charts/blob/4f599dcc17d941b3e83f84c292faf51f0740e122/charts/redpanda/templates/statefulset.yaml#L185

You can find that `{{ div .Values.statefulset.terminationGracePeriodSeconds 2 }}` is used to divide period by 2.

joejulian: So yes, if the `terminationGracePeriodSeconds` is configured to the default of 90 seconds, preStop will always be stopped by 45 seconds after which redpanda will get its SIGTERM.

If preStop finishes earlier, redpanda will get whatever's left at that time.

joejulian (approved):

PR #733, opened 2023-09-18
use a helper for volumes and mounts

joejulian:


alejandroEsc:
Great reduction of duplication. Could we move this to another helper file, something like _volume_helper.tpl ? the helper file is pretty big as is and it would make those pieces a bit more readable.

charts/redpanda/templates/tests/test-kafka-internal-tls-status.yaml
```diff
@@ -59,26 +59,10 @@ spec:
...
-      volumeMounts:
-        - name: {{ template "redpanda.fullname" . }}
-          mountPath: /tmp/base-config
-        - name: config
-          mountPath: /etc/redpanda
+      volumeMounts: {{ include "default-mounts" . | nindent 8 }}
   {{- range $name, $cert := .Values.tls.certs }}
         - name: redpanda-{{ $name }}-cert
           mountPath: {{ printf "/etc/tls/certs/%s" $name }}
   {{- end }}
```
RafalKorepta, line 66: Why this is left behind?

RafalKorepta: I see it being defined in https://github.com/redpanda-data/helm-charts/pull/733/files#diff-8407f32f870c88088a3041f8a5a8394806973dabdef154b7fde83d5e77f70ed8R697-R702

RafalKorepta (approved):
LGTM

charts/redpanda/templates/tests/test-kafka-internal-tls-status.yaml
joejulian: I just really enjoy making CI fail over and over while I go through and find all the mistakes I made. :facepalm:

PR #741, opened 2023-09-20
use redpanda.yaml for rpk commands instead of flags

joejulian:
configure the rpk section of redpanda.yaml to satisfy the configuration
needs of rpk to avoid using flags everywhere.

Add instructions for creating an rpk profile so external rpk users can
also run commands without having to set flags.

### DIFF
--- a/charts/redpanda/templates/NOTES.txt
+++ b/charts/redpanda/templates/NOTES.txt
@@ -37,11 +37,7 @@ Any rpk command that's given to the user in in this file must be defined in _exa
 {{- $anySASL := (include "sasl-enabled" . | fromJson).bool }}
 {{- $rpk := deepCopy . }}
 
-{{- $_ := set $rpk "rpk" (
-  printf "kubectl -n %s exec -ti %s-0 -c redpanda -- rpk"
-    .Release.Namespace
-    (include "redpanda.fullname" .))
-}}
+{{- $_ := set $rpk "rpk" "rpk" }}
 
 Congratulations on installing {{ .Chart.Name }}!
 
@@ -56,18 +52,42 @@ If you are using the load balancer service with a cloud provider, the services w
   {{ printf "helm upgrade %s redpanda/redpanda -n %s --set $(kubectl get svc -n %s -o jsonpath='{\"external.addresses={\"}{ range .items[*]}{.status.loadBalancer.ingress[0].ip }{.status.loadBalancer.ingress[0].hostname}{\",\"}{ end }{\"}\\n\"}')" (include "redpanda.name" .) .Release.Namespace .Release.Namespace }}
 {{- end }}
 
+Set up rpk for access to your external listeners:
+{{- $profile := keys .Values.listeners.kafka.external | first -}}
+{{ if (include "tls-enabled" . | fromJson).bool }}
+  {{- $external := dig "tls" "cert" .Values.listeners.kafka.tls.cert (get .Values.listeners.kafka.external $profile )}}
+  kubectl get secret -n {{ .Release.Namespace }} {{ include "redpanda.fullname" . }}-{{ $external }}-cert -o go-template='{{ "{{" }} index .data "ca.crt" | base64decode }}' > ca.crt
+  {{- if or .Values.listeners.kafka.tls.requireClientAuth .Values.listeners.admin.tls.requireClientAuth }}
+  kubectl get secret -n {{ .Release.Namespace }} {{ include "redpanda.fullname" . }}-client -o go-template='{{ "{{" }} index .data "tls.crt" | base64decode }}' > tls.crt
+  kubectl get secret -n {{ .Release.Namespace }} {{ include "redpanda.fullname" . }}-client -o go-template='{{ "{{" }} index .data "tls.key" | base64decode }}' > tls.key
+  {{- end }}
+{{- end }}
+  rpk profile create --from-profile <(kubectl get configmap -n {{ .Release.Namespace }} {{ include "redpanda.fullname" . }}-rpk -o go-template='{{ "{{" }} .data.profile }}') {{ $profile }}
+
+Set up dns to look up the pods on their Kubernetes Nodes. You can use this query to get the list of short-names to IP addresses. Add your external domain to the hostnames and you could test by adding these to your /etc/hosts:
+
+  kubectl get pod -n {{ .Release.Namespace }} -o custom-columns=node:.status.hostIP,name:.metadata.name --no-headers -l app.kubernetes.io/name=redpanda,app.kubernetes.io/component=redpanda-statefulset
+
+{{- if and $anySASL }}
+
+Set the credentials in the environment:
+
+  kubectl -n {{ .Release.Namespace }} get secret {{ .Values.auth.sasl.secretRef }} -o go-template="{{ "{{" }} range .data }}{{ "{{" }} . | base64decode }}{{ "{{" }} end }}" | IFS=: read -r {{ include "rpk-sasl-environment-variables" . }}
+  export {{ include "rpk-sasl-environment-variables" . }}
+
+{{- end }}
+
 Try some sample commands:
 
 {{- if and $anySASL }}
-{{- $_ := set $rpk "dummySasl" true }}
-
 Create a user:
 
   {{ include "rpk-acl-user-create" $rpk }}
 
 Give the user permissions:
 
-    {{ include "rpk-acl-create" $rpk }}
+  {{ include "rpk-acl-create" $rpk }}
+
 {{- end }}
 
 Get the api status:
--- a/charts/redpanda/templates/_configmap.tpl
+++ b/charts/redpanda/templates/_configmap.tpl
@@ -26,9 +26,9 @@ limitations under the License.
   It's impossible to do a rolling upgrade from not-tls-enabled rpc to tls-enabled rpc.
 */ -}}
 {{- $check := list
-    (include "redpanda-atleast-23-1-2" .|fromJson).bool
-    (include "redpanda-22-3-atleast-22-3-13" .|fromJson).bool
-    (include "redpanda-22-2-atleast-22-2-10" .|fromJson).bool
+  (include "redpanda-atleast-23-1-2" .|fromJson).bool
+  (include "redpanda-22-3-atleast-22-3-13" .|fromJson).bool
+  (include "redpanda-22-2-atleast-22-2-10" .|fromJson).bool
 -}}
 {{- $wantedRPCTLS := (include "rpc-tls-enabled" . | fromJson).bool -}}
 {{- if and (not (mustHas true $check)) $wantedRPCTLS -}}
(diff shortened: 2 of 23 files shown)


RafalKorepta:
```
2023-09-26T00:02:23.3555120Z >>> kubectl --request-timeout=30s logs redpanda-3rrmbectg8-test-kafka-produce-consume --namespace redpanda-3rrmbectg8 --container redpanda
2023-09-26T00:02:23.4348794Z unable to create topics [produce.consume.test.redpanda-3rrmbectg8-test-kafka-produce-consume]: broker closed the connection immediately after a dial, which happens if the client is using TLS when the broker is not expecting it: is TLS misconfigured on the client or the broker?
2023-09-26T00:02:23.4350581Z unable to create topics [produce.consume.test.redpanda-3rrmbectg8-test-kafka-produce-consume]: broker closed the connection immediately after a dial, which happens if the client is using TLS when the broker is not expecting it: is TLS misconfigured on the client or the broker?
2023-09-26T00:02:23.4351873Z unable to create topics [produce.consume.test.redpanda-3rrmbectg8-test-kafka-produce-consume]: broker closed the connection immediately after a dial, which happens if the client is using TLS when the broker is not expecting it: is TLS misconfigured on the client or the broker?
2023-09-26T00:02:23.4353120Z unable to create topics [produce.consume.test.redpanda-3rrmbectg8-test-kafka-produce-consume]: broker closed the connection immediately after a dial, which happens if the client is using TLS when the broker is not expecting it: is TLS misconfigured on the client or the broker?
2023-09-26T00:02:23.4354381Z unable to create topics [produce.consume.test.redpanda-3rrmbectg8-test-kafka-produce-consume]: broker closed the connection immediately after a dial, which happens if the client is using TLS when the broker is not expecting it: is TLS misconfigured on the client or the broker?
2023-09-26T00:02:23.4355928Z unable to create topics [produce.consume.test.redpanda-3rrmbectg8-test-kafka-produce-consume]: broker closed the connection immediately after a dial, which happens if the client is using TLS when the broker is not expecting it: is TLS misconfigured on the client or the broker?
2023-09-26T00:02:23.4357154Z unable to create topics [produce.consume.test.redpanda-3rrmbectg8-test-kafka-produce-consume]: broker closed the connection immediately after a dial, which happens if the client is using TLS when the broker is not expecting it: is TLS misconfigured on the client or the broker?
2023-09-26T00:02:23.4358380Z unable to create topics [produce.consume.test.redpanda-3rrmbectg8-test-kafka-produce-consume]: broker closed the connection immediately after a dial, which happens if the client is using TLS when the broker is not expecting it: is TLS misconfigured on the client or the broker?
2023-09-26T00:02:23.4359611Z unable to create topics [produce.consume.test.redpanda-3rrmbectg8-test-kafka-produce-consume]: broker closed the connection immediately after a dial, which happens if the client is using TLS when the broker is not expecting it: is TLS misconfigured on the client or the broker?
```

It seems like TLS is not propagated to rpk.yaml configuration file.

charts/redpanda/templates/tests/test-connector-via-console.yaml
```diff
@@ -57,32 +57,38 @@ spec:
...
+          RPK_SASL_MECHANISM=${RPK_SASL_MECHANISM:-{{ .Values.auth.sasl.mechanism | upper }}}
+          {{- else }}
+          REDPANDA_SASL_MECHANISM=${REDPANDA_SASL_MECHANISM:-{{ .Values.auth.sasl.mechanism | upper }}}
+          RPK_USER="${REDPANDA_SASL_USERNAME}"
+          RPK_PASS="${REDPANDA_SASL_PASSWORD}"
+          RPK_SASL_MECHANISM="${REDPANDA_SASL_MECHANISM}"
+          {{- end }}
+          export {{ include "rpk-sasl-environment-variables" . }}
+
+          JAAS_CONFIG_SOURCE="\"source.cluster.sasl.jaas.config\": \"org.apache.kafka.common.security.scram.ScramLoginModule required username=\\\\"\"${RPK_USER}\\\\"\" password=\\\\"\"${RPK_PASS}\\\\"\";\","
```
RafalKorepta, line 76: Why the user was changed from console to rpk?

charts/redpanda/templates/tests/test-connector-via-console.yaml
```diff
@@ -122,61 +128,37 @@ spec:
...
-              echo "Result unsuccessful"
-              sleep 1
-            fi
-          done
-
-          if [[ $result -ne 0 ]]
+          URL=http://{{ include "console.fullname" $consoleValues }}:{{ include "console.containerPort" $consoleValues }}/api/kafka-connect/clusters/connectors/connectors
+          {{/* outputting to /dev/null because the output contains the user password */}}
+          echo "Creating mm2 connector"
+          if curl -svm3 --fail --retry 120 --retry-max-time 120 --retry-all-errors -H 'Content-Type: application/json' -o /dev/null "${URL}" -d @/tmp/mm2-conf.json
```
RafalKorepta, line 141: Please don't use this short hand as it is incredible hard to debug failures. Curl will eat and not inform how each request failed with what error and what HTTP code.

charts/redpanda/templates/tests/test-rack-awareness.yaml
```diff
@@ -32,7 +30,7 @@ spec:
   securityContext: {{ include "pod-security-context" . | nindent 4 }}
   {{- with .Values.imagePullSecrets }}
   imagePullSecrets: {{- toYaml . | nindent 4 }}
-  {{- end }}
+{{- end }}
```
RafalKorepta, line 33: NIT: align indentation with the `with` clause.

charts/redpanda/templates/_configmap.tpl
```diff
@@ -403,61 +403,144 @@ limitations under the License.
...
+      key_file: {{ printf "/etc/tls/certs/%s-client/tls.key" (include "redpanda.fullname" .) }}
+    {{- end }}
+  {{- end }}
 {{- end -}}
 
 {{- define "configmap-server-list" -}}
-{{- $root := . }}
-{{- range (include "seed-server-list" $root | mustFromJson) }}
+  {{- range (include "seed-server-list" . | mustFromJson) }}
 - host:
```
RafalKorepta, line 485: Is this -host list indentation correct? I believe it at the `rpk:` level which is in my opinion wrong place.

charts/redpanda/templates/_helpers.tpl
```diff
@@ -760,9 +656,13 @@ return licenseSecretRef.key checks deprecated values entry if current values emp
       {{- $r :=  set $ "tempCert" ( dict "name" $name "cert" $cert ) }}
 - name: redpanda-{{ $name }}-cert
   secret:
-    secretName: {{ template "cert-secret-name" $r }}
+    secretName: {{ template "redpanda.fullname" $ }}-{{ $name }}-cert
```
RafalKorepta, line 659: Why the template for the secret name is not used here?

charts/redpanda/templates/NOTES.txt
```diff
@@ -56,18 +52,43 @@ If you are using the load balancer service with a cloud provider, the services w
...
 Create a user:
 
   {{ include "rpk-acl-user-create" $rpk }}
 
 Give the user permissions:
 
-    {{ include "rpk-acl-create" $rpk }}
+  {{ include "rpk-acl-create" $rpk }}
+
+
```
RafalKorepta, line 91: NIT: ✂️

charts/redpanda/templates/tests/test-connector-via-console.yaml
alejandroEsc: is it though? doesnt the verbosity aid in trouble shooting? I guess do we have an example on the difficulty for debugging?

charts/redpanda/templates/tests/test-connector-via-console.yaml
joejulian: It wasn't console. I changed the variables for consistency and so those same variables can be used for rpk.

charts/redpanda/templates/_helpers.tpl
```diff
@@ -99,10 +99,7 @@ Use AppVersion if image.tag is not set
 {{- $tag -}}
 {{- end -}}
 
-{{/*
-Generate configuration needed for rpk
-*/}}
-
+{{/* Generate internal fqdn */}}
```
alejandroEsc, line 102: thank you for this comment.

alejandroEsc (approved):
Nothing obviously breaking from what I can tell.

charts/redpanda/templates/tests/test-connector-via-console.yaml
joejulian: It's verbose, so all the retries will be printed. Besides, if it fails until it succeeds, we don't care. If it fails until it fails, that last failure is all we need to know anyway. That'll tell us why it failed.

RafalKorepta:
https://github.com/redpanda-data/helm-charts/actions/runs/6315915913/job/17149494209?pr=741#step:20:8661

```
  rpk topic delete test-topic-506 source.test-topic-506 mm2-offset-syncs.test-only-redpanda.internal
  TOPIC                                         STATUS
  mm2-offset-syncs.test-only-redpanda.internal  UNKNOWN_TOPIC_OR_PARTITION: This server does not host this topic-partition.
  source.test-topic-506                         OK
  test-topic-506                                OK
```

That test seems to not working as rpk should return none zero code ...

charts/redpanda/templates/_helpers.tpl
joejulian: I don't know.. probably got lost in the shuffle. Fixed.

joejulian:
> https://github.com/redpanda-data/helm-charts/actions/runs/6315915913/job/17149494209?pr=741#step:20:8661
> 
> ```
>   rpk topic delete test-topic-506 source.test-topic-506 mm2-offset-syncs.test-only-redpanda.internal
>   TOPIC                                         STATUS
>   mm2-offset-syncs.test-only-redpanda.internal  UNKNOWN_TOPIC_OR_PARTITION: This server does not host this topic-partition.
>   source.test-topic-506                         OK
>   test-topic-506                                OK
> ```
> 
> That test seems to not working as rpk should return none zero code ...

I'll open an issue for rpk.

The test did succeed, it's just the cleanup that's failing.

joejulian:
> > https://github.com/redpanda-data/helm-charts/actions/runs/6315915913/job/17149494209?pr=741#step:20:8661
> > ```
> >   rpk topic delete test-topic-506 source.test-topic-506 mm2-offset-syncs.test-only-redpanda.internal
> >   TOPIC                                         STATUS
> >   mm2-offset-syncs.test-only-redpanda.internal  UNKNOWN_TOPIC_OR_PARTITION: This server does not host this topic-partition.
> >   source.test-topic-506                         OK
> >   test-topic-506                                OK
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > That test seems to not working as rpk should return none zero code ...
> 
> I'll open an issue for rpk.
> 
> The test did succeed, it's just the cleanup that's failing.

Looking deeper, I think this is correct. It succeeded because the desired effect is that the non-existant topic be deleted. If it doesn't exist then if deletion is idempotent, it's a success. The topic is, indeed, deleted.

```+ rpk topic list
  NAME                          PARTITIONS  REPLICAS
  _internal_connectors_configs  1           1
  _internal_connectors_offsets  25          1
  _internal_connectors_status   5           1
  _schemas                      1           1
  source.test-topic-506         1           1
  test-topic-506                1           1
  + rpk topic delete test-topic-506 source.test-topic-506 mm2-offset-syncs.test-only-redpanda.internal
  TOPIC                                         STATUS
  mm2-offset-syncs.test-only-redpanda.internal  UNKNOWN_TOPIC_OR_PARTITION: This server does not host this topic-partition.
  source.test-topic-506                         OK
  test-topic-506                                OK
  ```

File: charts/redpanda/templates/tests/test-connector-via-console.yaml (hunk 2 above)
Comment (RafalKorepta) [reply charts/redpanda/templates/tests/test-connector-via-console.yaml:141]: > I guess do we have an example on the difficulty for debugging?

https://github.com/redpanda-data/helm-charts/pull/567/commits/cea5b4ddf09076912827a0092abb8601a32a7bd3
https://github.com/redpanda-data/helm-charts/pull/567

based on https://redpandadata.slack.com/archives/C04K2N6S2A3/p1687526434321209

> doesnt the verbosity aid in trouble shooting?

If you would like to trace what happen and what HTTP codes where returned you need verbosity.

> It's verbose

No, as curl has `-svm3`. The `v` stands to verbose, but `s` is for silent which will take precedence. 

```
Silent or quiet mode. Do not show progress meter or error messages. Makes Curl mute. It still outputs the data you ask for, potentially even to the terminal/stdout unless you redirect it.

Use [-S, --show-error](https://curl.se/docs/manpage.html#-S) in addition to this option to disable progress meter but still show error messages.
```
https://curl.se/docs/manpage.html#-s

If we would use upper case S maybe all error would be returned.

When we will hit the error I will come back to this message as for me problems in CI are not easily reproducible.

RafalKorepta:
> Looking deeper, I think this is correct. It succeeded because the desired effect is that the non-existant topic be deleted. If it doesn't exist then if deletion is idempotent, it's a success. The topic is, indeed, deleted.

When I was looking at the managed connectors that topic should be created by mm2 connector. I understand that rpk did the right thing.

joejulian:
> > Looking deeper, I think this is correct. It succeeded because the desired effect is that the non-existant topic be deleted. If it doesn't exist then if deletion is idempotent, it's a success. The topic is, indeed, deleted.
> 
> When I was looking at the managed connectors that topic should be created by mm2 connector. I understand that rpk did the right thing.

Maybe it cleaned up the topic itself when the config was deleted?

alejandroEsc:
tested this locally, any further issues can be addressed, afterwards, this is a very clean experience and users will love it.

PR #742, opened 2023-09-20
[redpanda] Allow set memory without SI suffix and Redpanda reserve memory can be 0 

RafalKorepta:
In Redpanda operator project the following comment was made:

Setting both `--memory` and `--reserve-memory` is undocumented, and
shouldn't be used. However, there is insufficient information to
correctly calculate what `--reserve-memory` should be. By setting
just `--memory`, seastar will reserve at least 1.5Gi for the OS,
which doesn't make sense in a container.

REF
https://github.com/redpanda-data/redpanda/commit/25f1055573c34320873c4296c48933db4796b99d
https://github.com/redpanda-data/redpanda/pull/3831

Test is added to Github Action Workflow definition.

P.S. Differentiate similar failures when memory is used.

Fixes https://github.com/redpanda-data/redpanda/issues/13346

### DIFF
--- a/charts/redpanda/templates/_helpers.tpl
+++ b/charts/redpanda/templates/_helpers.tpl
@@ -306,9 +306,6 @@ Generate configuration needed for rpk
       {{- $result = 1000 -}}
     {{- end -}}
   {{- end -}}
-  {{- if eq $result 0 -}}
-    {{- "unable to get memory value" | fail -}}
-  {{- end -}}
   {{- $result -}}
 {{- end -}}
 
--- a/charts/redpanda/templates/_helpers.tpl
+++ b/charts/redpanda/templates/_helpers.tpl
@@ -203,6 +203,7 @@ Generate configuration needed for rpk
   {{/*
   This template converts the incoming SI value to whole number bytes.
   Input can be: b | B | k | K | m | M | g | G | Ki | Mi | Gi
+  Or number without suffix
   */}}
   {{- $si := . -}}
   {{- $bytes := 0 -}}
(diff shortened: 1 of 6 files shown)


charts/redpanda/templates/_helpers.tpl
```diff
@@ -203,6 +203,7 @@ Generate configuration needed for rpk
   {{/*
   This template converts the incoming SI value to whole number bytes.
   Input can be: b | B | k | K | m | M | g | G | Ki | Mi | Gi
+  Or number without siffix
```
joejulian, line 206:
```suggestion
  Or number without suffix
```

charts/redpanda/templates/_helpers.tpl
```diff
@@ -226,8 +227,10 @@ Generate configuration needed for rpk
   {{- else if hasSuffix "Gi" $si -}}
     {{- $raw := $si | trimSuffix "Gi" | float64 -}}
     {{- $bytes = mulf $raw (mul 1024 1024 1024) | floor -}}
+  {{- else if or (hasSuffix "0" $si) (hasSuffix "1" $si) (hasSuffix "2" $si) (hasSuffix "3" $si) (hasSuffix "4" $si) (hasSuffix "5" $si) (hasSuffix "6" $si) (hasSuffix "7" $si) (hasSuffix "8" $si) (hasSuffix "9" $si) -}}
```
joejulian, line 230:
```suggestion
  {{- else if (mustRegexMatch "^[0-9]+$" $si) -}}
```

charts/redpanda/templates/_helpers.tpl
```diff
@@ -248,7 +251,7 @@ Generate configuration needed for rpk
     {{- $result = .Values.resources.memory.container.max | include "redpanda-memoryToMi" -}}
   {{- end -}}
   {{- if eq $result "" -}}
-    {{- "unable to get memory value" | fail -}}
+    {{- "unable to get memory value form container" | fail -}}
```
joejulian, line 254:
```suggestion
    {{- "unable to get memory value from container" | fail -}}
```

charts/redpanda/templates/_helpers.tpl
RafalKorepta: thx

charts/redpanda/templates/_helpers.tpl
RafalKorepta: thx

charts/redpanda/templates/_helpers.tpl
RafalKorepta: thx

joejulian (approved):

PR #744, opened 2023-09-21
deprecated non-used config{} entry for console

alejandroEsc:
Adding new entries to values file and removing old entry that was never used.

### DIFF
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 5.4.9
+version: 5.4.11
 
 # The app version is the default version of Redpanda to install.
 # ** NOTE for maintainers: please ensure the artifacthub image annotation is updated before merging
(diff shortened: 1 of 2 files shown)


RafalKorepta:
The change is correct, but I personally wouldn't expose console values inside Redpanda values. If we change the place where config is defined inside console chart (currently `console.config: {}`) you need to follow up in Redpanda chart. That repetition is a maintenance work which I would like to avoid if possible. 

My recommendation would be to remove config as it is noop. 

REF
https://helm.sh/docs/chart_template_guide/subcharts_and_globals/#overriding-values-from-a-parent-chart

alejandroEsc:
> The change is correct, but I personally wouldn't expose console values inside Redpanda values. If we change the place where config is defined inside console chart (currently `console.config: {}`) you need to follow up in Redpanda chart. That repetition is a maintenance work which I would like to avoid if possible.
> 
> My recommendation would be to remove config as it is noop.
> 
> REF https://helm.sh/docs/chart_template_guide/subcharts_and_globals/#overriding-values-from-a-parent-chart

I see what you are saying, but the additional `console: {}` will always be valid, if the console specific values file changes we will not have to change anything here. Remember that the change in the values file will be `console.console.config` -> `console.newconsole.config`

RafalKorepta (commented):
Based on the @joejulian it would be wise to move console chart value form `console.config` to `config` (one level up). Given that I'm not sure if this PR is valid.

alejandroEsc:
Closing this for now, changes will occur to the console chart first.

PR #750, opened 2023-09-25
[redpanda] Add tlsConfig for ServiceMonitor

guicholeo:
## Motivation

ServiceMonitor does not have configuration to set TLS for scraping with Prometheus.

## Approach
Enable tlsConfig in ServiceMonitor as-well keep default configuration if the cluster uses TLS. Also adds `enableHttp2` which by default is `true`.

### DIFF
--- a/charts/redpanda/templates/servicemonitor.yaml
+++ b/charts/redpanda/templates/servicemonitor.yaml
@@ -34,10 +34,17 @@ spec:
   - interval: {{ .Values.monitoring.scrapeInterval }}
     path: /public_metrics
     targetPort: admin
-  {{- if .Values.tls.enabled }}
+    {{- if dig "enableHttp2" "" .Values.monitoring }}
+    enableHttp2: .Values.monitoring.enableHttp2
+    {{- end }}
+  {{- if or .Values.tls.enabled .Values.monitoring.tlsConfig }}
     scheme: https
     tlsConfig:
+      {{- if  dig "tlsConfig" dict .Values.monitoring }}
+      {{- .Values.monitoring.tlsConfig | toYaml | nindent 6 }}
+      {{- else }}
       insecureSkipVerify: true
+      {{- end}}
   {{- end }}
   selector:
     matchLabels:
--- a/charts/redpanda/values.yaml
+++ b/charts/redpanda/values.yaml
@@ -275,6 +275,13 @@ monitoring:
   enabled: false
   scrapeInterval: 30s
   labels: {}
+  # Enables http2 for scraping metrics for prometheus. Used when Istio's mTLS is enabled and using tlsConfig.
+  # enableHttp2: true
+  tlsConfig: {}
+    # caFile: /etc/prom-certs/root-cert.pem
+    # certFile: /etc/prom-certs/cert-chain.pem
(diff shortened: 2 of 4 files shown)


alejandroEsc:
Will be reviewing today, the constant updating is due to the multiple changes going in.

charts/redpanda/templates/servicemonitor.yaml
```diff
@@ -34,10 +34,15 @@ spec:
   - interval: {{ .Values.monitoring.scrapeInterval }}
     path: /public_metrics
     targetPort: admin
-  {{- if .Values.tls.enabled }}
+    enableHttp2: {{ .Values.monitoring.enableHttp2 }}
+  {{- if or .Values.tls.enabled .Values.monitoring.tlsConfig }}
     scheme: https
     tlsConfig:
+      {{- if .Values.monitoring.tlsConfig }}
```
alejandroEsc, line 41:
```suggestion
      {{- if  dig "tlsConfig" dict .Values.monitoring }}
```

charts/redpanda/values.yaml
```diff
@@ -271,6 +271,8 @@ monitoring:
   enabled: false
   scrapeInterval: 30s
   labels: {}
+  enableHttp2: true
```
alejandroEsc, line 274: please update the values.schema.json file with these new entries.

alejandroEsc, line 274: why enable this true by default? Should we opt in or is the default value true in the resource spec?

charts/redpanda/values.yaml
```diff
@@ -271,6 +271,8 @@ monitoring:
...
+  tlsConfig: {}
```
alejandroEsc, line 275: Would you mind adding a commented sample of the tlsConfig?

alejandroEsc (commented):
A few asks, but looking good, thank you! May need to bump version of the chart, it is annoying but we are making quick progress.

charts/redpanda/values.yaml
guicholeo: I can remove it, there is no default value https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#endpoint will using `with` suffice to see if its present? 

```
{{- with .Values.monitoring.enableHttp2 }}
enableHttp2: {{ . }}
{{- end }}
```

guicholeo: Is there a plugin which generates the schema? I tried one but it changes most of the current file

alejandroEsc: No, not that I know of, You can manually add it, i can help if you need it.

alejandroEsc: The question is about the values file and the default value you set for it. Your suggestion makes sense, we can also try

```
{{- if dig "enableHttp2" "" .Values.monitoring }}
enableHttp2: .Values.monitoring.enableHttp2
{{- end }}
```

 and let's then comment it out on the values file, but with a note above about how to use it. Thank you,

RafalKorepta (approved):

RafalKorepta (approved):
I rebased and squashed all commits.

LGTM.

Thank you for contribution.

PR #756, opened 2023-09-27
refactor tiered storage configuration values

### LINKED ISSUE #755: redpanda: cloud cache should not use a separate mount unless specifically configured
Unless configured otherwise, the cloud cache directory should not have a mount and, instead, the data directory should be used.

This change, while not a breaking change, will have an impact as the old cloud-cache data will need to be re-hydrated. We will need to manage user expectations.

----

_This action was created from [Incident 370](https://app.incident.io/redpandadata/incidents/370) by joe, using [incident.io](https://app.incident.io) 🔥_

joejulian:
move all the tiered storage configuration under `storage.tiered`. Users were expressing confusion that some tiered storage configuration lied outside of the tieredConfig key.

This also now allows a user to choose "none" for the tiered storage mount type, allowing the data to be written to the redpanda data directory.

Fixes #755

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -48,8 +48,8 @@ jobs:
       matrix:
         version:
           - ""
-          - v22.3.17
-          - v22.2.11
+          - v23.1.18
+          - v22.3.23
         testvaluespattern:
           - '0[1-3]*'
           - '0[4-6]*'
@@ -57,7 +57,7 @@ jobs:
           - '1[0-2]*'
           - '1[3-5]*'
           - '1[6-7]*'
-      fail-fast: true
+      fail-fast: false
     runs-on: ubuntu-22.04
     steps:
       - name: Checkout
--- a/charts/redpanda/templates/statefulset.yaml
+++ b/charts/redpanda/templates/statefulset.yaml
@@ -102,16 +102,18 @@ spec:
           resources: {{- toYaml .Values.statefulset.initContainers.setDataDirOwnership.resources | nindent 12 }}
   {{- end }}
 {{- end }}
-{{- if and (include "is-licensed" . | fromJson).bool .Values.storage.tieredConfig.cloud_storage_enabled }}
+{{- if and (include "is-licensed" . | fromJson).bool (include "storage-tiered-config" .|fromJson).cloud_storage_enabled }}
         - name: set-tiered-storage-cache-dir-ownership
           image: {{ .Values.statefulset.initContainerImage.repository }}:{{ .Values.statefulset.initContainerImage.tag }}
-          command: ["/bin/sh", "-c", 'chown {{ $uid }}:{{ $gid }} -R {{ template "tieredStorage.cacheDirectory" . }}']
+          command: ["/bin/sh", "-c", 'chown {{ $uid }}:{{ $gid }} -R {{ include "tieredStorage.cacheDirectory" . }}']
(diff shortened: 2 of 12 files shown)


.github/workflows/nightly.yaml
```diff
@@ -48,16 +48,16 @@ jobs:
       matrix:
         version:
           - ""
-          - v22.3.17
-          - v22.2.11
```
RafalKorepta, line 52: Did we deprecated 22.2.X?

charts/redpanda/templates/statefulset.yaml
```diff
@@ -336,17 +338,20 @@ spec:
...
+        {{- if ne $tieredType "none" }}
         - name: tiered-storage-dir
-        {{- if .Values.storage.tieredStoragePersistentVolume.enabled }}
+          {{- if eq $tieredType "persistentVolume" }}
           persistentVolumeClaim:
             claimName: tiered-storage-dir
-        {{- else if .Values.storage.tieredStorageHostPath }}
+          {{- else if eq $tieredType "hostPath" }}
           hostPath:
-            path: {{ .Values.storage.tieredStorageHostPath | quote }}
```
RafalKorepta, line 346: This change does not guarantee that host path is not empty. Previous implementation didn't pass empty string to the host path.

charts/redpanda/values.schema.json
```diff
@@ -349,16 +349,142 @@
...
+              "type": "string",
+              "pattern": "^(none|hostPath|emptyDir|persistentVolume)$"
+            },
+            "hostPath": {
+              "type": "string"
+            },
+            "persistentVolume": {
+              "type": "object",
+              "required": [
+                "storageClass",
```
RafalKorepta, line 375: storageClass is not required in persistent volume claim definition.

charts/redpanda/values.schema.json
```diff
@@ -349,16 +349,142 @@
...
+                "labels",
+                "annotations"
+              ],
```
RafalKorepta, line 378: None of this fields are required in persistent volume claim definition.

charts/redpanda/values.schema.json
```diff
@@ -349,16 +349,142 @@
...
+              }
+            },
+            "config":{
+              "type": "object",
+              "required": [
+                "cloud_storage_enabled",
+                "cloud_storage_region",
+                "cloud_storage_bucket"
+              ],
+              "properties": {
```
RafalKorepta, line 398: Those fields are not guaranteed as per https://github.com/redpanda-data/helm-charts/pull/278#discussion_r1072782702

RafalKorepta (approved):
None blocking issues where found.

LGTM

.github/workflows/nightly.yaml
joejulian: Yes and no. I haven't removed any of the version checking yet, but it's not supported anymore.

charts/redpanda/values.schema.json
joejulian: The keys are required. This allows us to reference them as `.Values.storage.tiered.persistentVolume.storageClass` and not worry about dereferencing nulls instead of having to use dig.

charts/redpanda/templates/statefulset.yaml
joejulian: True. I'll add some validation to the `storage-tiered-mountType` template.

PR #760, opened 2023-09-28
validate tiered storage is configured

joejulian:
Addresses https://github.com/redpanda-data/helm-charts/pull/756/files#r1339606803
> @RafalKorepta: This change does not guarantee that host path is not empty. Previous implementation didn't pass empty string to the host path.


RafalKorepta (approved):
LGTM

PR #763, opened 2023-10-01
k8s: Allow ingress resource reconciliation in Redpanda controller

RafalKorepta:
A follow up on https://github.com/redpanda-data/redpanda/pull/13850

### DIFF
--- a/charts/operator/templates/role.yaml
+++ b/charts/operator/templates/role.yaml
@@ -275,6 +275,18 @@ rules:
     - update
     - watch
 - apiGroups:
+    - networking.k8s.io
+  resources:
+    - ingresses
+  verbs:
+    - create
+    - delete
+    - get
+    - list
+    - patch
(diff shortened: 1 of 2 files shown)


charts/operator/templates/role.yaml
```diff
@@ -274,6 +274,18 @@ rules:
     - patch
     - update
     - watch
+- apiGroups:
```
alejandroEsc, line 277: Can you add these to our operator repo in the redpanda controller file, if you havent already. We also discussed that we will continue to hit these missing crds continually and maintaining this is going to be difficult.

alejandroEsc (approved):
one comment

charts/operator/templates/role.yaml
alejandroEsc: ahh next PR, nm then, thank you.

PR #765, opened 2023-10-02
console: Allow setting NodePort

RafalKorepta:


charts/console/templates/service.yaml
```diff
@@ -30,4 +30,5 @@ spec:
       targetPort: {{ include "console.containerPort" . }}
       protocol: TCP
       name: http
+      nodePort: {{ .Values.service.nodePort }}
```
alejandroEsc, line 33: Should we gate these if the service.type is nodeport?

alejandroEsc: Also we have service files depending on type in redpanda, should we follow the same approach here?

alejandroEsc (commented):
a few questions.

charts/console/templates/service.yaml
RafalKorepta: Sure I will update this PR

RafalKorepta: > Also we have service files depending on type in redpanda, should we follow the same approach here?

We have only NOTES.txt that depend on type of the service. This is console chart, so I'm not sure how Redpanda chart applies here?

alejandroEsc: It doesn't, but in that chart we have two different service files, one per each type, just wondering if you think we need to do this here.

RafalKorepta: > It doesn't, but in that chart we have two different service files, one per each type, just wondering if you think we need to do this here.

Can you send link to Github where the service file is defined?

alejandroEsc: https://github.com/redpanda-data/helm-charts/blob/main/charts/redpanda/templates/service.loadbalancer.yaml and
https://github.com/redpanda-data/helm-charts/blob/main/charts/redpanda/templates/services.nodeport.yaml

alejandroEsc (approved):
I wasnt gating this PR. Just asking a question, lgtm.

charts/console/templates/service.yaml
RafalKorepta: We agree that changes to console can not be guarded by higher level charts. If console would depend on Redpanda then we could create some helper function.

RafalKorepta:
I opened new PR as CLA boot doesn't report status and I'm unable to merge this PR

PR #766, opened 2023-10-02
feat: when rp chart has rbac.enabled, we need permissions to deploy. 

alejandroEsc:
fixes https://github.com/redpanda-data/redpanda/issues/13825 

Cluster roles needed to allow for operator to deploy redpanda when rbac.enabled is set to Redpanda chart.

### DIFF
--- a/charts/operator/Chart.yaml
+++ b/charts/operator/Chart.yaml
@@ -6,7 +6,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 0.3.22
+version: 0.3.23
 
 # This is the default version of the operator being deployed.
 # ** NOTE for maintainers: please ensure the artifacthub image annotation is updated before merging
--- a/charts/operator/Chart.yaml
+++ b/charts/operator/Chart.yaml
@@ -10,7 +10,7 @@ version: 0.3.22
 
 # This is the default version of the operator being deployed.
 # ** NOTE for maintainers: please ensure the artifacthub image annotation is updated before merging
-appVersion: v23.2.7
+appVersion: v23.2.11
 
 sources:
   - https://github.com/redpanda-data/helm-charts
@@ -34,9 +34,9 @@ annotations:
       url: https://helm.sh/docs/intro/install/
   artifacthub.io/images: |
     - name: redpanda-operator
-      image: docker.redpanda.com/redpandadata/redpanda-operator:v23.2.5
+      image: docker.redpanda.com/redpandadata/redpanda-operator:v23.2.11
     - name: redpanda
-      image: docker.redpanda.com/redpandadata/redpanda:v23.2.5
+      image: docker.redpanda.com/redpandadata/redpanda:v23.2.11
     - name: kube-rbac-proxy
       image: gcr.io/kubebuilder/kube-rbac-proxy:v0.8.0
   artifacthub.io/crds: |
--- a/charts/operator/templates/clusterrole.yaml
+++ b/charts/operator/templates/clusterrole.yaml
@@ -9,7 +9,8 @@ the Business Source License, use of this software will be governed
 by the Apache License, Version 2.0
 */}}
 
-{{- if and .Values.rbac.create ( eq .Values.scope "Cluster" ) -}}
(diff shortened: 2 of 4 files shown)


charts/operator/templates/clusterrole.yaml
```diff
@@ -286,7 +287,7 @@ rules:
     - patch
     - update
 {{- end -}}
-{{- if and .Values.rbac.create (and ( eq .Values.scope "Namespace" ) .Values.rbac.createAdditionalControllerCRs ) -}}
```
RafalKorepta, line 289: `createAdditionalControllerCRs` is removed because Redpanda chart required cluster role?

charts/operator/templates/clusterrole.yaml
```diff
@@ -295,6 +296,53 @@ metadata:
...
+    - persistentvolumeclaims
+    - pods
+    - pods/log
+    - replicationcontrollers
+    - resourcequotas
+    - serviceaccounts
+    - services
+  verbs:
+    - get
+    - list
```
RafalKorepta, line 328: I'm against this change as namespaced operator will not be a thing.

charts/operator/templates/clusterrole.yaml
```diff
@@ -295,6 +296,53 @@ metadata:
...
+- apiGroups:
+    - ""
+  resources:
+    - configmaps
+    - endpoints
+    - events
+    - limitranges
+    - persistentvolumeclaims
+    - pods
+    - pods/log
```
RafalKorepta, line 321: Why logs subresource?

charts/operator/templates/clusterrole.yaml
```diff
@@ -295,6 +296,53 @@ metadata:
...
+    - list
+    - patch
+    - update
+    - watch
+- apiGroups:
+    - ""
+  resources:
+    - configmaps
+    - endpoints
+    - events
```
RafalKorepta, line 317: Why events?

charts/operator/templates/clusterrole.yaml
```diff
@@ -295,6 +296,53 @@ metadata:
...
+    - delete
+    - list
+    - patch
+    - update
+    - watch
+- apiGroups:
+    - ""
+  resources:
+    - configmaps
+    - endpoints
```
RafalKorepta, line 316: Why endpoints?

charts/operator/templates/clusterrole.yaml
alejandroEsc: No, it was moved down, i added a new clusterrole that is deployed by default. This role is required to create rbac roles that redpanda chart needs. The rbac roles are the same as found in the redpanda/templates/rbac.yaml, that will answer your questions.

charts/operator/templates/clusterrole.yaml
alejandroEsc: Let's Discuss this PR then, The issue needs to be resolved.

alejandroEsc: Best alternative i found after talking to @joejulian was to just gate this option. So added a new values that will deploy this if user really wants to enable rpk-bundle clusterroles to redpandas.

charts/operator/Chart.yaml
```diff
@@ -6,11 +6,11 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 0.3.22
+version: 0.3.23
 
 # This is the default version of the operator being deployed.
 # ** NOTE for maintainers: please ensure the artifacthub image annotation is updated before merging
-appVersion: v23.2.7
+appVersion: v23.2.10
```
RafalKorepta, line 13: NIT: 
```suggestion
appVersion: v23.2.11
```

RafalKorepta (approved):

charts/operator/templates/clusterrole.yaml
```diff
@@ -313,4 +361,4 @@ rules:
     - patch
     - update
     - watch
-{{- end -}}
```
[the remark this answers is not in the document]
alejandroEsc: To avoid things like this:

```
    namespace: default---
```

https://helm.sh/docs/chart_template_guide/control_structures/

RafalKorepta (approved):

PR #767, opened 2023-10-02
chore: re-enable prometheus testing

alejandroEsc:
WIP: trying to see if there is any additional work needed to allow the prometheus test to succeed.


RafalKorepta (approved):

alejandroEsc:
It seems the tests installs, upgrades, then installs again the values file. I will have to check the details, but I wonder if something breaks during the process of testing. I will try this locally to check.

PR #771, opened 2023-10-04
fix: config parsing

### LINKED ISSUE #770: node configuration not correctly parsed
### What happened?

I configured developer_mode in my values and it was not included in the `redpanda.yaml` key in the configmap.

Have developer mode turned on

<details>

```console
# on HEAD
$ helm upgrade --install --namespace infra --create-namespace --set config.node.developer_mode=true --dry-run --debug redpanda charts/redpanda | grep developer_mode
output contains 2x developer_mode (computer and user provided)

# introduced at
$ git checkout 11b82c515c704da1600abb5f589e2fbc3af92fcb
$ helm upgrade --install --namespace infra --create-namespace --set config.node.developer_mode=true --dry-run --debug redpanda charts/redpanda | grep developer_mode
output contains 2x developer_mode (computer and user provided)

# works
$ git checkout 11b82c515c704da1600abb5f589e2fbc3af92fcb~1
$ helm upgrade --install --namespace infra --create-namespace --set config.node.developer_mode=true --dry-run --debug redpanda charts/redpanda | grep developer_mode
output contains 3x developer_mode (computer and user provided)
```

</details>

Introduced: https://github.com/redpanda-data/helm-charts/blame/11b82c515c704da1600abb5f589e2fbc3af92fcb/charts/redpanda/templates/_configmap.tpl#L132

_No response_

### Chart Version(s)

<details>
Introduced **after** redpanda-5.4.0
</details>

### Cloud provider

N/A

zegerius:
fixes a bug introduced at 11b82c515c704da1600abb5f589e2fbc3af92fcb that prevented the configuration set in values.yaml from being parsed. Any other key but `crash_loop_limit` would be removed from the output.

verify and check if other values are still added correctly:

```
$ helm upgrade --install --namespace infra --create-namespace --set config.node.developer_mode=true --set config.node.crash_loop_limit=3 --dry-run --debug redpanda charts/redpanda | grep -C1 developer_mode

history.go:56: [debug] getting history for release redpanda
install.go:193: [debug] Original chart version: ""
install.go:210: [debug] CHART PATH: /Users/aram/projects/redpanda-helm-charts/charts/redpanda

    crash_loop_limit: 3
    developer_mode: true

--
    crash_loop_limit: 3
    developer_mode: true
  pandaproxy_client: {}
--
      crash_loop_limit: 3
      developer_mode: true
```

closes #770

### DIFF
--- a/charts/redpanda/templates/_configmap.tpl
+++ b/charts/redpanda/templates/_configmap.tpl
@@ -130,7 +130,7 @@ redpanda.yaml: |
 {{- end }}
 {{- with dig "node" dict .Values.config }}
   {{- range $key, $element := .}}
-    {{- if and (or (eq (typeOf $element) "bool") $element) (and (eq $key "crash_loop_limit") (include "redpanda-atleast-23-1-1" $root | fromJson).bool) }}
+    {{- if and (or (eq (typeOf $element) "bool") $element) (or (and (eq $key "crash_loop_limit") (include "redpanda-atleast-23-1-1" $root | fromJson).bool) (ne $key "crash_loop_limit")) }}
     {{ $key }}: {{ $element | toYaml }}
     {{- end }}
   {{- end }}
@@ -487,7 +487,9 @@ rpk:
   additional_start_flags:
     - "--smp={{ include "redpanda-smp" . }}"
     - "--memory={{ template "redpanda-memory" . }}M"
+    {{- if not .Values.config.node.developer_mode }}
     - "--reserve-memory={{ template "redpanda-reserve-memory" . }}M"
+    {{- end }}
     - "--default-log-level={{ .Values.logging.logLevel }}"
   {{- with .Values.statefulset.additionalRedpandaCmdFlags -}}
   {{- toYaml . | nindent 4 }}
--- a/charts/redpanda/templates/_configmap.tpl
+++ b/charts/redpanda/templates/_configmap.tpl
@@ -130,8 +130,15 @@ redpanda.yaml: |
 {{- end }}
 {{- with dig "node" dict .Values.config }}
   {{- range $key, $element := .}}
-    {{- if and (or (eq (typeOf $element) "bool") $element) (or (and (eq $key "crash_loop_limit") (include "redpanda-atleast-23-1-1" $root | fromJson).bool) (ne $key "crash_loop_limit")) }}
-    {{ $key }}: {{ $element | toYaml }}
+    {{- $line := dict $key (toYaml $element) }}
+    {{- if and (eq $key "crash_loop_limit") (not (include "redpanda-atleast-23-1-1" $root | fromJson).bool) }}
+      {{- $line = dict }}
+    {{- end }}
+    {{- if not (or (eq (typeOf $element) "bool") $element) }}
+      {{- $line = dict }}
+    {{- end }}
+    {{- with $line }}
+      {{  toYaml . | nindent 4 }}
     {{- end }}
   {{- end }}
 {{- end -}}
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 5.6.0
+version: 5.6.1
 
 # The app version is the default version of Redpanda to install.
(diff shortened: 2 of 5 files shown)


zegerius:
After more testing, I noticed that this exposes another bug in this branch:

```
redpanda WARNING: This is a setup for development purposes only; in this mode your clusters may run unrealistically fast and data can be corrupted any time your computer shuts down uncleanly.

redpanda Error: configuration conflict. Flag '--reserve-memory' is also present in 'rpk.additional_start_flags' in configuration file '/etc/redpanda/redpanda.yaml'. Please remove it and pass '--reserve-memory' directly to `rpk start`
```

If you apply the fix to the commit referenced above, the original issue is resolved.

zegerius:
This has been fixed. `developer_mode` was not taken into account when moving the flags to redpanda.yaml, suspected commit: d4d3bf31e15b992c5b346cb2a47dec5e8b7f7c6a. Seems that `developer_mode: true` and `reserve-memory` are not compatible. Took a look at the CLI but really have to get other things done :)

charts/redpanda/templates/_configmap.tpl
```diff
@@ -130,7 +130,7 @@ redpanda.yaml: |
 {{- end }}
 {{- with dig "node" dict .Values.config }}
   {{- range $key, $element := .}}
-    {{- if and (or (eq (typeOf $element) "bool") $element) (and (eq $key "crash_loop_limit") (include "redpanda-atleast-23-1-1" $root | fromJson).bool) }}
+    {{- if or (eq (typeOf $element) "bool") (and (eq $key "crash_loop_limit") (include "redpanda-atleast-23-1-1" $root | fromJson).bool) }}
```
RafalKorepta, line 133:
```suggestion
    {{- if and (or (eq (typeOf $element) "bool") $element) (or (and (eq $key "crash_loop_limit") (include "redpanda-atleast-23-1-1" $root | fromJson).bool) (ne $key "crash_loop_limit")) }}
```

The first check `or (eq (typeOf $element) "bool") $element` needs to stay as is. Bug was introduced in second part of the condition statement. The key crash_loop_limit can always be defined, but it needs to be pass only if Redpanda semver is at least 23.1.1. The rest of the keys should be pass as is.

RafalKorepta:
Hi @zegerius,

Could you bump semver to `5.6.1`?

https://github.com/redpanda-data/helm-charts/blob/9706db148194cbfdbc12aec82e086190d2ec64de/charts/redpanda/Chart.yaml#L26

charts/redpanda/templates/_configmap.tpl
zegerius: Sharp eye! This syntax remains hard to read. Fixed 👍

RafalKorepta: Indeed 😮‍💨 I'm not fun of go template.

charts/redpanda/templates/_configmap.tpl
```diff
@@ -130,7 +130,7 @@ redpanda.yaml: |
 {{- end }}
 {{- with dig "node" dict .Values.config }}
   {{- range $key, $element := .}}
-    {{- if and (or (eq (typeOf $element) "bool") $element) (and (eq $key "crash_loop_limit") (include "redpanda-atleast-23-1-1" $root | fromJson).bool) }}
+    {{- if and (or (eq (typeOf $element) "bool") $element) (or (and (eq $key "crash_loop_limit") (include "redpanda-atleast-23-1-1" $root | fromJson).bool) (ne $key "crash_loop_limit")) }}
     {{ $key }}: {{ $element | toYaml }}
     {{- end }}
   {{- end }}
```
joejulian, line 136: Another way of doing this that doesn't build those long complex (and error prone) if statements:

```suggestion
  {{- range $key, $element := .}}
    {{- $line := dict $key (toYaml $element) }}
    {{- if and (eq $key "crash_loop_limit") (include "redpanda-atleast-23-1-1" $root | fromJson).bool) }}
      {{- $line = dict }}
    {{- end }}
    {{- if not (or (eq (typeOf $element) "bool") $element) }}
      {{- $line = dict }}
    {{- with $line}
      {{  . | nindent 4 }}
    {{- end }}
  {{- end }}
  ```

joejulian: added a commit with this change

joejulian (approved):

RafalKorepta (approved):

joejulian:
Thanks @zegerius!

PR #774, opened 2023-10-04
Provide duration for CA certificate

njhartwell:
The value is present in config but was ignored, resulting in the root CA certificates having a default duration of 90 days, which is not ideal in most situations.

### DIFF
--- a/charts/redpanda/templates/cert-issuers.yaml
+++ b/charts/redpanda/templates/cert-issuers.yaml
@@ -21,6 +21,7 @@ limitations under the License.
   {{- range $name, $data := $values.tls.certs }}
     {{/* If secretRef is defined, do not create any of these certificates. */}}
     {{- if not (hasKey $data "secretRef") }}
+      {{- $d := $data.duration }}
 
 ---
 {{/* If issuerRef is defined, use the specified issuer for the certs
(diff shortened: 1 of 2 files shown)


charts/redpanda/templates/cert-issuers.yaml
```diff
@@ -65,6 +66,7 @@ metadata:
   {{- . | nindent 4 }}
 {{- end }}
 spec:
+  duration: {{ $d | default "43800h" }}
```
alejandroEsc, line 69: The default is 90d if unset right? Why 5 years?

alejandroEsc (commented):
great contribution, just one question.

charts/redpanda/templates/cert-issuers.yaml
njhartwell: That's what the server cert defaults to, if not overridden. Since this is the CA that signs that cert, having the it expire (long) before the server certs doesn't make a lot of sense, and could lead to so some weird behaviors (e.g. some clients continuing to work and others rejecting it). Also, having a very long duration for the root CA cert seems seems to be common practice and makes sense intuitively -- trusted CA sets are often baked into images, applications, etc. and aren't always easy to manage.

alejandroEsc:
Please bump the redpanda chart version so I can review, thank you

RafalKorepta (approved):

PR #777, opened 2023-10-05
Size can be expressed by integer or string

RafalKorepta:
When tiered storage is defined as `200Gi` then schema is violated and helm release fails.


alejandroEsc (approved):
Lgtm

joejulian (approved):

PR #780, opened 2023-10-05
Tiered storage configuration bytes calculations

RafalKorepta:
Redpanda doesn't allow `cloud_storage_secret_key` to be defined as string.

Output from the `post-upgrade` job:
```
PROPERTY                            PRIOR                NEW
cloud_storage_cache_size  11000000000  11G
cloud_storage_secret_key  [secret]             [redacted]

Validation errors:
 * cloud_storage_cache_size: expected type integer
```


alejandroEsc (approved):
lgtm, but I thought this would be a schema issue.

PR #787, opened 2023-10-11
Revert: Change the lint job to be noob so it is not required

RafalKorepta:


joejulian:
I'm just going to close this and rework the workflows to do the lint correctly.

PR #788, opened 2023-10-12
Split test cases to lower the amount of time test takes in CI

RafalKorepta:
The test case 12, 13 and 15 as each takes more than 15 minutes.


alejandroEsc (approved):
lgtm, let's agree that this will not be permanent but needed for investigations.

PR #792, opened 2023-10-13
Make connectors test failure more verbose

RafalKorepta:
* Manage Connectors now will have debug log level during tests
* When mirror maker 2 test fails bash execution gather much more information about the state of the connectors
* Bump connector image tag/version


alejandroEsc (approved):
Lgtm, nice

Issue #799, opened 2023-10-16
large ints converted to scientific notation


joejulian:
### What happened?

With a values file like:
```
config:
  cluster:
    retention_local_target_ms_default: 21600000
```
the rpk command in post-update job is created like:
```
rpk cluster config set retention_local_target_ms_default 2.16e+07
```

### What did you expect to happen?

The command should be
```
rpk cluster config set retention_local_target_ms_default 21600000
```

### How can we reproduce it (as minimally and precisely as possible)?. Please include values file.

<details>

```console
$ helm get values <redpanda-release-name> -n <redpanda-release-namespace> --all
# paste output here
```

</details>

### Chart Version(s)

<details>

```console
$ helm -n <redpanda-release-namespace> list 
# paste output here
```

</details>

### Cloud provider

<details>

</details>

PR #800, opened 2023-10-16
fix large int produces scientific notation rpk command

joejulian:
Fixes https://github.com/redpanda-data/helm-charts/issues/799

### DIFF
--- a/charts/redpanda/ci/02-one-node-cluster-no-tls-no-sasl-values.yaml
+++ b/charts/redpanda/ci/02-one-node-cluster-no-tls-no-sasl-values.yaml
@@ -30,3 +30,9 @@ connectors:
   enabled: true
   logging:
     level: debug
+
+# tests to ensure this large int isn't converted to scientific notation for the rpk commands
+# in post-upgrade job.
+config:
+  cluster:
+    retention_local_target_ms_default: 21600000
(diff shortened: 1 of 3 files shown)


alejandroEsc (approved):
At which level does this get changed to scientific notation?

charts/redpanda/ci/02-one-node-cluster-no-tls-no-sasl-values.yaml
```diff
@@ -30,3 +30,9 @@ connectors:
   enabled: true
   logging:
     level: debug
+
+# tests to ensure this large int isn't converted to scientific notation for the rpk commands
+# in post-upgrade job.
+config:
+  cluster:
+    retention_local_target_ms_default: 21600000
```
RafalKorepta, line 38: The same apply to tiered storage

https://github.com/redpanda-data/helm-charts/blob/9744a28e9cb9dc8da141a2f2fa0a6af849ea6bb5/charts/redpanda/templates/_configmap.tpl#L290-L303
and
https://github.com/redpanda-data/helm-charts/blob/9744a28e9cb9dc8da141a2f2fa0a6af849ea6bb5/charts/redpanda/templates/_configmap.tpl#L83-L97

joejulian: Please open a new issue.

PR #803, opened 2023-10-17
Unify connectors test suite

RafalKorepta:
Use curl retry mechanisms and remove bash loop, by new curl common parameters. Both test execution will use trap function to report current state.


alejandroEsc (approved):
nice change, lgtm

PR #805, opened 2023-10-17
fix: do not generate client certs if mtls is not used

### LINKED ISSUE #804: Custom certificates no longer supported
### What happened?

As of 5.6.0 custom certificates, defined by `secretRef` under the `tls` key, no longer work. The root-issuer is not installed (correct behavior) but mTLS client certs are installed referencing the former (incorrect behavior). We do not use mTLS, all of the `*.requireClientAuth` keys are set to `false`.

Besides that, it seems that the mTLS implementation is incomplete. Client certificates are only generated given the `.Values.listeners.kafka.tls.cert` configuration and the same key is ignored for `admin`, `rpc`, `schemaRegistry`, and `http`. If for any of the keys mentioned varying `*.tls.cert` values are used, this implementation breaks.

This is also discussed here in the following Slack thread: https://redpandacommunity.slack.com/archives/C01AJDUT88N/p1697552873739689

Using the custom provided certificates.

Provide any secret to the secret ref, the output contains mentions of issuer and mtls.

_No response_

_No response_

### Chart Version(s)

5.6.0 and up

### Cloud provider

n/a

zegerius:
closes #804

### DIFF
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 5.6.25
+version: 5.6.26
 
 # The app version is the default version of Redpanda to install.
 # ** NOTE for maintainers: please ensure the artifacthub image annotation is updated before merging
--- a/charts/redpanda/templates/_helpers.tpl
+++ b/charts/redpanda/templates/_helpers.tpl
@@ -639,8 +639,10 @@ return licenseSecretRef.key checks deprecated values entry if current values emp
 - name: redpanda-{{ $name }}-cert
   mountPath: {{ printf "/etc/tls/certs/%s" $name }}
     {{- end }}
+    {{- if (include "client-auth-required" . | fromJson).bool }}
 - name: mtls-client
   mountPath: /etc/tls/certs/{{ template "redpanda.fullname" $ }}-client
+    {{- end }}
   {{- end }}
 {{- end -}}
 
@@ -661,10 +663,12 @@ return licenseSecretRef.key checks deprecated values entry if current values emp
     secretName: {{ template "cert-secret-name" $r }}
     defaultMode: 0o440
     {{- end }}
+    {{- if (include "client-auth-required" . | fromJson).bool }}
 - name: mtls-client
   secret:
     secretName: {{ template "redpanda.fullname" $ }}-client
     defaultMode: 0o440
+    {{- end }}
   {{- end -}}
   {{- if and .Values.auth.sasl.enabled (not (empty .Values.auth.sasl.secretRef )) }}
 - name: users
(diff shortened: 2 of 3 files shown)


zegerius:
If a larger fix is required, I would propose to merge this to fix the current blocker for those using custom certs using secretRef and keeping the issue open (renamed) and fixing the mTLS related issues.

charts/redpanda/templates/_helpers.tpl
```diff
@@ -637,8 +637,10 @@ return licenseSecretRef.key checks deprecated values entry if current values emp
 - name: redpanda-{{ $name }}-cert
   mountPath: {{ printf "/etc/tls/certs/%s" $name }}
     {{- end }}
+    {{- if .Values.listeners.kafka.tls.requireClientAuth }}
```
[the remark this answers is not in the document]
joejulian: The client cert created in the certificates template will be signed by the one single CA so will be valid for any service that uses that CA for validation.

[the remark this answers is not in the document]
joejulian: We haven't had any specific use-case guiding this implementation so we're open to input. I think you're probably right that we should be checking the requireClientAuth for all the listeners.

zegerius:
I had to rebase with upstream, apologies for the conflicts. Should be fine to merge now. @joejulian

charts/redpanda/templates/_helpers.tpl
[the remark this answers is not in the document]
alejandroEsc: Would be best to have this in a helper function, something called "client-auth-required" to make this a bit more tidy.

charts/redpanda/Chart.yaml
```diff
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 5.6.19
+version: 5.6.20
```
alejandroEsc, line 26: will need this bumped to 21 for now.

charts/redpanda/templates/_helpers.tpl
```diff
@@ -639,8 +639,10 @@ return licenseSecretRef.key checks deprecated values entry if current values emp
...
 - name: mtls-client
-  mountPath: /etc/tls/certs/{{ template "redpanda.fullname" $ }}-client
+  mountPath: /etc/ls/certs/{{ template "redpanda.fullname" $ }}-client
```
alejandroEsc, line 644: /ls/certs? should this not be tls? If you are correct here we should fix this, that doesnt make sense.

charts/redpanda/templates/certs.yaml
```diff
@@ -76,6 +76,8 @@ spec:
 ---
   {{- $name := .Values.listeners.kafka.tls.cert }}
   {{- $data := get .Values.tls.certs $name }}
+  {{- if (empty $data.secretRef ) }}
+  {{- if or (.Values.listeners.kafka.tls.requireClientAuth .Values.listeners.admin.tls.requireClientAuth .Values.listeners.schemaRegistry.tls.requireClientAuth .Values.listeners.rpc.tls.requireClientAuth .Values.listeners.http.tls.requireClientAuth) }}
```
alejandroEsc, line 80: helper function here for readability please.

alejandroEsc (commented):
A few asks, but looking great, thank you!

zegerius:
@alejandroEsc thanks for the feedback. Resolved and bumped to latest again.

charts/redpanda/templates/certs.yaml
```diff
@@ -76,6 +76,8 @@ spec:
 ---
   {{- $name := .Values.listeners.kafka.tls.cert }}
   {{- $data := get .Values.tls.certs $name }}
+  {{- if (empty $data.secretRef ) }}
```
alejandroEsc, line 79: can we merge these two if statements?

alejandroEsc (commented):
looking good! just one question and another version bump. I will prioritize this PR next. Thank you.

alejandroEsc:
> @alejandroEsc does the added template function require change after #810 ?

I will take care of it, thank you.

alejandroEsc (approved):
lgtm

joejulian (approved):
Thanks!

PR #806, opened 2023-10-17
Redpanda: change default for redpanda users, opt in to user list. 

alejandroEsc:


RafalKorepta (approved):

joejulian:
We should add a check that will fail if the users list is empty and there is not secretRef.

RafalKorepta (approved):

PR #810, opened 2023-10-20
Add secretRef to listener TLS definition

RafalKorepta:

PR #816, opened 2023-10-21
Fix Redpanda schema values

RafalKorepta:
In previous PR https://github.com/redpanda-data/helm-charts/pull/810 I made mistake where secretRef is defined and it's schema.

REF:

https://github.com/redpanda-data/helm-charts/blob/3c27d893fa51e709390352f3c55a7a8a0178f712/charts/redpanda/templates/_helpers.tpl#L579-L580
https://github.com/redpanda-data/helm-charts/blob/3c27d893fa51e709390352f3c55a7a8a0178f712/charts/redpanda/templates/_helpers.tpl#L657-L661
https://github.com/redpanda-data/helm-charts/blob/3c27d893fa51e709390352f3c55a7a8a0178f712/charts/redpanda/values.yaml#L187-L188

### DIFF
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 5.6.26
+version: 5.6.27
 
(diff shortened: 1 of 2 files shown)


alejandroEsc (approved):
lgtm

PR #825, opened 2023-10-25
redpanda: change fullname when nameOverride is added

### LINKED ISSUE #727: fullname vs name in Redpanda helm chart
### What happened?

The Redpanda helm chart currently assigns `redpanda.fullname` as the name for the created StatefulSet.

`redpanda.fullname` is currently rendered as either fullnameOverride or release-name as of now. It does not make use of `redpanda.name` (e.g., via `printf "%s-%s" .Release.Name $name`):

```
{{- define "redpanda.fullname" -}}
{{- if .Values.fullnameOverride -}}
{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" -}}
{{- else -}}
{{- printf "%s" .Release.Name | trunc 63 | trimSuffix "-" -}}
{{- end -}}
{{- end -}}
```

I'm using Redpanda as a dependency in a Helm chart that brings up several services. But due to the above rendering, the StatefulSet and pods from the Redpanda chart show up with the top-level chart's release name (the stateful set is `release-name` and the pods are `release-name-0`...). That makes it hard to tell apart the redpanda resources from the top-level chart's own resources.

I was hoping to only set `redpanda.nameOverride=foo` in my chart's `values.yaml` and have the Redpanda StatefulSets and pods show up as something like `release-name-foo` instead.

<details>

```console
$ helm template ./redpanda --set nameOverride="foo"

<snip>
Source: redpanda/templates/statefulset.yaml
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: release-name # <------- does not use $name, could be release-name-$name perhaps.
  labels:
    app.kubernetes.io/component: foo
    app.kubernetes.io/instance: release-name
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: foo
    helm.sh/chart: redpanda-5.4.2
<snip>
```

</details>

_No response_

Redpanda

### Chart Version(s)

<details>

I noticed it first on 2.6.1, but this is also on the latest chart and current main.

</details>

### Cloud provider

NA

alejandroEsc:
Fixes #727 

Attempting to append to fullname whe nameOverride set.
We also try to keep chart backwards compatible and not change fullname if override not set.

### DIFF
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 5.6.29
+version: 5.6.30
 
 # The app version is the default version of Redpanda to install.
 # ** NOTE for maintainers: please ensure the artifacthub image annotation is updated before merging
--- a/charts/redpanda/templates/_helpers.tpl
+++ b/charts/redpanda/templates/_helpers.tpl
@@ -29,7 +29,11 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this 
 {{- if .Values.fullnameOverride -}}
 {{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" -}}
 {{- else -}}
+{{- if .Values.nameOverride }}
+{{- printf "%s-%s" .Release.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
+{{- else -}}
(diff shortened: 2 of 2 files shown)


RafalKorepta (changes requested):
As per https://github.com/redpanda-data/helm-charts/issues/727#issuecomment-1780938265 comment `nameOverride` should not be used in fullname template.

joejulian:
This is the wrong way to do this. See https://github.com/redpanda-data/helm-charts/issues/727#issuecomment-1782025771 Closing this PR.

PR #832, opened 2023-10-26
convert values passed to SI-to-bytes to a string before processing

joejulian:
Buildkite nightlies are failing with:

```
Error: INSTALLATION FAILED: template: redpanda/templates/statefulset.yaml:56:41: executing "redpanda/templates/statefulset.yaml" at <include "statefulset-checksum-annotation" .>: error calling include: template: redpanda/templates/_statefulset.tpl:116:44: executing "statefulset-checksum-annotation" at <include "configmap-content-no-seed" .>: error calling include: template: redpanda/templates/_configmap.tpl:91:23: executing "configmap-content-no-seed" at <include "SI-to-bytes" $element>: error calling include: template: redpanda/templates/_helpers.tpl:207:27: executing "SI-to-bytes" at <$si>: wrong type for value; expected string; got float64
```
because the value being passed to the helper is a float64. Convert it to a string so `hasSuffix` doesn't error.

### DIFF
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 5.6.33
+version: 5.6.34
 
 # The app version is the default version of Redpanda to install.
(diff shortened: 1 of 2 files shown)


alejandroEsc (approved):
lgtm, should be backwards compatible.

RafalKorepta (approved):

PR #838, opened 2023-10-27
bump the minimum required helm to 3.8.0

joejulian:
We're using combined logic that fails prior to 3.8.0. Enforce a minimum helm version.


alejandroEsc (approved):
Awesome change. Thank you

RafalKorepta (approved):

PR #851, opened 2023-11-03
fix:redpanda Set loadbalancer address only when extDns is enabled

josedev-union:


RafalKorepta:
Hi, @josedev-union Could you sign CLA?

RafalKorepta (approved):

alejandroEsc (approved):
The version needs to be bumped, otherwise lgtm

RafalKorepta:
https://github.com/redpanda-data/helm-charts/blob/e0403548ceb6a831de8fda3330453e61f3264f41/charts/redpanda/Chart.yaml#L26

I forgot about asking you to bump patch version.

PR #856, opened 2023-11-08
redpanda/fix: Fix the conditions for the values of console

### LINKED ISSUE #833: redpanda: console's secret is always created
### What happened?

charts/redpanda/value.yaml:

```
console:
  enabled: true
  configmap:
    create: false
  secret:
  deployment:
    create: false
  config: {}
```

<details>

```console
```

</details>

_No response_

Redpanda

### Chart Version(s)

- chart version : 5.6.33

### Cloud provider

<details>

</details>

koooge:
Fix #833

I suppose the conditions of .Values.console are opposite.

### DIFF
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 5.6.40
+version: 5.6.41
 
 # The app version is the default version of Redpanda to install.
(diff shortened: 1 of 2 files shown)


alejandroEsc:
> Fix #833
> 
> I suppose the conditions of .Values.console are opposite.

This isnt doing what you think it is doing. The overrides here are to determine if we create those for you, based on the values file of redpanda, or if you rather have the defaults deployed. It is confusing, i grant you that point. I am reviewing your original   issue and i may have some solutions. For now I think we should close this PR.

PR #861, opened 2023-11-09
 chore: do not print console secret unless we have a license

### LINKED ISSUE #833: redpanda: console's secret is always created

alejandroEsc:
First requires: https://github.com/redpanda-data/helm-charts/pull/870

Fixes #833

### DIFF
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 5.6.44
+version: 5.6.45
 
 # The app version is the default version of Redpanda to install.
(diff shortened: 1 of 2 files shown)


RafalKorepta (approved):
LGTM, but I don't understand comment included in the configmap-and-deployment.yaml file.

alejandroEsc:
> LGTM, but I don't understand comment included in the configmap-and-deployment.yaml file.

It is confusing. So we were not originally printing this resource. specially when it was specified that it was not being created. So now we are always printing because of the possibility of including a license. To avoid this we remove the printing always and put it behind a conditional.

PR #869, opened 2023-11-13
redpanda - fs validation initContainer

### LINKED ISSUE #863: Bash script to validate mount fs and exit with error if fails

alejandroEsc:
Fixes #863


charts/redpanda/values.schema.json
```diff
@@ -982,6 +982,23 @@
...
+                },
+                "expectedFS": {
+                  "type": "string"
+                },
+                "resources": {
+                  "type": "object"
+                },
+                "extraVolumeMounts": {
+                  "type": "string"
+                }
```
RafalKorepta, line 999: Should this be type string or object?

RafalKorepta (approved):
LGTM

charts/redpanda/values.schema.json
alejandroEsc: Yes good question. Maybe a bad choice from me, I was considering this to be an array for but after looking at a few existing charts with prior art it seems that the choice was to capture this as a string. Maybe this is too confusing.

PR #870, opened 2023-11-13
chore: console logic when secret creation is set to false

### LINKED ISSUE #833: redpanda: console's secret is always created

alejandroEsc:
partially fixes #833

### DIFF
--- a/charts/console/templates/deployment.yaml
+++ b/charts/console/templates/deployment.yaml
@@ -245,13 +245,6 @@ spec:
                 secretKeyRef:
                   name: {{ include "console.fullname" . }}
                   key: enterprise-license
-            {{- else if dig "licenseSecretRef" "name" "" .Values.enterprise }}
-            - name: LICENSE
-              valueFrom:
-                secretKeyRef:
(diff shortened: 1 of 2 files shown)


RafalKorepta (approved):

charts/console/templates/deployment.yaml
```diff
@@ -271,6 +264,14 @@ spec:
             - name: REDPANDA_ADMINAPI_TLS_CERTFILEPATH
               value: "/etc/console/secrets/redpanda-admin-api-tls-cert"
             {{- end }}
+            {{- else }}
```
joejulian, line 267: could you fix the indentation to make it clear where the logic block start/end points are?

alejandroEsc: I will only fix the relevant pieces for now.

PR #873, opened 2023-11-14
console: fix missing end statement

alejandroEsc:
Fix, not sure how I missed this or the linter.


alejandroEsc:
Note, tests will fail since they are relying on previous chart version.

RafalKorepta (approved):

PR #874, opened 2023-11-14
chore: fix post-upgrade job when it is set to null

alejandroEsc:
If a cert is null, this should still process and work. Here we add checks against that possibility.

### DIFF
--- a/charts/redpanda/templates/post-upgrade.yaml
+++ b/charts/redpanda/templates/post-upgrade.yaml
@@ -91,7 +91,7 @@ spec:
             if [ -d "/etc/secrets/users/" ]; then
                 IFS=":" read -r USER_NAME PASSWORD MECHANISM < <(grep "" $(find /etc/secrets/users/* -print))
                 curl -svm3 --fail --retry "120" --retry-max-time "120" --retry-all-errors --ssl-reqd \
-  {{- if $cert.caEnabled }}
+  {{- if and $cert ( dig "caEnabled" false $cert ) }}
                 --cacert /etc/tls/certs/{{ $service.tls.cert }}/ca.crt \
   {{- end }}
                 -X PUT -u ${USER_NAME}:${PASSWORD} \
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 5.6.44
+version: 5.6.46
 
 # The app version is the default version of Redpanda to install.
 # ** NOTE for maintainers: please ensure the artifacthub image annotation is updated before merging


charts/redpanda/templates/post-upgrade.yaml
```diff
@@ -91,7 +91,7 @@ spec:
             if [ -d "/etc/secrets/users/" ]; then
                 IFS=":" read -r USER_NAME PASSWORD MECHANISM < <(grep "" $(find /etc/secrets/users/* -print))
                 curl -svm3 --fail --retry "120" --retry-max-time "120" --retry-all-errors --ssl-reqd \
-  {{- if $cert.caEnabled }}
+  {{- if and $cert ( dig "caEnabled" false $cert ) }}
```
RafalKorepta, line 94: That problem repeats in many files

- https://github.com/redpanda-data/helm-charts/blob/9c4c5854a6deb22558d6dca25ccb485ea2b15eb8/charts/redpanda/templates/tests/test-pandaproxy-internal-tls-status.yaml#L65
- https://github.com/redpanda-data/helm-charts/blob/9c4c5854a6deb22558d6dca25ccb485ea2b15eb8/charts/redpanda/templates/tests/test-pandaproxy-internal-tls-status.yaml#L74
- https://github.com/redpanda-data/helm-charts/blob/9c4c5854a6deb22558d6dca25ccb485ea2b15eb8/charts/redpanda/templates/tests/test-schemaregistry-internal-tls-status.yaml#L96
- https://github.com/redpanda-data/helm-charts/blob/9c4c5854a6deb22558d6dca25ccb485ea2b15eb8/charts/redpanda/templates/tests/test-schemaregistry-internal-tls-status.yaml#L107
- https://github.com/redpanda-data/helm-charts/blob/9c4c5854a6deb22558d6dca25ccb485ea2b15eb8/charts/redpanda/templates/tests/test-schemaregistry-internal-tls-status.yaml#L123
- https://github.com/redpanda-data/helm-charts/blob/9c4c5854a6deb22558d6dca25ccb485ea2b15eb8/charts/redpanda/templates/tests/test-schemaregistry-internal-tls-status.yaml#L139
- https://github.com/redpanda-data/helm-charts/blob/9c4c5854a6deb22558d6dca25ccb485ea2b15eb8/charts/redpanda/templates/tests/test-schemaregistry-internal-tls-status.yaml#L149
- https://github.com/redpanda-data/helm-charts/blob/9c4c5854a6deb22558d6dca25ccb485ea2b15eb8/charts/redpanda/templates/_configmap.tpl#L600
- https://github.com/redpanda-data/helm-charts/blob/9c4c5854a6deb22558d6dca25ccb485ea2b15eb8/charts/redpanda/templates/_configmap.tpl#L613

RafalKorepta:
> chore: fix post-upgrade job when it is set to null

NIT: It's not chore change in my opinion as @vuldin needs a Redpanda deployment that can be achieved without cert manager. 

I would like to suggest that any occurrence of

```
{{- $cert := get .Values.tls.certs .Values.listeners.kafka.tls.cert }}
```

is a place to address entire problem.

alejandroEsc:
> > chore: fix post-upgrade job when it is set to null
> 
> NIT: It's not chore change in my opinion as @vuldin needs a Redpanda deployment that can be achieved without cert manager.
> 
> I would like to suggest that any occurrence of
> 
> ```
> {{- $cert := get .Values.tls.certs .Values.listeners.kafka.tls.cert }}
> ```
> 
> is a place to address entire problem.

🤔 but the null issue only appears in the operator. If i understand correctly, you are ok with the solution, but you want to be holistic and take care of each instance. Could I open an issue and do so for your findings in another PR?

RafalKorepta (approved):
I retested and it is working fine. I'm not sure why other occurrences mentioned in my comment are not complaining.

PR #880, opened 2023-11-20
Set TTL for jobs

vuldin:
This small change makes the `redpanda-configuration` and `redpanda-post-upgrade` jobs (and their associated pods) automatically be cleaned up after successful runs. This frees up their resources and is recommended (see note [here](https://kubernetes.io/docs/concepts/workloads/controllers/job/#ttl-mechanism-for-finished-jobs).

Keeping in draft for now since I need to ensure the commit happens with the proper email address, and I need to add the change to the 2nd job.

### DIFF
--- a/charts/redpanda/templates/post-install-upgrade-job.yaml
+++ b/charts/redpanda/templates/post-install-upgrade-job.yaml
@@ -40,6 +40,7 @@ metadata:
   {{- toYaml . | nindent 4 }}
 {{- end }}
 spec:
+  ttlSecondsAfterFinished: 10
   template:
     metadata:
       generateName: "{{ .Release.Name }}-post-"


RafalKorepta:
@vuldin Could you sign CLA?

RafalKorepta:
I still feel it might be not enough to scrape logs from those jobs in customer clusters. It will be very hard to debug such environment. 

If you feel it should be available for customer then create as a variable in values.yaml

alejandroEsc:
> I still feel it might be not enough to scrape logs from those jobs in customer clusters. It will be very hard to debug such environment.
> 
> If you feel it should be available for customer then create as a variable in values.yaml

I agree, this is why we have not done this yet, we would like logs in case of failures.

vuldin:
> @vuldin Could you sign CLA?
Hmm, I've signed the CLA but this still shows up for some reason (no matter how many times I sign it).

david-yu:

PR #881, opened 2023-11-20
Move TLS parameters inside if statements

vuldin:
Small change to move the start of the two TLS sections `admin_api_tls` and `kafka_api_tls` into their associated if statement.

### DIFF
--- a/charts/redpanda/templates/_configmap.tpl
+++ b/charts/redpanda/templates/_configmap.tpl
@@ -240,8 +240,8 @@ redpanda.yaml: |
         port: {{ $listener.port }}
   {{- end }}
 {{- end }}
-    admin_api_tls:
 {{- if (include "admin-internal-tls-enabled" . | fromJson).bool }}
+    admin_api_tls:
       - name: internal
         enabled: true
         cert_file: /etc/tls/certs/{{ $service.tls.cert }}/tls.crt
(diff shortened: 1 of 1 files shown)


RafalKorepta (approved):
LGTM, please sign CLA

charts/redpanda/templates/_configmap.tpl
```diff
@@ -165,8 +165,8 @@ redpanda.yaml: |
         port: {{ $listener.port }}
   {{- end }}
 {{- end }}
-    admin_api_tls:
```
alejandroEsc, line 168: To be clear, nothing will break if this is empty nor if this is left out?

alejandroEsc (commented):
one question, non blocking.

charts/redpanda/templates/_configmap.tpl
vuldin: It seems this is the case, which is surprising to me as it could result in an empty section in the yaml. I'm not sure why disabling TLS works without the change in this PR.

vuldin:
I've signed the CLA, but Github is now thinking showing me two CLA notices in this thread and only one is green:

image

ivotron:
that's a race we've seen elsewhere. fine to ignore since we have that green one

david-yu:

PR #882, opened 2023-11-20
WIP Stretch cluster support

vuldin:
This PR adds the ability to configure helm deployments so that Redpanda can form a stretch cluster. A few notes:

- allows enabling RPC communication by external hostnames
- allows adding additional seed servers so subsequent deployments can reach out to already existing deployments to find leader
- new parameters are intentionally left out of the values.yaml (but they should likely be added to the schema regardless). The reason is because modifying `additionalSeedServers` or enabling external RPC is error prone and is difficult to configure. The goal is to enable setting these values but then provide a guide showing how a tool like Crossplane could be used to manage multiple helm deployments (and managing the proper values for these new parameters in values.yaml).

TODOs:

- external hostname must be resolvable, so I've ensure this is the case by updating the hosts file on Redpanda pods with the external hostnames and LB IP addresses
- this is the start of other work, including an investigation and PoC with [Crossplane](https://docs.crossplane.io/knowledge-base/guides/multi-tenant/#multi-cluster-multi-tenancy)

### DIFF
--- a/charts/redpanda/templates/_configmap.tpl
+++ b/charts/redpanda/templates/_configmap.tpl
@@ -560,6 +560,10 @@ rpk:
     {{- $server := dict "host" (dict "address" . "port" $.Values.listeners.rpc.port) -}}
     {{- $serverList = append $serverList $server -}}
   {{- end -}}
+  {{- range $seed := .Values.additionalSeedServers -}}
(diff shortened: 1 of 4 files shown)


david-yu:
Closing as this work is on being driven in the Operator.

PR #883, opened 2023-11-21
Bump Operator chart version

RafalKorepta:
In this PR:

* nightly script that detects version bump
  * works in Mac and Linux environment
  * can bump minor and patch release
* bumps Operator chart version and AppVersion
* Introduces new Roles required for automatic migration

### DIFF
--- a/.github/bump_chart_versions.sh
+++ b/.github/bump_chart_versions.sh
@@ -20,7 +20,7 @@ get_chart_appversion() {
 }
 
 get_chart_app_constraint() {
-    CONSTRAINT=$(yq -r .appVersion "${CHARTFILE}" | awk '{patch=split($1,version,".");version[patch]++; print "~"version[1],version[2]}' OFS=.)
+    CONSTRAINT=$(yq -r .appVersion "${CHARTFILE}" | awk '{patch=split($1,version,".");version[patch]++; print "~"version[1]}' OFS=.)
 }
 
 get_repo() {
--- a/charts/operator/Chart.yaml
+++ b/charts/operator/Chart.yaml
@@ -6,11 +6,11 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 0.4.1
+version: 0.4.2
 
 # This is the default version of the operator being deployed.
 # ** NOTE for maintainers: please ensure the artifacthub image annotation is updated before merging
-appVersion: v2.0.0
+appVersion: v2.1.0
 
 sources:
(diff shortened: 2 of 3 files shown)


.github/bump_chart_versions.sh
```diff
@@ -51,6 +51,6 @@ if [[ "$CHART" == "redpanda" ]]; then
 fi
 
 bump_patch_version
-sed -i "s@^version: .*\$@version: ${NEW_VERSION}@" "${CHARTFILE}"
-sed -i "s@^appVersion: .*\$@appVersion: ${LATEST}@" "${CHARTFILE}"
-sed -i "s@${REPO}:${APPVERSION}\$@${REPO}:${LATEST}@" "${CHARTFILE}"
+perl -i -pe "s@^version: .*\$@version: ${NEW_VERSION}@" "${CHARTFILE}"
+perl -i -pe "s@^appVersion: .*\$@appVersion: ${LATEST}@" "${CHARTFILE}"
+perl -i -pe "s@${REPO}:${APPVERSION}\$@${REPO}:${LATEST}@" "${CHARTFILE}"
```
sbocinec, line 56: question: why are we going replace perl for sed? I like perl, but find it not always being installed on the modern OSs by default. sed is usually always available.

RafalKorepta: In the MacOS sed with in place replacement doesn't work correctly. I'm working on Mac, so I needed this script to work.

Based on https://stackoverflow.com/questions/4247068/sed-command-with-i-option-failing-on-mac-but-works-on-linux I tested `sed -i'' -e ...`, but it created backup file which is useless (named `Chart.yaml-e`). That's why I switched to perl.

sbocinec (approved):
Thank you, LGTM

.github/bump_chart_versions.sh
c4milo: `brew install gnu-sed` and create an alias to it or put /usr/local/bin first in your PATH.

sbocinec: From what I see,  using `sed -i '.bak' -e ...` might work to not create that cryptic file extension?, THough I see, that if this additionally created backup file is causing issues during the build and the `perl` approach works, let's go with perl :-)  I have approved. Thank you!

c4milo (approved):

alejandroEsc:
An I reading right, are you using perl? In the script?

RafalKorepta:
> An I reading right, are you using perl? In the script?

No. After https://github.com/redpanda-data/helm-charts/pull/883#discussion_r1400296271 @c4milo suggestion I dropped that commit.

PR #887, opened 2023-11-22
redpanda: allow recovery_mode 

### LINKED ISSUE #795: Provide a way to start a cluster in recovery mode

https://github.com/redpanda-data/core-internal/issues/783 introduces recovery mode. This tracks ensuring it can be enabled in k8s environments. 

OperatorV2 deployments should test and document configuring recovery mode.

### Why is this needed?

New 23.3 feature

alejandroEsc:
Note that it may be necessary to roll manually (delete the pods) after leaving recovery_mode. This is just a matter of restarting the pods, no need to change update strategy from my own testing. 

Fixes #795 

c.c. @JakeSCahill

### DIFF
--- a/charts/redpanda/templates/secrets.yaml
+++ b/charts/redpanda/templates/secrets.yaml
@@ -80,9 +80,11 @@ stringData:
       touch /tmp/postStartHookFinished
     }
 
+  {{- if not ( dig "node" "recovery_mode_enabled" false .Values.config ) }}
     postStartHook
-    true
+  {{- end }}
 
+    true
 
   preStop.sh: |-
     #!/usr/bin/env bash
@@ -118,11 +120,11 @@ stringData:
       touch /tmp/preStopHookFinished
     }
 
-{{- if gt ( .Values.statefulset.replicas | int64 ) 2 }}
+{{- if and ( gt ( .Values.statefulset.replicas | int64 ) 2) ( not ( dig "node" "recovery_mode_enabled" false .Values.config ) ) }}
     preStopHook
 {{- else }}
     touch /tmp/preStopHookFinished
-    echo "Not enough replicas to put a broker into maintenance mode."
+    echo "Not enough replicas or in recovery mode, cannot put a broker into maintenance mode."
 {{- end }}
     true
 {{- if and (not (empty .Values.auth.sasl.secretRef)) (and .Values.auth.sasl.enabled .Values.auth.sasl.users) }}
--- a/charts/redpanda/templates/secrets.yaml
+++ b/charts/redpanda/templates/secrets.yaml
@@ -80,10 +80,7 @@ stringData:
       touch /tmp/postStartHookFinished
     }
 
-  {{- if not ( dig "node" "recovery_mode_enabled" false .Values.config ) }}
     postStartHook
-  {{- end }}
-
     true
 
   preStop.sh: |-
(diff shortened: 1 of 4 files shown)


charts/redpanda/templates/statefulset.yaml
```diff
@@ -250,8 +250,12 @@ spec:
                 - -c
                 - |
                   set -x
+                  {{- if ( dig "node" "recovery_mode_enabled" false .Values.config ) }}
```
RafalKorepta, line 253: I would suggest to disable whole readiness probe, so that there will be no initial delay (via `initialDelaySeconds`).

charts/redpanda/templates/secrets.yaml
```diff
@@ -80,9 +80,11 @@ stringData:
       touch /tmp/postStartHookFinished
     }
 
+  {{- if not ( dig "node" "recovery_mode_enabled" false .Values.config ) }}
     postStartHook
-    true
+  {{- end }}
 
+    true
```
RafalKorepta, line 87: In my opinion this will backfire in the worst possible moment. Disable or exit from maintenance mode should be possible even in recovery mode.

charts/redpanda/templates/secrets.yaml
```diff
@@ -80,9 +80,11 @@ stringData:
       touch /tmp/postStartHookFinished
     }
 
+  {{- if not ( dig "node" "recovery_mode_enabled" false .Values.config ) }}
```
RafalKorepta, line 83: Should this condition have something like
```suggestion
  {{- if and (not ( dig "node" "recovery_mode_enabled" false .Values.config )) (include "redpanda-atleast-23-3-1" $root | fromJson).bool) }}
```

charts/redpanda/templates/secrets.yaml
alejandroEsc: What does this mean:
```
backfire in the worst possible moment
```

charts/redpanda/templates/secrets.yaml
alejandroEsc: Good point, but honestly RP will complain about not knowing this configuration before we hit this.

charts/redpanda/templates/statefulset.yaml
alejandroEsc: sure.

charts/redpanda/templates/secrets.yaml
RafalKorepta: When you will be in the incident after long hours of the back and forth between multiple stakeholders then it will hard to remember to validate if maintenance mode was disabled (the main logic of postStartHook). In my opinion it should always run and Redpadna should accept that REST request.

RafalKorepta (approved):

JakeSCahill:
> it may be necessary to roll manually (delete the pods) after leaving recovery_mode

@alejandroEsc do you know in what circumstances a roll is necessary?

alejandroEsc:
> > it may be necessary to roll manually (delete the pods) after leaving recovery_mode
> 
> @alejandroEsc do you know in what circumstances a roll is necessary?

in cases where cluster health cannot return true, for example let's say you have a nodeID that is down and cannot be recovered (ghost broker).

JakeSCahill:
Thanks 🙏 so in that case you roll the whole statefulset?

PR #899, opened 2023-11-28
Move Pod from ClusterRole to Role

RafalKorepta:


alejandroEsc (approved):
LGTM, I am guessing you verified this with the roles in the manifests right?

RafalKorepta:
> I am guessing you verified this with the roles in the manifests right?

No. I'm relaying on our chart testing test suite.

In our operator code the definition is scoped to namespace only https://github.com/redpanda-data/redpanda-operator/blob/6c288ea2d50c57e2fe8cde38c624554c348b8036/src/go/k8s/internal/controller/redpanda/redpanda_decommission_controller.go#L37

PR #905, opened 2023-11-30
Redpanda: taking proper care of configs that are arrays

alejandroEsc:
Fixes: https://github.com/redpanda-data/team-kubernetes-internal/issues/14

### DIFF
--- a/charts/redpanda/templates/tests/test-kafka-nodelete.yaml
+++ b/charts/redpanda/templates/tests/test-kafka-nodelete.yaml
@@ -1,0 +1,103 @@
+{{/*
+Licensed to the Apache Software Foundation (ASF) under one or more
+contributor license agreements.  See the NOTICE file distributed with
+this work for additional information regarding copyright ownership.
+The ASF licenses this file to You under the Apache License, Version 2.0
+(the "License"); you may not use this file except in compliance with
+the License.  You may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/}}
+{{- if dig "kafka_nodelete_topics" "[]" $.Values.config.cluster }}
+{{- $sasl := .Values.auth.sasl }}
+{{- $root := deepCopy . }}
+{{- $rpk := deepCopy . }}
+apiVersion: v1
+kind: Pod
+metadata:
+  name: {{ include "redpanda.fullname" . }}-test-kafka-nodelete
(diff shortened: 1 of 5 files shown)


RafalKorepta (changes requested):
Could you add test?

RafalKorepta:
I would leverage https://docs.redpanda.com/current/reference/cluster-properties/#kafka_nodelete_topics and create test Pod that tries delete topic.

charts/redpanda/templates/tests/test-kafka-nodelete.yaml
```diff
@@ -0,0 +1,103 @@
...
+          {{- else }}
+          REDPANDA_SASL_MECHANISM=${REDPANDA_SASL_MECHANISM:-{{ .Values.auth.sasl.mechanism | upper }}}
+          {{- end }}
+          export {{ include "rpk-sasl-environment-variables" . }}
+          if [[ -n "$old_setting" ]]; then set -x; fi
+{{- end }}
+          {{- $i := .Values.statefulset.replicas }}
+          {{- $default_topic_replicas := sub (add $i (mod $i 2)) 1 }}
+          # wait for post-upgrade job to update the default_topic_replications value
+          timeout 600 bash -c "until [[ $(rpk cluster config get default_topic_replications) = {{ $default_topic_replicas }} ]]; do sleep 1; done"
```
RafalKorepta, line 75:
```suggestion
          timeout 120 bash -c "until [[ $(rpk cluster config get default_topic_replications) = {{ $default_topic_replicas }} ]]; do sleep 1; done"
```

The time out should not be grater than 120 seconds defined in https://github.com/redpanda-data/helm-charts/pull/905/files#diff-3f599d2fdc6489c00bfddc47fc46223cd8fa1dc87c65833bd5f0ecfac9326aeeR51

charts/redpanda/templates/tests/test-kafka-nodelete.yaml
```diff
@@ -0,0 +1,103 @@
...
+  
+          exists=$(rpk topic list | grep my_sample_topic | awk '{print $1}')
```
RafalKorepta, line 77: If possible the `helm test` command should be successful even if `kafka_nodelete_topics` is defined but lack of `my_sample_topic`.

Simple check in go temaplte if array has `my_sample_topic` included.

alejandroEsc: ok, ill add a check

charts/redpanda/templates/tests/test-kafka-nodelete.yaml
```diff
@@ -89,12 +90,14 @@ spec:
          
           # now check if we can delete the topic (we should not)
           rpk topic delete my_sample_topic
+
+          {{- if has "my_sample_topic" $noDeleteTopics }}
```
RafalKorepta, line 94: I was thinking that check would be at the top, so the Pod would not be executed at all.

PR #909, opened 2023-12-02
Use enabled switch for each listener

RafalKorepta:
Each external listener in the values-schema has enabled boolean flag that was not used. This commit starts to use it.


charts/redpanda/Chart.yaml
```diff
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 5.6.55
+version: 5.6.56
```
sbocinec, line 26: nit: helm-docs should be executed so that the README reflects the version changes

RafalKorepta: Ok I will do that :)

sbocinec (approved):
redpanda README should be updated to reflect the changes, other than that LGTM

PR #913, opened 2023-12-05
redpanda: sidecar controllers were not being tested

alejandroEsc:
Needed rbac roles to deploy controllers, this enables the ci testing of the additional controllers.

### DIFF
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 5.6.57
+version: 5.6.58
 
 # The app version is the default version of Redpanda to install.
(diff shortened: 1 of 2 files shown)


RafalKorepta:
Why this didn't fail in CI?

RafalKorepta (approved):
I don't understand why we are changing something that we don't validate. From my point of view missing RBAC should indicate that sidecar conrtoller should fail.

alejandroEsc:
> I don't understand why we are changing something that we don't validate. From my point of view missing RBAC should indicate that sidecar conrtoller should fail.

I found this issue from our helm testing in the operator. Once we added the right configurations to turn on the controllers those tests started to fail.

alejandroEsc:
> I don't understand why we are changing something that we don't validate. From my point of view missing RBAC should indicate that sidecar conrtoller should fail.

Example: 
https://buildkite.com/redpanda/redpanda-operator/builds/263#018c3a5f-e97a-4955-bda4-6a2123f99219

PR #904, opened 2023-11-30
Add clarification to readinessProbe

JFlath:

PR #923, opened 2023-12-07
Re-clarify readiness probe

JFlath:
In #904 I typo'd the name of the parameter - it's a minor thing, but best to get these details right!

Sorry!


alejandroEsc (approved):
ha, i should have caught this, thank you for fixing it.

PR #930, opened 2023-12-12
Bump operator image tag version to v2.1.10-23.2.18

RafalKorepta:


alejandroEsc (approved):
Lgtm

YanivKunda:
@RafalKorepta / @alejandroEsc - Post-merge question:
What about the appVersion/image in helm-charts/charts/redpanda/Chart.yaml ?
In https://github.com/redpanda-data/helm-charts/blob/main/charts/redpanda/Chart.yaml it still points to v23.2.17 -
Is there a different schedule for aligning it?

YanivKunda:
> 
> That refers to the binary for redpanda, the image changed for the operator is for the operator itself.
So the using the operator via its new chart should be able to upgrade the redpanda instance to v23.2.18 (I saw it references the new image)?
From what I read in https://docs.redpanda.com/current/deploy/deployment-option/self-hosted/kubernetes/kubernetes-production-deployment/ I can't seem to understand whether the operator depends on the redpanda helm chart (which wasn't updated yet) or not.

alejandroEsc:
> > That refers to the binary for redpanda, the image changed for the operator is for the operator itself.
> 
> So the using the operator via its new chart should be able to upgrade the redpanda instance to v23.2.18 (I saw it references the new image)? From what I read in https://docs.redpanda.com/current/deploy/deployment-option/self-hosted/kubernetes/kubernetes-production-deployment/ I can't seem to understand whether the operator depends on the redpanda helm chart (which wasn't updated yet) or not.

It can upgrade to that version and should support all previous supported versions of redpanda. You are right that we should also update the redpanda chart to reflect the patch change also. Sorry we missed that.

PR #934, opened 2023-12-12
redpanda: expose audit logging feature

alejandroEsc:
Fixes https://github.com/redpanda-data/team-kubernetes-internal/issues/31

Note that this is essentially a new feature we are adding to the helm chart, as such, we expect the following to happen to consider this complete:

Sample values: 
```
auditLogging:
  enabled: true
  listener: default
  partitions: 12
```

### DIFF
--- a/charts/redpanda/ci/96-audit-logging-values.yaml.tpl
+++ b/charts/redpanda/ci/96-audit-logging-values.yaml.tpl
@@ -12,11 +12,18 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-debug: true
-remote: origin
-target-branch: main
-helm-extra-args: --timeout 600s
-chart-repos:
-  - redpanda=https://charts.redpanda.com
-charts:
-  - charts/redpanda
+---
+enterprise:
+  license: "${REDPANDA_LICENSE}"
+
+auth:
+  sasl:
+    enabled: true
+    users:
+      - name: admin
+        password: change-me
+        mechanism: SCRAM-SHA-512
+
+auditLogging:
+  enabled: true
+  listeners: default
--- a/charts/redpanda/templates/_configmap.tpl
+++ b/charts/redpanda/templates/_configmap.tpl
@@ -90,6 +90,44 @@ bootstrap.yaml: |
   {{- if and (not (hasKey .Values.config.cluster "storage_min_free_bytes")) ((include "redpanda-atleast-22-2-0" . | fromJson).bool) }}
   storage_min_free_bytes: {{ include "storage-min-free-bytes" . }}
   {{- end }}
+{{/* AUDIT LOGS */}}
+{{- if (include "redpanda-atleast-23-3-0" . | fromJson).bool }}
+  {{- if and ( dig "enabled" "false" .Values.auditLogging ) (include "sasl-enabled" $root | fromJson).bool }}
+  audit_enabled: true
+  {{- if not (eq (int .Values.auditLogging.clientMaxBufferSize) 16777216 ) }}
+  audit_client_max_buffer_size: {{ .Values.auditLogging.clientMaxBufferSize }}
+  {{- end }}
+  {{- if not (eq (int .Values.auditLogging.queueDrainIntervalMs) 500) }}
+  audit_queue_drain_interval_ms: {{ .Values.auditLogging.queueDrainIntervalMs }}
+  {{- end }}
+  {{- if not (eq (int .Values.auditLogging.queueMaxBufferSizePerShard) 1048576) }}
+  audit_queue_max_buffer_size_per_shard: {{ .Values.auditLogging.queueMaxBufferSizePerShard }}
+  {{- end }}
+  {{- if not (eq (int .Values.auditLogging.partitions) 12) }}
+  audit_log_num_partitions: {{ .Values.auditLogging.partitions }}
+  {{- end }}
+    {{- if dig "enabledEventTypes" "" .Values.auditLogging }}
+  audit_enabled_event_types:
+      {{- with .Values.auditLogging.enabledEventTypes }}
+        {{- toYaml . | nindent 2 }}
+      {{- end }}
+    {{- end }}
+    {{- if dig "excludedTopics" "" .Values.auditLogging }}
+  audit_excluded_topics:
+      {{- with .Values.auditLogging.excludedTopics }}
+        {{- toYaml . | nindent 2 }}
+      {{- end }}
+    {{- end }}
+    {{- if dig "excludedPrincipals" "" .Values.auditLogging }}
+  audit_excluded_principals:
+      {{- with .Values.auditLogging.excludedPrincipals }}
+        {{- toYaml . | nindent 2 }}
+      {{- end }}
+    {{- end }}
+  {{- else }}
+  audit_enabled: false
+  {{- end }}
+{{- end }}
 {{- if and (include "is-licensed" . | fromJson).bool (include "storage-tiered-config" .|fromJson).cloud_storage_enabled }}
   {{- $tieredStorageConfig := (include "storage-tiered-config" .|fromJson) }}
   {{- $tieredStorageConfig = unset $tieredStorageConfig "cloud_storage_cache_directory" }}
@@ -163,6 +201,44 @@ redpanda.yaml: |
     {{- end }}
   {{- end }}
 {{- end -}}
+{{/* AUDIT LOGS */}}
+{{- if (include "redpanda-atleast-23-3-0" . | fromJson).bool }}
+  {{- if and ( dig "enabled" "false" .Values.auditLogging ) (include "sasl-enabled" $root | fromJson).bool }}
+    audit_enabled: true
+  {{- if not (eq (int .Values.auditLogging.clientMaxBufferSize) 16777216) }}
+    audit_client_max_buffer_size: {{ .Values.auditLogging.clientMaxBufferSize }}
+  {{- end }}
+  {{- if not (eq (int .Values.auditLogging.queueDrainIntervalMs) 500) }}
+    audit_queue_drain_interval_ms: {{ .Values.auditLogging.queueDrainIntervalMs }}
+  {{- end }}
+  {{- if not (eq (int .Values.auditLogging.queueMaxBufferSizePerShard) 1048576) }}
+    audit_queue_max_buffer_size_per_shard: {{ .Values.auditLogging.queueMaxBufferSizePerShard }}
+  {{- end }}
+  {{- if not (eq (int .Values.auditLogging.partitions) 12) }}
+    audit_log_num_partitions: {{ .Values.auditLogging.partitions }}
+  {{- end }}
+    {{- if dig "enabledEventTypes" "" .Values.auditLogging }}
+    audit_enabled_event_types:
+      {{- with .Values.auditLogging.enabledEventTypes }}
+        {{- toYaml . | nindent 4 }}
+      {{- end }}
+    {{- end }}
+    {{- if dig "excludedTopics" "" .Values.auditLogging }}
+    audit_excluded_topics:
+      {{- with .Values.auditLogging.excludedTopics }}
+        {{- toYaml . | nindent 4 }}
+      {{- end }}
+    {{- end }}
+    {{- if dig "excludedPrincipals" "" .Values.auditLogging }}
+    audit_excluded_principals:
+      {{- with .Values.auditLogging.excludedPrincipals }}
+        {{- toYaml . | nindent 4 }}
+      {{- end }}
+    {{- end }}
+  {{- else }}
+    audit_enabled: false
+  {{- end }}
+{{- end }}
 {{/* LISTENERS */}}
 {{/* Admin API */}}
 {{- $service := .Values.listeners.admin }}
@@ -186,7 +262,7 @@ redpanda.yaml: |
         require_client_auth: {{ $service.tls.requireClientAuth }}
   {{- $cert := get .Values.tls.certs $service.tls.cert }}
   {{- if empty $cert }}
-    {{- fail (printf "Certificate, '%s', used but not defined")}}
+    {{- fail (printf "Certificate used but not defined")}}
   {{- end }}
   {{- if $cert.caEnabled }}
         truststore_file: /etc/tls/certs/{{ $service.tls.cert }}/ca.crt
@@ -247,7 +323,7 @@ redpanda.yaml: |
         require_client_auth: {{ $kafkaService.tls.requireClientAuth }}
   {{- $cert := get .Values.tls.certs $kafkaService.tls.cert }}
   {{- if empty $cert }}
-    {{- fail (printf "Certificate, '%s', used but not defined")}}
+    {{- fail (printf "Certificate used but not defined")}}
   {{- end }}
   {{- if $cert.caEnabled }}
         truststore_file: /etc/tls/certs/{{ $kafkaService.tls.cert }}/ca.crt
@@ -293,7 +369,7 @@ redpanda.yaml: |
       require_client_auth: {{ $service.tls.requireClientAuth }}
   {{- $cert := get .Values.tls.certs $service.tls.cert }}
   {{- if empty $cert }}
-    {{- fail (printf "Certificate, '%s', used but not defined")}}
+    {{- fail (printf "Certificate used but not defined")}}
   {{- end }}
   {{- if $cert.caEnabled }}
       truststore_file: /etc/tls/certs/{{ $service.tls.cert }}/ca.crt
@@ -325,10 +401,10 @@ redpanda.yaml: |
   {{- $schemaRegistryService := .Values.listeners.schemaRegistry }}
   schema_registry_client:
     brokers:
-      {{- range (include "seed-server-list" $root | mustFromJson) }}
-    - address: {{ . }}
+    {{- range $id, $item := $root.tempConfigMapServerList }}
+    - address: {{ $item.host.address }}
       port: {{  $kafkaService.port }}
-      {{- end }}
+    {{- end }}
     {{- if (include "kafka-internal-tls-enabled" . | fromJson).bool }}
     broker_tls:
       enabled: true
@@ -337,7 +413,7 @@ redpanda.yaml: |
       key_file: /etc/tls/certs/{{ $kafkaService.tls.cert }}/tls.key
       {{- $cert := get .Values.tls.certs $kafkaService.tls.cert }}
       {{- if empty $cert }}
-        {{- fail (printf "Certificate, '%s', used but not defined")}}
+        {{- fail (printf "Certificate used but not defined")}}
       {{- end }}
       {{- if $cert.caEnabled }}
       truststore_file: /etc/tls/certs/{{ $kafkaService.tls.cert }}/ca.crt
@@ -383,7 +459,7 @@ redpanda.yaml: |
         require_client_auth: {{ $schemaRegistryService.tls.requireClientAuth }}
     {{- $cert := get .Values.tls.certs $schemaRegistryService.tls.cert }}
     {{- if empty $cert }}
-      {{- fail (printf "Certificate, '%s', used but not defined")}}
+      {{- fail ( printf "Certificate used but not defined" )}}
     {{- end }}
     {{- if $cert.caEnabled }}
         truststore_file: /etc/tls/certs/{{ $schemaRegistryService.tls.cert }}/ca.crt
@@ -401,7 +477,7 @@ redpanda.yaml: |
       {{- $certPath := printf "/etc/tls/certs/%s" $certName }}
       {{- $cert := get $values.tls.certs $certName }}
       {{- if empty $cert }}
-        {{- fail (printf "Certificate, '%s', used but not defined")}}
+        {{- fail ( printf "Certificate, '%s', used but not defined" $certName )}}
       {{- end }}
       - name: {{ $name }}
         enabled: true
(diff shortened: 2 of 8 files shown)


charts/redpanda/values.yaml
```diff
@@ -87,6 +87,19 @@ license_secret_ref: {}
...
+#    for auditing to work
+auditLog:
+  # -- Enable or disable audit logging, for production clusters we suggest you enable.
+  #    however, this will only work if you also enable sasl and a listener with sasl enabled.
+  enabled: true
+  # -- Kafka external listener name, note that it must have `authenticationMethod` set to sasl
+  listener: default
+  # --  Integer value defining the number of partitions used by a newly created audit topic
+  partitions: 12
+  #
```
michael-redpanda, line 100: would also suggest adding in:
* `audit_enabled_event_types: ["management","authenticate","admin"]`
* `audit_excluded_topics: []`
* `audit_excluded_principals: []`

Reasonings:

* `audit_enabled_event_types` is probably the 'hottest' thing a customer would want to modify, this controls the content of the audit log and what operations are audited
* `audit_excluded_topics` maybe not as popular but I think this provides a hint to the customer that they can excluded certain topics from auditing that would reduce the amount of audit messages in the topic
* `audit_excluded_principals` along the same vein as `audit_excluded_topics` but probably more useful.  They may assign a specific user to be the auditor (the principal responsible for collecting audit logs).  Regardless of the value of `audit_enabled_event_types`, any access to the audit log is audited, except by principals listed in this list.  So the idea would be that they have an auditor user with ACLS that permit them to only read from the audit log topic.  Without adding that principal to this list, every access to that topic by that user is audited.  By adding the user to this list, those events are not generated, reducing 'junk' in the audit log

alejandroEsc:
Please do not merge yet!

charts/redpanda/values.yaml
```diff
@@ -87,6 +87,26 @@ license_secret_ref: {}
...
+auditLogging:
+  # -- Enable or disable audit logging, for production clusters we suggest you enable.
+  #    however, this will only work if you also enable sasl and a listener with sasl enabled.
+  enabled: true
+  # -- Kafka external listener name, note that it must have `authenticationMethod` set to sasl
+  listener: default
+  # -- Integer value defining the number of partitions used by a newly created audit topic
+  partitions: 12
+  # -- Integer valued specifies the replication factor for a newly created audit log topic
+  logReplicationFactor: 3
```
michael-redpanda, line 101: Not sure we want to add this.  By default, if this config is not set, the audit log system will use `internal_topic_replication_factor` to determine it's RF.  By default that is 3.  What's helpful about using `internal_topic_replication_factor` is that if the cluster is only one node, the value auto-updates to 1.  By hard setting this to 3, the topic will attempt to be created with an RF of 3 which will fail on a single node cluster.

My suggestion is to remove this.

charts/redpanda/values.schema.json
```diff
@@ -1552,6 +1552,32 @@
...
+        "enabledEventTypes": {
+          "type": ["array", "null"]
+        },
+        "excludedTopics": {
+          "type": ["array", "null"]
+        },
+        "excludedPrincipals": {
+          "type": ["array", "null"]
+        }
+      }
```
michael-redpanda, line 1580: adding in remaining cluster configs

```suggestion
    "auditLogging": {
      "type": "object",
      "properties": {
        "enabled": {
          "type": "boolean"
        },
        "listener": {
          "type": "string"
        },
        "partitions": {
          "type": "integer"
        },
        "logReplicationFactor": {
          "type": ["integer", "null"]
        },
        "enabledEventTypes": {
          "type": ["array", "null"]
        },
        "excludedTopics": {
          "type": ["array", "null"]
        },
        "excludedPrincipals": {
          "type": ["array", "null"]
        },
        "audit_client_max_buffer_size": {
          "type": "integer"
        },
        "audit_queue_drain_interval_ms": {
          "type": "integer"
        },
        "audit_queue_max_buffer_size_per_shard": {
          "type": "integer"
        }
      }
```

alejandroEsc: Are you saying you want these additional items exposed to the customer. The schema is just to validate the values file.

charts/redpanda/templates/_helpers.tpl
```diff
@@ -582,6 +585,97 @@ return a warning if the chart is configured with insufficient CPU
...
+      {{- if $trustStoreFile -}}
+        {{- $broker_tls = set $broker_tls "truststore_file" $trustStoreFile -}}
+      {{- end -}}
+      {{- if $broker_tls -}}
+        {{- $result = set $result "broker_tls" $broker_tls -}}
+      {{- end -}}
+    {{- end -}}
+  {{- else -}}
+    {{- range $name, $listener := $kafkaService.external -}}
+      {{- if and ( and ( and $listener.port $name (dig "enabled" true $listener)) ( eq (default "sasl" $listener.authenticationMethod) "sasl" )) ( eq $name $auditLogging.listener ) -}}
```
RafalKorepta, line 637:
```suggestion
      {{- if and ( and ( and $listener.port $name (dig "enabled" false $listener)) ( eq (default "sasl" $listener.authenticationMethod) "sasl" )) ( eq $name $auditLogging.listener ) -}}
```

charts/redpanda/templates/_helpers.tpl
```diff
@@ -582,6 +585,97 @@ return a warning if the chart is configured with insufficient CPU
...
+          -}}
+          {{- $brokers = append $brokers $brokerItem -}}
+        {{- end -}}
+        {{- if $brokers -}}
+          {{- $result = set $result "brokers" $brokers -}}
+        {{- end -}}
+        {{- if dig "tls" "enabled" $tlsEnabled $listener -}}
+          {{- $cert := get $tlsCerts $listener.tls.cert -}}
+          {{- if empty $cert -}}
+            {{- fail (printf "Certificate, '%s', used but not defined") -}}
```
RafalKorepta, line 652: ditto

charts/redpanda/templates/_helpers.tpl
```diff
@@ -582,6 +585,97 @@ return a warning if the chart is configured with insufficient CPU
...
+        {{- $broker_tls = set $broker_tls "truststore_file" $trustStoreFile -}}
+      {{- end -}}
+      {{- if $broker_tls -}}
+        {{- $result = set $result "broker_tls" $broker_tls -}}
+      {{- end -}}
+    {{- end -}}
+  {{- else -}}
+    {{- range $name, $listener := $kafkaService.external -}}
+      {{- if and ( and ( and $listener.port $name (dig "enabled" true $listener)) ( eq (default "sasl" $listener.authenticationMethod) "sasl" )) ( eq $name $auditLogging.listener ) -}}
+        {{- range $address := (include "seed-server-list" $root | mustFromJson) }}
```
RafalKorepta, line 638: I know other clients configuration (schema_registry and pandaproxy) have seed server list, but we excluded seed server list from statefulset annotation config-map-hash. I don't believe using chart-testing we could verify that downscaling is not triggering full rollout.

RafalKorepta: https://github.com/redpanda-data/helm-charts/pull/692

and

https://github.com/redpanda-data/helm-charts/issues/690

charts/redpanda/templates/_configmap.tpl
```diff
@@ -90,6 +90,36 @@ bootstrap.yaml: |
   {{- if and (not (hasKey .Values.config.cluster "storage_min_free_bytes")) ((include "redpanda-atleast-22-2-0" . | fromJson).bool) }}
   storage_min_free_bytes: {{ include "storage-min-free-bytes" . }}
   {{- end }}
+{{/* AUDIT LOGS */}}
+{{- if (include "redpanda-atleast-23-3-0" . | fromJson).bool }}
+  {{- if and ( dig "enabled" "false" .Values.auditLogging ) (include "sasl-enabled" $root | fromJson).bool }}
+  audit_enabled: true
+  audit_client_max_buffer_size: {{ .Values.auditLogging.clientMaxBufferSize | default 16777216 }}
```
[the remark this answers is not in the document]
alejandroEsc: oh i havent seen that, its an issue that has been fixed before, ill look into it.

charts/redpanda/templates/_configmap.tpl
```diff
@@ -90,6 +90,36 @@ bootstrap.yaml: |
...
+  audit_queue_drain_interval_ms: {{ .Values.auditLogging.queueDrainIntervalMs | default 500 }}
+  audit_queue_max_buffer_size_per_shard: {{ .Values.auditLogging.maxBufferSizePerShard | default 1048576 }}
+  audit_log_num_partitions: {{ .Values.auditLogging.partitions | default 12 }}
```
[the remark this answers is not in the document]
alejandroEsc: I do not necessarily agree with this. How do we allow dials for customers for these situations, should i just ignore them if the defaults are set?

charts/redpanda/ci/96-audit-logging-values.yaml.tpl
```diff
@@ -0,0 +1,34 @@
...
+        mechanism: SCRAM-SHA-512
+
+# This should be changed to 23.3.0 or removed when we finally release this version for testing.
+image:
+  repository: docker.redpanda.com/redpandadata/redpanda-unstable
+  tag: v23.3.1-rc3
+
+auditLogging:
+  enabled: true
+  listeners: default
```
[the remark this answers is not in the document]
alejandroEsc: Yes, but i rather test external as this one fails when we do not have a license in place.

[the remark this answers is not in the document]
alejandroEsc: Required also to have license.

charts/redpanda/templates/_helpers.tpl
```diff
@@ -582,6 +585,97 @@ return a warning if the chart is configured with insufficient CPU
...
+      -}}
+      {{- $brokers = append $brokers $brokerItem -}}
+    {{- end -}}
+    {{- if $brokers -}}
+      {{- $result = set $result "brokers" $brokers -}}
+    {{- end -}}
+    {{- if dig "tls" "enabled" $tlsEnabled $kafkaService -}}
+      {{- $cert := get $tlsCerts $kafkaService.tls.cert -}}
+      {{- if empty $cert -}}
+        {{- fail (printf "Certificate, '%s', used but not defined") -}}
```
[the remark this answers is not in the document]
alejandroEsc: right, still going to have it fail.

charts/redpanda/templates/_helpers.tpl
alejandroEsc: OH man thank you, i completely forgot this.

alejandroEsc: This seems to exist in the configmap and may also be a bigger issue: 

for example:
```
  schema_registry_client:
    brokers:
      {{- range (include "seed-server-list" $root | mustFromJson) }}
    - address: {{ . }}
      port: {{  $kafkaService.port }}
      {{- end }}
```

which occurs in lines
https://github.com/redpanda-data/helm-charts/blob/5d9a78033a29f6d36bf44877574151b523f39b10/charts/redpanda/templates/_configmap.tpl#L326
https://github.com/redpanda-data/helm-charts/blob/5d9a78033a29f6d36bf44877574151b523f39b10/charts/redpanda/templates/_configmap.tpl#L425

Not sure why we did not capture this before. I will check the hash but I think i would like to fix this in another PR if you do not mind.

charts/redpanda/templates/_configmap.tpl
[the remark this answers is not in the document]
RafalKorepta: I don't understand what `dials for customers` mean. As John mentioned it should be not set at all. Redpanda is responsible to handle any configuration. In extreme case only `audit_enabled: true` and `audit_log_client` should be set. With latest changes provided it is hard to maintain with every line if statement. Can we create more convenient for us configuration design were range loop could go through any additional audit configuration option?

```
auditLogging:
  enabled: true
  configuration:
    client_max_buffer_size: 888
    queue_drain_interval_ms: 777
    queue_max_buffer_size_per_shard: 666
  listeners: default
```

In the above design we would need to prefix key with `audit_`. The simple range will be open for extension when Redpanda would add new configuration option for audit logs.

charts/redpanda/ci/96-audit-logging-values.yaml.tpl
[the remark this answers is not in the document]
RafalKorepta: Why not both `default` and `internal`?

charts/redpanda/templates/_helpers.tpl
```diff
@@ -582,6 +585,97 @@ return a warning if the chart is configured with insufficient CPU
...
+      {{- if $trustStoreFile -}}
+        {{- $broker_tls = set $broker_tls "truststore_file" $trustStoreFile -}}
+      {{- end -}}
+      {{- if $broker_tls -}}
+        {{- $result = set $result "broker_tls" $broker_tls -}}
+      {{- end -}}
+    {{- end -}}
+  {{- else -}}
+    {{- range $name, $listener := $kafkaService.external -}}
+      {{- if and $listener.port $name (dig "enabled" true $listener) ( eq (default "sasl" $listener.authenticationMethod) "sasl" ) ( eq $name $auditLogging.listener ) -}}
```
RafalKorepta, line 637:
```suggestion
      {{- if and $listener.port $name (dig "enabled" false $listener) ( eq (default "sasl" $listener.authenticationMethod) "sasl" ) ( eq $name $auditLogging.listener ) -}}
```

I don't understand why the default is `true`?

charts/redpanda/templates/_helpers.tpl
RafalKorepta: Still not fixed the `%s` doesn't have argument.

charts/redpanda/templates/_helpers.tpl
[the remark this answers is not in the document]
RafalKorepta: Still not fixed the %s doesn't have argument.

charts/redpanda/templates/_helpers.tpl
RafalKorepta: I'm not sure if this is correct to introduce something that shouldn't be there, to then work on proper fix.

Never the less go for another PR.

charts/redpanda/templates/connectors/connectors.yaml
```diff
@@ -50,12 +50,18 @@ limitations under the License.
   {{ $consoleSASLConfig = cat $consoleSASLConfig " echo $CONNECT_SASL_PASSWORD > /opt/kafka/connect-password/rc-credentials/password;" }}
   {{ $consoleSASLConfig = cat $consoleSASLConfig " exec /opt/kafka/bin/kafka_connect_run.sh" }}
   {{ $command = append $command $consoleSASLConfig }}
+
+  {{ $extraVolumes = concat $extraVolumes .Values.connectors.storage.volume }}
```
RafalKorepta, line 54: I'm not sure why this `concat` is added in this PR. How that relates to audit logging?

charts/redpanda/templates/connectors/connectors.yaml
```diff
@@ -50,12 +50,18 @@ limitations under the License.
...
+
   {{ $extraVolumes = append $extraVolumes (dict
     "name" (printf "%s-users" (include "redpanda.fullname" .))
     "secret" (dict
       "secretName" .Values.auth.sasl.secretRef
     )
   )}}
+
+  {{ $extraVolumeMounts = concat $extraVolumeMounts .Values.connectors.storage.volumeMounts }}
```
RafalKorepta, line 63: ditto

charts/redpanda/templates/connectors/connectors.yaml
RafalKorepta: It was added in https://github.com/redpanda-data/helm-charts/pull/941. I don't understand why Github shows this change..

charts/redpanda/templates/_configmap.tpl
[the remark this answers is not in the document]
michael-redpanda: My vote would be to allow a user access to these configs.  How full these audit buffers get is very dependent on how the customer is using Redpanda.  The audit system tries its best to be conservative when it comes to the usage of the buffers (grouping the 'same' audit events together using a count field).  However, distinct audit events will still generate new messages and use up more memory reserved for these buffers.

This is also dependent on which events the user wants to audit (`audit_enabled_event_types`).  Enabling more of these event types will generate more audit events.  The defaults we selected are reasonable, but if there is a lot of traffic to different topics from different users, the audit buffers may fill up and this will cause Redpanda to start rejecting commands (either through broker_not_available errors on Kafka API or 500 errors on admin API).

Our suggestion for customers is to monitor the metrics and, if necessary, increase the size of the buffers.

charts/redpanda/values.schema.json
```diff
@@ -1552,6 +1552,41 @@
...
+        },
+        "listener": {
+          "type": "string"
+        },
+        "partitions": {
+          "type": "integer"
+        },
+        "logReplicationFactor": {
+          "type": "integer"
+        },
```
michael-redpanda, line 1570: This type should be `["integer", "null"]` with `null` being the default.

charts/redpanda/templates/_helpers.tpl
alejandroEsc: My point is that it already is there, did you not see the lines?

charts/redpanda/templates/connectors/connectors.yaml
alejandroEsc: rebased from another PR.

charts/redpanda/values.schema.json
alejandroEsc: oh thank you, removed.

charts/redpanda/templates/_helpers.tpl
alejandroEsc: because it is. It is usually commented out and assumed to be enabled. The override exists in case you wish to enable it, or disable it on purpose.

charts/redpanda/ci/96-audit-logging-values.yaml.tpl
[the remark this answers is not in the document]
alejandroEsc: you can only use one listener. I chose default since it fails consistently when license does not exist, which internal always worked.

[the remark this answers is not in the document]
alejandroEsc: If your question is about testing the feature, we are only testing that we can functionally get things to work. Not to test the feature itself so one time showing things work is more than enough for us.

charts/redpanda/templates/_configmap.tpl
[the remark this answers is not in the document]
alejandroEsc: No, i would prefer not to do this. It would cause the confusion we are seeing with the configurations we find in tiered storage, where customers ask why there are configurations in two or more places and how they would interact with each other. Instead I gated the default values when applied to the configuration to resolve the "do not set defaults" concern you bring up (which i do not agree with). Remember our job here is to set `production` defaults for customers and those are not always the default values as defined by core. Given i do not know what they are i left them as the default values for the moment.

charts/redpanda/templates/_configmap.tpl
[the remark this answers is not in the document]
alejandroEsc: I am not seeing that    

```
helm template redpanda charts/redpanda/ -n redpanda  --set image.repository=docker.redpanda.com/redpandadata/redpanda-unstable --set image.tag=v23.3.1-rc2 --set auditLogging.enabled=true --set auditLogging.enabled=true --set auth.sasl.enabled=true --set auditLogging.clientMaxBufferSize=16777215 | grep  audit_client_max_buffer_size    audit_client_max_buffer_size: 16777215
      audit_client_max_buffer_size: 16777215
```

charts/redpanda/ci/96-audit-logging-values.yaml.tpl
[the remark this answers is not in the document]
RafalKorepta: > you can only use one listener. I chose default since it fails consistently when license does not exist, which internal always worked.

I didn't know that. Let's leave it.

> Not to test the feature itself so one time showing things work is more than enough for us.

Ok, but we don't have new test in https://github.com/redpanda-data/helm-charts/tree/main/charts/redpanda/templates/tests.

charts/redpanda/templates/_helpers.tpl
RafalKorepta: > My point is that it already is there, did you not see the lines?

I know that schema registry and pandaproxy client is using the seed server. That was my first observation in the first comment 

https://github.com/redpanda-data/helm-charts/pull/934#discussion_r1431995422

From my point of view if we introduce third place where seed server list is part of the checksum, then we will have more work to exclude 3 places not 2. I'm aware of the problem existence even if you do the work to exclude audit_log_client, that's I'm ok with pushing it to the next PR, but not closing the whole effort.

charts/redpanda/ci/96-audit-logging-values.yaml.tpl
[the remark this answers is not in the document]
alejandroEsc: Nothing much we can test, anytime something failed the chart failed to deploy. Any other test would be for functionality. If you want i can write a test to check that we do have the right topics, yeah i can do that.

charts/redpanda/templates/_helpers.tpl
alejandroEsc: Fixed it, now we have the right configmaps and 
```
helm template redpanda charts/redpanda/ -n redpanda  --set image.repository=docker.redpanda.com/redpandadata/redpanda-unstable --set image.tag=v23.3.1-rc2 --set auditLogging.enabled=true --set auditLogging.listener=internal --set auth.sasl.enabled=true --set auditLogging.clientMaxBufferSize=16777215 --set statefulset.replicas=3 | grep config.redpanda.com/checksum
        config.redpanda.com/checksum: 8a0c69ef04db448f5b9136443b7d933b63a22a36fc87f2842343f86bfd6cb493
```

```
helm template redpanda charts/redpanda/ -n redpanda  --set image.repository=docker.redpanda.com/redpandadata/redpanda-unstable --set image.tag=v23.3.1-rc2 --set auditLogging.enabled=true --set auditLogging.listener=internal --set auth.sasl.enabled=true --set auditLogging.clientMaxBufferSize=16777215 --set statefulset.replicas=6 | grep config.redpanda.com/checksum
        config.redpanda.com/checksum: 8a0c69ef04db448f5b9136443b7d933b63a22a36fc87f2842343f86bfd6cb493
```

RafalKorepta: Please see my suggestion https://github.com/redpanda-data/helm-charts/pull/934#discussion_r1432908552

charts/redpanda/templates/_helpers.tpl
```diff
@@ -597,14 +597,14 @@ return a warning if the chart is configured with insufficient CPU
   {{- $trustStoreFile := "" -}}
   {{- $requireClientAuth := dig "tls" "requireClientAuth" false $kafkaService -}}
   {{- if and ( eq "internal" $auditLogging.listener ) ( eq (default "sasl" $kafkaService.authenticationMethod) "sasl" ) -}}
-    {{- range $address := (include "seed-server-list" $root | mustFromJson) }}
+    {{- range $id, $item := $root.tempConfigMapServerList }}
```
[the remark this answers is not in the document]
alejandroEsc: Not needed the intent is to use that tempConfigMapServerList as we do in the case of including rpk-config-internal. This totally is the same thing and the output is the same also. We defer calculation to the point were we included the configmap-content-no-seed, in the no seed-list version it does not produce anything as we intend.

charts/redpanda/values.yaml
```diff
@@ -87,6 +87,30 @@ license_secret_ref: {}
...
+  # -- Enable or disable audit logging, for production clusters we suggest you enable,
+  # however, this will only work if you also enable sasl and a listener with sasl enabled.
+  enabled: false
+  # -- Kafka listener name, note that it must have `authenticationMethod` set to sasl
+  # 'internal' if using internal listener, else use external listener name, e.g., default.
+  listener: internal
+  # -- Integer value defining the number of partitions used by a newly created audit topic.
+  partitions: 12
+  # -- Event types that should be captured by audit logs.
+  enabledEventTypes: ["management", "authenticate"]
```
michael-redpanda, line 102: FYI `admin` is also a default value for this list.

RafalKorepta (approved):
LGTM

PR #962, opened 2024-01-12
Enable users to add connectors extra environment variables

### LINKED ISSUE #959: Connectors extraEnv values not applying if sasl is enabled
### What happened?

Similar to https://github.com/redpanda-data/helm-charts/issues/940 (handled by @alejandroEsc ).
If "auth.sasl.enabled" is true, then values added via "connectors.deployment.extraEnv" will not be applied.
Once sasl is disabled, all is working.

I expected the additional set env var to be applied

<details>
Minimal Helm values for the redpanda chart, that doesn't add my env var:

```
auth:
  sasl:
    enabled: true
    mechanism: SCRAM-SHA-512
    secretRef: kafka-credentials
    users: []
connectors:
  deployment:
    extraEnv:
      - name: REDPANDA_PLUGINS_TARGET_DIR
        value: /opt/kafka/connect-plugins
  enabled: true
```

Minimal Helm values for the redpanda chart, that does add my env var:

```
auth:
  sasl:
    enabled: false
    mechanism: SCRAM-SHA-512
    secretRef: kafka-credentials
    users: []
connectors:
  deployment:
    extraEnv:
      - name: REDPANDA_PLUGINS_TARGET_DIR
        value: /opt/kafka/connect-plugins
  enabled: true
```
</details>

I am using the operator, but it is unrelated as I've been able to replicate this using pure "helm template" commands as such:
helm template redpanda/redpanda -f values.yaml

Redpanda, Connectors

### Chart Version(s)

<details>

```console
redpanda chart version 5.7.5
```

</details>

### Cloud provider

<details>
Linode
</details>

RafalKorepta:
Using the following helm values I was able to confirm that extra environment variable was added.

```
auth:
  sasl:
    enabled: true
connectors:
  enabled: true
  deployment:
    extraEnv:
    - name: test
      value: some-test
```

Fix #959

### DIFF
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 5.7.5
+version: 5.7.6
 
 # The app version is the default version of Redpanda to install.
 # ** NOTE for maintainers: please ensure the artifacthub image annotation is updated before merging
--- a/charts/redpanda/templates/connectors/connectors.yaml
+++ b/charts/redpanda/templates/connectors/connectors.yaml
@@ -40,6 +40,7 @@ limitations under the License.
 
 {{ $extraVolumes := list }}
 {{ $extraVolumeMounts := list }}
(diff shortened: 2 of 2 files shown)


alejandroEsc (approved):
lgtm

PR #963, opened 2024-01-12
Give a way to set up cloud_storage_secret_key from K8S Secret

### LINKED ISSUE #960: No "secretRef" option for tiered-storage credentials
### What happened?

I was trying to configure tiered-storage, when I noticed the chart requires me to provide the access/secret keys as part of the helm values, which is a big security issue for us (not to mention the fact it complicates GitOps).

I expected there to be an option to provide a "secretRef" to a kubernetes secret containing the credentials (which I can then manage securely), same as there is for SASL, or for the enterprise license.

Nothing to reproduce

_No response_

Redpanda

### Chart Version(s)

<details>

```console
redpanda chart version - 5.7.5
```

</details>

### Cloud provider

<details>
Linode
</details>

RafalKorepta:
Fix #960

### DIFF
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 5.7.6
+version: 5.7.7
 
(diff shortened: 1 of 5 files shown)


alejandroEsc (approved):
you may need to change version, lgtm

PR #969, opened 2024-01-17
Update the Operator docs template

JakeSCahill:
Updates the template to mirror the existing templates for the other charts. Also removes deployment instructions and links to docs instead as a single source of truth.


alejandroEsc (approved):
lgtm

Issue #973, opened 2024-01-18
Clearly document the tiered storage config values.


jrkinley:
### What would you like to be added?

The documentation for `storage.tiered.config.*` simply references the property reference documentation on `docs.redpanda.com`, which itself doesn't provide enough information.

### Why is this needed?

An example of why this is needed. In `values.yaml` the comment for `storage.tiered.config.cloud_storage_api_endpoint` simply points to the link: [cloud_storage_api_endpoint](https://docs.redpanda.com/docs/reference/cluster-properties/#cloud_storage_api_endpoint), which just states:

```
 Optional API endpoint.
```

However, if a user sets this in EKS then the deployment will fail because, as stated in a different document [tiered-storage-configuration-properties](https://docs.redpanda.com/current/manage/tiered-storage/#tiered-storage-configuration-properties):

```
For AWS, this can be left blank. It’s generated automatically using the region and bucket.
```

In this case the documentation isn't complete and this information is hard to find, which results in a sub-par user experience.

PR #974, opened 2024-01-18
#973 clearly document the tiered storage config values

jrkinley:


alejandroEsc (approved):
lgtm, thank you!

charts/redpanda/values.yaml
```diff
@@ -470,40 +470,49 @@ storage:
...
+      # -- AWS or GCP access key (required for AWS and GCS authentication with access keys).
       # See the [property reference documentation](https://docs.redpanda.com/docs/reference/cluster-properties/#cloud_storage_access_key).
       cloud_storage_access_key: ""
-      # -- Required for AWS and GCS authentication with access keys.
+      # -- AWS or GCP secret key (required for AWS and GCS authentication with access keys).
       # See the [property reference documentation](https://docs.redpanda.com/docs/reference/cluster-properties/#cloud_storage_secret_key).
       cloud_storage_secret_key: ""
-      # -- See the [property reference documentation](https://docs.redpanda.com/docs/reference/cluster-properties/#cloud_storage_api_endpoint).
+      # -- AWS or GCP API endpoint.
+      # - For AWS, this can be left blank as it is generated automatically using the bucket and region (e.g. "<bucket>.s3.<region>.amazonaws.com")
```
github-amine-kherbouche, line 492: for AWS it ***should be*** left blank if the region and bucket name are set

github-amine-kherbouche (commented):
my 2 cents

charts/redpanda/values.yaml
RafalKorepta: We have the same wording in our docs https://docs.redpanda.com/current/manage/tiered-storage/#:~:text=For%20AWS%2C%20this%20can%20be%20left%20blank.%20It%E2%80%99s%20generated%20automatically%20using%20the%20region%20and%20bucket

Could you file an issue in https://github.com/redpanda-data/docs/issues? 

Then we would change in both places.

PR #977, opened 2024-01-18
chore: pass image overrides to console subchart

alejandroEsc:
Allow passing image override from top level chart to subchart


chrisseto (approved):
LGTM!

PR #984, opened 2024-01-25
redpanda: don't set `tlsConfig` in ServiceMonitor if `tls.enabled` is false

### LINKED ISSUE #573: Unable to enable monitoring without specifically set TLS to false when TLS is not enable with the original cluster
### What happened?

When the running cluster does not have TLS enabled, when I try to add monitoring and update with helm chart, despite the cluster does not need TLS, it still throws error asking for cert-manager. (Even when I have reuse value set to true)

There are work around that I need to deliberately disable TLS. 

For more detail view this thread:
https://redpandadata.slack.com/archives/C01H6JRQX1S/p1687962811469649

That I don't have to set **TLS enable to false** if my cluster was not configure it orginally. 

<details>

Go to my Killercoda scenario https://killercoda.com/christinalin/scenario/redpanda-k8s-day2

In step one,  run the following:
```
cat <<EOF > redpanda-monitoring.yaml
monitoring:
  enabled: true
  scrapeInterval: 10s
EOF


helm upgrade --install redpanda redpanda/redpanda   --namespace redpanda   --create-namespace   --values redpanda-monitoring.yaml   --reuse-values
```

</details>

_No response_

Redpanda

### Chart Version(s)

<details>
Lastest

</details>

### Cloud provider

<details>
N/A
</details>

chrisseto:
Previously, if `tls.enabled` was false but monitoring was enabled (`monitoring.enabled=true`) the `ServiceMonitor` object would incorrectly render a `tlsConfig` and set `scheme` to `https`.

This was due to the default value of `monitoring.tlsConfig={}` being evaluated as truthy.

This commit removes the `monitoring.tlsConfig` check entirely as there are further checks within the stanza. To mitigate any confusion an note as been added to the values.yml file.

Fixes #573

---

This is a _speculative_ fix. I've realized that I can't reproduce the original problem as described. Though I do not have a cluster on hand to test against so I've just been using `--dry-run='client'`.

`helm template foo . --dry-run='client' --debug --set 'tls.enabled=false,monitoring.enabled=true'` doesn't produce the bugged output and I haven't been able to track down helm's documentation on truthiness vs falseyness. Just yet.

### DIFF
--- a/charts/redpanda/templates/servicemonitor.yaml
+++ b/charts/redpanda/templates/servicemonitor.yaml
@@ -37,7 +37,7 @@ spec:
     {{- if dig "enableHttp2" "" .Values.monitoring }}
     enableHttp2: .Values.monitoring.enableHttp2
     {{- end }}
-  {{- if or .Values.tls.enabled .Values.monitoring.tlsConfig }}
+  {{- if .Values.tls.enabled }}
     scheme: https
     tlsConfig:
       {{- if  dig "tlsConfig" dict .Values.monitoring }}
--- a/charts/redpanda/values.yaml
+++ b/charts/redpanda/values.yaml
@@ -308,6 +308,7 @@ monitoring:
   # Enables http2 for scraping metrics for prometheus. Used when Istio's mTLS is enabled and using tlsConfig.
   # enableHttp2: true
   tlsConfig: {}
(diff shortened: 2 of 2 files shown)


chrisseto:
Agreed @RafalKorepta , closing until we can successfully reproduce the linked issue.

PR #985, opened 2024-01-26
Add ability to overwrite volumeClaimTemplate tiered storage name

RafalKorepta:
In migration from Cluster custom resource (previous operator v1) to Redpanda custom resource (new version operator v2) there was problem as tiered storage was named `shadow-index-cache`. Due to that StatefulSet does not want rollout new PodSpec as Volume would change. This change allow to overwrite and unblocks migration.


chrisseto (approved):
LGTM!

PR #989, opened 2024-01-28
Added support for setting `cloud_storage_azure_shared_key`

missingcharacter:
I noticed no way to use an environment variable to set `cloud_storage_azure_shared_key` and saw that `configurator.sh` only configures `cloud_storage_secret_key` from an environment variable.

I thought it was a good idea to use the same logic but toggle between `cloud_storage_secret_key` and `cloud_storage_azure_shared_key,` AFAIK `cloud_storage_azure_shared_key` is mutually exclusive with `cloud_storage_secret_key`


charts/redpanda/templates/secrets.yaml
```diff
@@ -351,10 +351,14 @@ stringData:
...
+    {{- $cloudStorageSecretKey := "cloud_storage_secret_key" }}
+    {{- if .Values.storage.tiered.credentialsSecretRef.azure }}
+      {{- $cloudStorageSecretKey = "cloud_storage_azure_shared_key" }}
+    {{- end }}
     {{- if and .Values.storage.tiered.credentialsSecretRef.name .Values.storage.tiered.credentialsSecretRef.key }}
     set +x
-    echo Setting cloud_storage_secret_key configuration
-    rpk redpanda config --config "$CONFIG" set cloud_storage_secret_key $CLOUD_STORAGE_SECRET_KEY
+    echo Setting {{ $cloudStorageSecretKey }} configuration
+    rpk redpanda config --config "$CONFIG" set {{ $cloudStorageSecretKey }} $CLOUD_STORAGE_SECRET_KEY
```
RafalKorepta, line 361: Looking at the history of Redpanda the configuration key could be changed for any other S3 equivalent storage.

I would suggest
```
    credentialsSecretRef:
      configurationKey: cloud_storage_secret_key
      name: XYZ
      key: ABC
```

That way you would be able to change that for azure. 

BTW We have opened issue for adding access key to this script too.

missingcharacter: @RafalKorepta how about now?

RafalKorepta (approved):

missingcharacter:
@RafalKorepta should I change something before this can be merged?

RafalKorepta:
I'm waiting for confirmation that CLA is signed. It's confusing as bot already posted 2 comments.

alejandroEsc:
@RafalKorepta could we create a ticket or update the crd in this case to account for the change in schema?

missingcharacter:
When I click on `CLA Not Signed yet` I get this page saying I already signed it

image


I can close this PR and make a new one

RafalKorepta:
Turns out there is small bug in CLA bot

RafalKorepta:
> @RafalKorepta could we create a ticket or update the crd in this case to account for the change in schema?

I need to update CRD for other changes I will include this too.

RafalKorepta:
@alejandroEsc https://github.com/redpanda-data/redpanda-operator/pull/75

PR #995, opened 2024-01-30
Add ability to reference secret and access keys as Kubernetes Secret

### LINKED ISSUE #981: tiered-storage credentials access key from secretRef
### What happened?

Following up on https://github.com/redpanda-data/helm-charts/issues/960, I am looking for the tiered-storage access key to also be read from a secretRef.

Our current flow is:
1. Create access key and push it to secret manager (via terraform)
2. Use operator to create k8s secrets from secret manager secrets
3. Use secretRefs to point to the pre-existing secrets

This is not so fluid considering the access key needs to be set manually in the value currently.
I would expect to replace this:

```
      tiered:
        config:
          cloud_storage_access_key: xxxx
        credentialsSecretRef:
          name: redpanda-tiered-storage-secret
          key: SECRET_KEY
```

With perhaps something like this:
```
      tiered:
        credentialsSecretRef:
          accessKey:
            name: redpanda-tiered-storage-secret
            key: ACCESS_KEY
          secretKey:
            name: redpanda-tiered-storage-secret
            key: SECRET_KEY
```

Nothing to reproduce

_No response_

Redpanda, Operator

### Chart Version(s)

<details>

```console
redpanda-5.7.9
```

</details>

### Cloud provider

<details>
Linode
</details>

RafalKorepta:
This change is backwards compatible.

Fix #981

### DIFF
--- a/charts/redpanda/values.yaml
+++ b/charts/redpanda/values.yaml
@@ -459,12 +459,25 @@ storage:
       # -- Additional annotations to apply to the created PersistentVolumeClaims.
       annotations: {}
 
-    # credentialsSecretRef can be used to set cloud_storage_secret_key from
+    # credentialsSecretRef can be used to set `cloud_storage_secret_key` and/or `cloud_storage_access_key` from
     # referenced Kubernetes Secret
     credentialsSecretRef:
-      configurationKey: cloud_storage_secret_key
+      accessKey:
+        # https://docs.redpanda.com/current/reference/cluster-properties/#cloud_storage_access_key
+        configurationKey: cloud_storage_access_key
+        # name:
+        # key:
+      secretKey:
+        # https://docs.redpanda.com/current/reference/cluster-properties/#cloud_storage_secret_key
+        # or
+        # https://docs.redpanda.com/current/reference/cluster-properties/#cloud_storage_azure_shared_key
+        configurationKey: cloud_storage_secret_key
+        # name:
+        # key
+      # -- DEPRECATED `configurationKey`, `name` and `key`. Please use `accessKey` and `secretKey`
+      # configurationKey: cloud_storage_secret_key
       # name:
-      # key
+      # key:
     #
     # -- Tiered Storage settings
     # Requires `enterprise.licenseKey` or `enterprised.licenseSecretRef`
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 5.7.16
+version: 5.7.17
 
 # The app version is the default version of Redpanda to install.
 # ** NOTE for maintainers: please ensure the artifacthub image annotation is updated before merging
--- a/charts/redpanda/README.md
+++ b/charts/redpanda/README.md
@@ -3,7 +3,7 @@
 description: Find the default values and descriptions of settings in the Redpanda Helm chart.
 ---
 
-![Version: 5.7.14](https://img.shields.io/badge/Version-5.7.14-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v23.3.3](https://img.shields.io/badge/AppVersion-v23.3.3-informational?style=flat-square)
+![Version: 5.7.17](https://img.shields.io/badge/Version-5.7.17-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: v23.3.4](https://img.shields.io/badge/AppVersion-v23.3.4-informational?style=flat-square)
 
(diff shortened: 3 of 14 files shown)


charts/redpanda/values.yaml
```diff
@@ -459,12 +459,25 @@ storage:
       # -- Additional annotations to apply to the created PersistentVolumeClaims.
       annotations: {}
 
-    # credentialsSecretRef can be used to set cloud_storage_secret_key from
+    # credentialsSecretRef can be used to set `cloud_storage_secret_key` and/or `cloud_storage_access_key` from
```
chrisseto, line 462: This way of setting values is quite confusing to me. Wouldn't it be more ergonomic, both for code and documentation, to allow the tiered_storage settings to be either a static value or a secret ref object?

Better yet, could we just expose the envvar configuration and let users control `valueFrom`?

chrisseto (approved):
Just a non-blocking question

charts/redpanda/values.yaml
RafalKorepta: Redpanda has 3 different type of configuration.
1. [Cluster](https://docs.redpanda.com/current/reference/cluster-properties/)
2. [Node](https://docs.redpanda.com/current/reference/node-properties/)
3. [Tunable](https://docs.redpanda.com/current/reference/tunable-properties/)

## Restart required

If user would change node configuration that would require full rollout for the StatefulSet controller. The checksum would need to change every time the referenced secret changes 👇  https://github.com/redpanda-data/helm-charts/blob/a4d526f692c11311270582d75c23cb222513484c/charts/redpanda/templates/statefulset.yaml#L57

## Admin API config changes

The tunable and cluster configuration does not need to have full rollout, but post-update job needs to be re-triggered 👇 
https://github.com/redpanda-data/helm-charts/blob/a4d526f692c11311270582d75c23cb222513484c/charts/redpanda/templates/post-upgrade.yaml#L73

User would need to run `rpk cluster import` every time the secret content changes or re-run `post-upgrade` job.

## Dynamic configuration

Few configuration properties that relay on the Pod being scheduled (e.g. Pod ordinal) will be adjusted in initContainer 👇 
https://github.com/redpanda-data/helm-charts/blob/a4d526f692c11311270582d75c23cb222513484c/charts/redpanda/templates/statefulset.yaml#L144-L172

The script is defined 👇 
https://github.com/redpanda-data/helm-charts/blob/a4d526f692c11311270582d75c23cb222513484c/charts/redpanda/templates/secrets.yaml#L288-L359 

## Size of the Volume

The `storage_min_free_bytes` depends on `storage.persistentVolume.size` which is not part of the configuration, but dictates that buffer for Redpanda configuration, so that in the end we don't see alloc failures.  

## Changes to K8S resources

We can support such configuration, but some places that uses configuration properties changes other K8S resource like listed below 👇 :

- https://github.com/redpanda-data/helm-charts/blob/a4d526f692c11311270582d75c23cb222513484c/charts/redpanda/templates/statefulset.yaml#L486
- https://github.com/redpanda-data/helm-charts/blob/a4d526f692c11311270582d75c23cb222513484c/charts/redpanda/templates/statefulset.yaml#L459 the `(include "storage-tiered-config" .|fromJson).cloud_storage_enabled` part
- https://github.com/redpanda-data/helm-charts/blob/a4d526f692c11311270582d75c23cb222513484c/charts/redpanda/templates/statefulset.yaml#L385
- https://github.com/redpanda-data/helm-charts/blob/a4d526f692c11311270582d75c23cb222513484c/charts/redpanda/templates/statefulset.yaml#L373
- https://github.com/redpanda-data/helm-charts/blob/a4d526f692c11311270582d75c23cb222513484c/charts/redpanda/templates/statefulset.yaml#L298
- https://github.com/redpanda-data/helm-charts/blob/a4d526f692c11311270582d75c23cb222513484c/charts/redpanda/templates/statefulset.yaml#L254
- https://github.com/redpanda-data/helm-charts/blob/a4d526f692c11311270582d75c23cb222513484c/charts/redpanda/templates/statefulset.yaml#L166
- https://github.com/redpanda-data/helm-charts/blob/a4d526f692c11311270582d75c23cb222513484c/charts/redpanda/templates/statefulset.yaml#L126
- https://github.com/redpanda-data/helm-charts/blob/a4d526f692c11311270582d75c23cb222513484c/charts/redpanda/templates/_configmap.tpl#L616
- https://github.com/redpanda-data/helm-charts/blob/a4d526f692c11311270582d75c23cb222513484c/charts/redpanda/templates/_helpers.tpl#L334
- https://github.com/redpanda-data/helm-charts/blob/a4d526f692c11311270582d75c23cb222513484c/charts/redpanda/templates/post-upgrade.yaml#L74-L93
- https://github.com/redpanda-data/helm-charts/blob/a4d526f692c11311270582d75c23cb222513484c/charts/redpanda/templates/_configmap.tpl#L190-L192
- https://github.com/redpanda-data/helm-charts/blob/a4d526f692c11311270582d75c23cb222513484c/charts/redpanda/templates/_configmap.tpl#L90-L92

Issue #996, opened 2024-01-30
Redpanda: configuration changes may result in a cluster wide restart


chrisseto:
### What happened?

Reported from an [internal slack thread](https://redpandadata.slack.com/archives/C01H6JRQX1S/p1706523420474149), certain configuration changes (`values.yaml`) may result in a rolling restart of Redpanda pods. This is most notable when scaling the STS up or down.

### What did you expect to happen?

When it's possible to avoid a rolling restart, we absolutely should.

To start, let's say that scaling a cluster up or down should not result in other Pods restarting.

### How can we reproduce it (as minimally and precisely as possible)?. Please include values file.

N/A

### Which are the affected charts?

Redpanda, Operator

### Chart Version(s)

<details>

```console
$ helm -n <redpanda-release-namespace> list 
# paste output here
```

</details>

### Cloud provider

<details>
N/A
</details>

chrisseto:
I'm betting this has to do with the [statefulset's pod template containing the hash of the red panda configuration](https://github.com/redpanda-data/helm-charts/blob/a4d526f692c11311270582d75c23cb222513484c/charts/redpanda/templates/statefulset.yaml#L57). The configuration (AFAIK) will change anytime replicas scale up or down.

PR #814, opened 2023-10-20
change the checksum to roll pods when the advertized hostnames change

joejulian:
Fixes #813

Issue #813, opened 2023-10-20
Updating `external.domain` doesn't update advertised listeners


vuldin:
### What happened?

Ran the following command:
```
helm upgrade --install redpanda redpanda --repo https://charts.redpanda.com/ -n redpanda --create-namespace --wait --timeout 1h -f values-1.yaml
```
With this values.yaml:
```
tls:           
  enabled: false         
statefulset:   
  initContainers:
    setDataDirOwnership:
      enabled: true
#external:     
#  domain: local
```

Then I uncommented `external.domain` and re-ran the command.

### What did you expect to happen?

Externally advertised listeners should have been updated to include the domain but they weren't:
```
> k exec -it -n redpanda redpanda-0 -c redpanda -- rpk cluster info --brokers localhost:9094
CLUSTER
=======
redpanda.519d6b1b-d680-4c60-b321-3634b812e5d6

BROKERS
=======
ID    HOST        PORT
0*    redpanda-0  31092
1     redpanda-1  31092
2     redpanda-2  31092

TOPICS
======
NAME      PARTITIONS  REPLICAS
_schemas  1           3
```

### How can we reproduce it (as minimally and precisely as possible)?. Please include values file.

<details>

```console
$ helm get values <redpanda-release-name> -n <redpanda-release-namespace> --all
# paste output here
```

</details>

### Anything else we need to know?

Advertised listeners is set in the initContainer, so when updated there is no way for the deployment to know that pods need to be rolled (as this is based on updates to the ConfigMap). But advertised listeners depends on runtime data, so it can't be moved to the ConfigMap without other changes.

### Which are the affected charts?

Redpanda

### Chart Version(s)

<details>

```console
> helm -n redpanda list                     
NAME            NAMESPACE       REVISION        UPDATED                                 STATUS          CHART           APP VERSION
redpanda        redpanda        1               2023-10-20 11:53:58.656755902 -0500 CDT deployed        redpanda-5.6.24 v23.2.12
```

</details>

### Cloud provider

<details>
minikube (self-hosted)
</details>

PR #998, opened 2024-01-31
Allow single template for all Redpanda Pods external advertise addresses

RafalKorepta:
The implementation of external addresses was done for static list. In some environments there could be automation developed to register DNS name dynamically based on certain property of a Pod like:
* The K8S Node IP that can be hash out
* Pod ordinal

With the combination of `external` `domain`, `prefixTemplate` and single entry in `addresses` list the full StatefulSet rollout wouldn't be needed when user scale up or down the cluster.

## Reference

https://github.com/redpanda-data/helm-charts/blob/75ec4bb892124a9e8a71b911e3a920f227a8e7d7/charts/redpanda/templates/statefulset.yaml#L57
https://github.com/redpanda-data/helm-charts/pull/814
https://github.com/redpanda-data/helm-charts/issues/813
https://github.com/redpanda-data/helm-charts/issues/996

### DIFF
--- a/charts/redpanda/values.yaml
+++ b/charts/redpanda/values.yaml
@@ -267,6 +267,8 @@ external:
   # The number of brokers is defined in statefulset.replicas.
   # The values can be IP addresses or DNS names.
   # If external.domain is set, the domain is appended to these values.
+  # There is an option to define a single external address for all brokers and leverage
+  # prefixTemplate as it will be calculated during initContainer execution.
   # addresses:
   # - redpanda-0
   # - redpanda-1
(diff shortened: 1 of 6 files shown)


charts/redpanda/values.yaml
```diff
@@ -267,6 +267,8 @@ external:
   # The number of brokers is defined in statefulset.replicas.
   # The values can be IP addresses or DNS names.
   # If external.domain is set, the domain is appended to these values.
+  # There is option to define single external address for all brokers and leverage
+  # prefixTemplate as it will calculated during initContainer execution.
```
chrisseto, line 271: Grammar nit:
```suggestion
  # There is an option to define a single external address for all brokers and leverage
  # prefixTemplate as it will be calculated during initContainer execution.
```

chrisseto (approved):
LGTM! Thanks for adding the test case.

PR #1001, opened 2024-02-01
Fix volume name reference when name is overwritten

RafalKorepta:
After setting `.storage.persistentVolume.nameOverwrite` the volume that
references `persistentVolumeClaim` should be adjusted too, because StatefulSet
controller rightfully reports an event:
```
0/6 nodes are available: 6 persistentvolumeclaim "tiered-storage-dir" not found. preemption: 0/6 nodes are available: 6 Preemption is not helpful for scheduling.
```


chrisseto (approved):
LGTM!

PR #1004, opened 2024-02-02
Fix volume definition when volumeClaimTemplate is used

RafalKorepta:
Dedicated PesistentVolumeClaim will be created by the volumeClaimTemplate definition inside StatefulSet and reference will be added to Pod volume list.

Currently Pods are getting none existent reference to PersistentVolumeClaim if storage.tiered.mountType is set to `persistentVolume`.
```
volumes:
  - name: shadow-index-cache
    persistentVolumeClaim:
      claimName: shadow-index-cache-rp-cmue6bgoajp9du05hqvg-2
  - name: datadir
    persistentVolumeClaim:
      claimName: datadir-rp-cmue6bgoajp9du05hqvg-2
  - name: tiered-storage-dir
    persistentVolumeClaim:
      claimName: shadow-index-cache
```

### DIFF
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 5.7.21
+version: 5.7.22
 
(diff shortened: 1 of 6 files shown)


chrisseto (approved):
LGTM!

PR #1018, opened 2024-02-08
redpanda: set tiered storage configuration in a batch

chrisseto:
Previously, tiered storage settings where being set via the `bootstrap.yml` and via an init container.

Only non-secret values could be set via `bootstrap.yml` which could result in validation errors as `cloud_storage_enabled` requires `cloud_storage_{access,secret}_key` to be set.

Secret settings set via the init container could fail in newly configured clusters as quorum had yet to be established.

This commit moves all tiered storage configuration to the post-install-upgrade-job. It passes secret and non-secret values via environment variables and then stitches them into a yaml file suitable for `rpk cluster config import`. This ensures that dependent configurations "see" their dependents and that the cluster is configured post quorum/bootstrap.

### DIFF
--- a/charts/redpanda/templates/_helpers.tpl
+++ b/charts/redpanda/templates/_helpers.tpl
@@ -869,25 +869,79 @@ REDPANDA_SASL_USERNAME REDPANDA_SASL_PASSWORD REDPANDA_SASL_MECHANISM
   {{- end -}}
 {{- end -}}
 
-{{- define "storage-tiered-credentials-secret-key" -}}
-{{- $oldCondtion := (and .Values.storage.tiered.credentialsSecretRef.name .Values.storage.tiered.credentialsSecretRef.key) -}}
-{{- $newCondtion := (and .Values.storage.tiered.credentialsSecretRef.secretKey.name .Values.storage.tiered.credentialsSecretRef.secretKey.key) -}}
-{{- $configurationKey := (dig "configurationKey" "" .Values.storage.tiered.credentialsSecretRef) -}}
-{{- if empty $configurationKey -}}
-  {{- $configurationKey = .Values.storage.tiered.credentialsSecretRef.secretKey.configurationKey -}}
+{{/* secret-ref-or-value
+        in: {Value: string?, SecretKey: string?, SecretName: string?}
+        out: corev1.Envvar | nil
+    secret-ref-or-value converts a set of values into a structure suitable for
+    use as an environment variable or nil.
+*/}}
+{{- define "secret-ref-or-value" -}}
+    {{- if and (empty .Value) (or (empty .SecretName) (empty .SecretKey)) -}}
+        {{- mustToJson nil -}}
+    {{- else -}}
+        {{- $out := (dict
+            "name" .Name
+            "value" .Value
+            "valueFrom" (dict
+                "secretKeyRef" (dict
+                    "name" .SecretName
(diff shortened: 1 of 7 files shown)


chrisseto:
Hmm, so tests are failing due to `test-kafka-nodelete` being able to delete a topic but there's no explanation as to why we shouldn't be able to delete said topic.

@RafalKorepta, you mentioned that CI is currently failing. Is this the issue you mentioned or did I break something uniquely?

charts/redpanda/templates/_helpers.tpl
```diff
@@ -869,25 +869,79 @@ REDPANDA_SASL_USERNAME REDPANDA_SASL_PASSWORD REDPANDA_SASL_MECHANISM
...
+            {{/* iterate over the sorted keys of $config for deterministic output */}}
+            {{- range $i, $key := ($config | keys | sortAlpha) -}}
+                {{- $value := (get $config $key) -}}
+
+                {{/* Special case for cache size */}}
+                {{- if eq $key "cloud_storage_cache_size" -}}
+                    {{- $value = (include "SI-to-bytes" $value | int64) -}}
+                {{- end -}}
+
+                ,
```
RafalKorepta, line 933: I think there is problem with comma (`,`). After `RPK_CLOUD_STORAGE_ACCESS_KEY` env var there might be issue of missing comma.

charts/redpanda/templates/_helpers.tpl
```diff
@@ -869,25 +869,79 @@ REDPANDA_SASL_USERNAME REDPANDA_SASL_PASSWORD REDPANDA_SASL_MECHANISM
...
+        {{- else -}}
+            {{- $_ := unset $out "valueFrom" -}}
+        {{- end -}}
+        {{- mustToJson $out -}}
+    {{- end -}}
+{{- end -}}
+
+{{- define "tiered-storage-env-vars" -}}
+    {{- $config := (include "storage-tiered-config" . | fromJson) -}}
+    [
```
RafalKorepta, line 903: Maybe it would be easier to create list of corev1.Envvar using `prepend` or `append` and then pipe it to `toJson` to not deal with commas.

RafalKorepta (commented):
`redpanda.yaml` has the same tiered config, which might state in conflict with this change. As far as I understand we will set tiered configuration only in one place (job in post-install/post-upgrade).

https://github.com/redpanda-data/helm-charts/blob/0a81aa9882c579f4249b347c7fad8b88d2f81d68/charts/redpanda/templates/_configmap.tpl#L387-L401

RafalKorepta:
The problem is https://github.com/redpanda-data/helm-charts/blob/0a81aa9882c579f4249b347c7fad8b88d2f81d68/charts/redpanda/values.yaml#L971 configuration is not applied in post-install job.

The problem current job/test-case `test-kafka-nodelete` is that bash doesn't print anything, but when I enabled printing commands (`set -x`) then I realised
```
kubectl logs redpanda-x189n49tug-test-kafka-nodelete
++ rpk cluster config get default_topic_replications
+ timeout 120 bash -c 'until [[ 1 = 3 ]]; do sleep 1; done'
```

`timeout` bash function is returning exist code `124`, because bash process after 120 seconds times out.

chrisseto:
Closing in favor of https://github.com/redpanda-data/helm-charts/pull/1054

PR #1022, opened 2024-02-11
skip generating unused resources if issuerRef is set

reytech-dev:
Hi,
I added one change avoiding the generation of not used/usable issuers and certificates if a custom issuer is provided.
I added a dedicated values.yaml with a working test configuration.
But I'm missing the required custom cluster issuer: "custom-internal-issuer-ref"
Where can I add this cluster issuer to be available during the test run?
Regards

### DIFF
--- a/charts/redpanda/templates/cert-issuers.yaml
+++ b/charts/redpanda/templates/cert-issuers.yaml
@@ -79,4 +79,5 @@ spec:
     group: cert-manager.io
       {{- end }}
   {{- end }}
+  {{- end }}
 {{- end }}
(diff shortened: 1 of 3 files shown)


charts/redpanda/templates/cert-issuers.yaml
```diff
@@ -79,4 +79,5 @@ spec:
     group: cert-manager.io
       {{- end }}
   {{- end }}
+  {{- end }}
```
RafalKorepta, line 82: I think that will not work as `{{- end }}` is already present at 42 line

https://github.com/redpanda-data/helm-charts/blob/61a8558032862f5d52c79fca8e3fa15063dd5ea5/charts/redpanda/templates/cert-issuers.yaml#L42

PR #1023, opened 2024-02-11
avoid empty dnsNames if custom issuerRef is set

reytech-dev:
Hi,
after turning off the "external"-attribute in the redpanda chart, I ran into an issue because the certificate got created with no "dnsNames". 
The proposed change always adds "dnsNames" to the certificates if no external domains are provided.
Please let me know if there are further changes required or anything is not right with my changes.
Regards


charts/redpanda/Chart.yaml
```diff
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 5.7.23
```
RafalKorepta, line 26: I'm confused why patch version was bumped +2, but not +1?

charts/redpanda/ci/20-certs-dns-names-values.yaml
```diff
@@ -0,0 +1,26 @@
+
```
RafalKorepta, line 1: This file name should be `charts/redpanda/ci/19-certs-dns-names-values.yaml`, right?

charts/redpanda/ci/20-certs-dns-names-values.yaml
```diff
@@ -0,0 +1,26 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+---
+tls:
+  enabled: true
+  certs:
+    default:
+      issuerRef:
+        name: custom-internal-issuer-ref
+        kind: ClusterIssuer
```
RafalKorepta, line 23: The test will always fail as CI doesn't creates external issuer. It would require injecting `ClusterIssuer` in to template folder. That will only test `helm install`, but upgrade test would not work at all (`helm upgrade`) as `chart-testing` downloads previous version from the upstream.

RafalKorepta:
Redpanda chart is now moved to https://github.com/redpanda-data/redpanda-operator/tree/main/charts/redpanda. If this feature is missing please rise an issue or PR against new code.

PR #1025, opened 2024-02-12
charts/kminion - allow setting priorityClassName for kminion pod

Vince-Chenal:
Hey,

I've been using [cloudworkz/kafka-minion-helm-chart](https://github.com/cloudworkz/kafka-minion-helm-chart) and I was able to set `priorityClassName` on pods, which is not the case in this Chart.

Tell me if you expect anything else in the PR 🙏

Also I would need a new version to be released as soon as this PR is merged, should I change the version in `Chart.yaml` myself?

### DIFF
--- a/charts/kminion/templates/daemonset.yaml
+++ b/charts/kminion/templates/daemonset.yaml
@@ -47,6 +47,9 @@ spec:
       securityContext:
         {{- toYaml .Values.podSecurityContext | nindent 8}}
       serviceAccountName: {{ include "kminion.serviceAccountName" . }}
+      {{- if .Values.priorityClassName }}
+      priorityClassName: {{ .Values.priorityClassName }}
+      {{- end }}
       volumes:
         - name: config
           configMap:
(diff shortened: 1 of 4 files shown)


charts/kminion/templates/daemonset.yaml
```diff
@@ -47,6 +47,9 @@ spec:
       securityContext:
         {{- toYaml .Values.podSecurityContext | nindent 8}}
       serviceAccountName: {{ include "kminion.serviceAccountName" . }}
+      {{- if .Values.priorityClassName }}
```
RafalKorepta, line 50: NIT: I'm more used to 
```suggestion
      {{- wiith .Values.priorityClassName }}
```

RafalKorepta (approved):

charts/kminion/templates/daemonset.yaml
Vince-Chenal: done :+1:

Vince-Chenal:
I'm not sure to understand the `check-values` check. Yes I'm adding a new root-level value
Am I doing anything wrong?

RafalKorepta:
> I'm not sure to understand the check-values check. Yes I'm adding a new root-level value
Am I doing anything wrong?

No, it's fine. The check was invented for us to be always aware that values are changing, so that we should carefully check if such change is backward compatible. In your case, all is good.

Vince-Chenal:
It seems that I won't be able to update branch, wait for checks approval and then merge :sweat_smile:

PR #1033, opened 2024-02-21
Set container ports only if they are configured for external exposure

### LINKED ISSUE #1029: External Ports configured in StatefulSet even if disabled
### What happened?

The following configurations are set in the values.yaml to prevent that ports are exposed externally:

```
external:
  service:
    enabled: false
  enabled: false

listeners:
  admin:
    external:
      default:
        enabled: false
```

When templating the helm chart the StatefulSet nevertheless still configures the admin-default port for the redpanda container. 
```
ports:
- name: admin
  containerPort: 9644
- name: admin-default
  containerPort: 9645
```

The same happens for the other listeners as well.

Expected that the external listeners which are not needed are not configured as ports for the redpanda container in the StatefulSet

<details>

external:
  service:
    enabled: false
  enabled: false

listeners:
  admin:
    external:
      default:
        enabled: false

</details>

The problem is probably line 286 in redpanda/templates/statefulset.yaml:

```
{{- if $external.port }}
```

This should probably be something like:
```
{{- if or ($external.enabled) (dig "externalDns" "enabled" true $values.external) }}
```

_No response_

### Chart Version(s)

<details>

5.7.23

</details>

### Cloud provider

<details>
Azure
</details>

may-cDev:
The idea is:

- If the listener is explicitly configured for external port, then set it (setting listener.external.default.enabled). No matter what is configured for external.enabled, as it's stated in the docs that this value overrides the global external.enabled if set

- If external.enabled is set to true, then also add those container ports if those are not explicitly disabled (listener.external.default.enabled = false). If the listener setting is not set or set to true as well, then add it. If the listener setting is set to false, then do not set it.

Closes #1029

### DIFF
--- a/charts/redpanda/templates/statefulset.yaml
+++ b/charts/redpanda/templates/statefulset.yaml
@@ -283,7 +283,7 @@ spec:
             - name: {{ lower $name }}
               containerPort: {{ $listener.port }}
   {{- range $externalName, $external := $listener.external }}
-  {{- if or ($external.enabled) (and ($values.external.enabled) (not (eq $external.enabled false))) }}
+  {{- if and $external.port (or $external.enabled (and $values.external.enabled (dig "enabled" true $external))) }}
             - name: {{ lower $name | trunc 6 }}-{{ lower $externalName | trunc 8 }}
               containerPort: {{ $external.port }}
     {{- end }}
--- a/charts/redpanda/templates/statefulset.yaml
+++ b/charts/redpanda/templates/statefulset.yaml
@@ -283,7 +283,7 @@ spec:
             - name: {{ lower $name }}
               containerPort: {{ $listener.port }}
   {{- range $externalName, $external := $listener.external }}
-    {{- if $external.port }}
+  {{- if or ($external.enabled) (and ($values.external.enabled) (not (eq $external.enabled false))) }}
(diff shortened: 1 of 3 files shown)


charts/redpanda/templates/statefulset.yaml
```diff
@@ -283,7 +283,7 @@ spec:
             - name: {{ lower $name }}
               containerPort: {{ $listener.port }}
   {{- range $externalName, $external := $listener.external }}
-    {{- if $external.port }}
+  {{- if or ($external.enabled) (and ($values.external.enabled) (not (eq $external.enabled false))) }}
```
RafalKorepta, line 286: Thank you for the fix!

My only concern is that previous check is not longer there.

The following helm values would generate invalid configuration:

```yaml
listeners:
  admin:
    external:
      test:
        enabled: true
```

Can the check be combined?
```suggestion
  {{- if and $external.port (or $external.enabled (and $values.external.enabled (dig "enabled" true $external))) }}
```

P.S. 
I removed the parentheses.
I replaced `not (eq $external.enabled false)` with `dig` as it's easer to understand.

REF:
https://helm.sh/docs/chart_template_guide/function_list/#dig

RafalKorepta (approved):

PR #1038, opened 2024-02-21
feat(redpanda): add option to disable tests

may-cDev:
In our GitOps approach we use "helm template" and store the resulting yaml file of the helm chart. When using helm template also the test yaml definitions are included, which we don't need for our use case.

Therefore, aligning to https://github.com/redpanda-data/helm-charts/pull/1028 , adding an option to disable the tests


charts/redpanda/templates/tests/test-auditLogging.yaml
```diff
@@ -18,7 +18,7 @@
   This feature is gated by having a license, and it must have sasl enabled, we assume these conditions are met
   as part of setting auditLogging being enabled.
 */}}
-{{- if and .Values.auditLogging.enabled (include "redpanda-atleast-23-3-0" . | fromJson).bool }}
+{{- if and (.Values.tests.enabled) (and .Values.auditLogging.enabled (include "redpanda-atleast-23-3-0" . | fromJson).bool) }}
```
RafalKorepta, line 21: NIT:

`and` function can work with multiple arguments

 ```suggestion
{{- if and .Values.tests.enabled .Values.auditLogging.enabled (include "redpanda-atleast-23-3-0" . | fromJson).bool }}
```

RafalKorepta (approved):
LGTM

RafalKorepta (changes requested):
Due to latest merge the Chart version needs to be bumped. 

The patch version should be +1.

RafalKorepta (approved):

PR #1039, opened 2024-02-21
Issue 990/fix broker tls volumes

RafalKorepta:
A copy of https://github.com/redpanda-data/helm-charts/pull/1016 with fixed sleep in test


charts/connectors/templates/tests/01-mm2-values.yaml
```diff
@@ -143,6 +143,8 @@ spec:
 
           curl {{ template "curl-options" . }} -H 'Content-Type: application/json' http://{{ include "connectors.serviceName" . }}:{{ .Values.connectors.restPort }}/connectors -d @/tmp/mm2-conf.json
 
+          sleep 30
```
chrisseto, line 146: Let's add a comment here about why we're sleeping before merging. If there's a known error message, it would be good to include it in the comment as well.

charts/connectors/ci/02-broker-tls-values.yaml
```diff
@@ -0,0 +1,28 @@
...
+    enabled: true
+    ca:
+      secretRef: redpanda-default-cert
+    cert:
+      secretRef: redpanda-default-cert
+    key:
+      secretRef: redpanda-default-cert
+
+logging:
+  level: trace
```
chrisseto, line 28: Can we add a newline to the end of this file before merging?

chrisseto (approved):
LGTM!

charts/connectors/ci/02-broker-tls-values.yaml
RafalKorepta: Yes, I will add that

charts/connectors/templates/tests/01-mm2-values.yaml
RafalKorepta: Good idea I will add short explanation.

PR #1041, opened 2024-02-21
Fix link in Console doc template

JakeSCahill:


RafalKorepta (approved):

RafalKorepta:
Any change to file not related to chart will not trigger Github action workflow. This project has set up required steps like `lint` and `summary`, which will not be triggered and will block merging.

PR #1042, opened 2024-02-21
Pass console values to subchart deployment template

RafalKorepta:
If deployment is generated by Redpanda chart, then a lot fields where not passed
from parent (Redpanda) helm chart to its sub-chart Console.


chrisseto (approved):
LGTM 😬 we'll need to get some linting on the README.md's soon...

PR #1043, opened 2024-02-21
Add a MVP `go test` harness

chrisseto:
This commit contains an MVP of a `go test` based testing harness for helm charts. This includes thin clients for Kubernetes and Helm.

The single test `TestRedpandaChart` installs cert-manager and redpanda into whatever cluster `KUBECONFIG` is pointing at. The default installation of redpanda and the default cluster created by kind are incompatible due to resource mismatches. `NoWait{ForJobs}` are both set to `true` as redpanda will never become ready.

Future work will include automatic spinning up and destruction of kind clusters, forwards and backwards compatibility tests, and helm "stories" that upgrade a cluster across a series of values files.


pkg/helm/helm.go
```diff
@@ -0,0 +1,264 @@
...
+	Name            string
+	Namespace       string
+	Values          map[string]any
+	Version         string
+	NoWait          bool
+	NoWaitForJobs   bool
+}
+
+func (o *InstallOptions) asFlags() ([]string, error) {
+	valuesFile, err := os.CreateTemp(os.TempDir(), "helm-values")
```
RafalKorepta, line 124: NIT: The temporary file should be deleted after each test.

main_test.go
```diff
@@ -0,0 +1,65 @@
...
+		}
+	}
+
+	if err := client.RepoAdd(ctx, "jetstack", "https://charts.jetstack.io"); err != nil {
+		return err
+	}
+
+	return client.Install(ctx, "jetstack/cert-manager", helm.InstallOptions{
+		Name:            "cert-manager",
+		Version:         "v1.11.0",
```
RafalKorepta, line 32: NIT: Latest cert-manager is `v1.14.2`

pkg/helm/helm.go
```diff
@@ -0,0 +1,264 @@
...
+		return err
+	}
+
+	var result map[string]any
+	if err := json.Unmarshal([]byte(stdout), &result); err != nil {
+		return err
+	}
+
+	// TODO: Do somethign with the results
+	fmt.Printf("Result:\n%#v\n", result)
```
RafalKorepta, line 207: NIT: Totally agree. The output is currently unreadable with `go test -v ./...`. I would drop it for the moment.

pkg/helm/helm.go
chrisseto: Going to punt on this in this PR. A future commit will handle placing that within the tempDir from `testing.T`

main_test.go
chrisseto: Fixed.

pkg/helm/helm.go
chrisseto: Removed for now!

PR #1048, opened 2024-02-26
redpanda: correct types of some tunables

chrisseto:
Prior to this commit, these values had written defaults that included their units which led to them being parsed as strings. This would cause the post-upgrade/install job(s) to fail as `rpk` would reject the values.

This commit corrects the default values of `log_retention_ms` and `group_initial_rebalance_delay` and adds type checking to `values.schema.json`.

### DIFF
--- a/charts/redpanda/values.schema.json
+++ b/charts/redpanda/values.schema.json
@@ -1523,7 +1523,12 @@
           "type": "object"
         },
         "tunable": {
-          "type": "object"
+            "type": "object",
+            "additionalProperties": true,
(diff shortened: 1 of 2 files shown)


RafalKorepta:
I'm not sure if we can have part of the schema for `tunable` configuration. Currently I see
```
redpanda:
invalid character '"' after object key:value pair
```

We would need to have full schema for tunable which in my opinion is not possible as we would like to be open for Redpanda extensions.

RafalKorepta (approved):

PR #1050, opened 2024-02-26
Add a workflow for non-chart changes

chrisseto:
This repository is configured to require "lint" and "summary" jobs to pass. The existing workflows only run for chart specific changes. This leaves any non-chart changes stuck in limbo as jobs will never run.

This commit adds in a (fragile) github action that can be used to pass CI for any non-chart change. When a PR is stuck in limbo, the author may update the new workflow to explicitly watch for changes to the file they made.


RafalKorepta (approved):
I hope this will unblock us in the future

chrisseto:
> Funny that at this point in time we don't have go files and CI is stuck.

Hmm, I was expecting the watch on this workflow file itself to trigger the runs 🤔

PR #1054, opened 2024-02-27
redpanda: set tiered storage configuration in a batch

chrisseto:
Previously, tiered storage settings where being set via the `bootstrap.yml` and via an init container.

Only non-secret values could be set via `bootstrap.yml` which could result in validation errors as `cloud_storage_enabled` requires `cloud_storage_{access,secret}_key` to be set.

Secret settings set via the init container could fail in newly configured clusters as quorum had yet to be established.

This commit moves all tiered storage configuration to the post-install-upgrade-job. It passes secret and non-secret values via environment variables and then stitches them into a yaml file suitable for `rpk cluster config import`. This ensures that dependent configurations "see" their dependents and that the cluster is configured post quorum/bootstrap.

To mitigate a race condition with the post-upgrade job, this commit additionally swaps the hook weights. This ensures that the `rpk cluster config import` call in the post-upgrade job does not wipe any configurations set by the post-install job. This solution is less than ideal and we'll need to consolidate all cluster configuration changes into a single job in the near future.

### DIFF
--- a/charts/redpanda/templates/_helpers.tpl
+++ b/charts/redpanda/templates/_helpers.tpl
@@ -869,25 +869,79 @@ REDPANDA_SASL_USERNAME REDPANDA_SASL_PASSWORD REDPANDA_SASL_MECHANISM
   {{- end -}}
 {{- end -}}
 
-{{- define "storage-tiered-credentials-secret-key" -}}
-{{- $oldCondtion := (and .Values.storage.tiered.credentialsSecretRef.name .Values.storage.tiered.credentialsSecretRef.key) -}}
-{{- $newCondtion := (and .Values.storage.tiered.credentialsSecretRef.secretKey.name .Values.storage.tiered.credentialsSecretRef.secretKey.key) -}}
-{{- $configurationKey := (dig "configurationKey" "" .Values.storage.tiered.credentialsSecretRef) -}}
-{{- if empty $configurationKey -}}
-  {{- $configurationKey = .Values.storage.tiered.credentialsSecretRef.secretKey.configurationKey -}}
+{{/* secret-ref-or-value
+        in: {Value: string?, SecretKey: string?, SecretName: string?}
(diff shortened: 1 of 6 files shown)


RafalKorepta (approved):
LGTM

RafalKorepta:
Interesting that ct lint doesn't complain about missing Redpanda `Chart.yaml` version bump.

RafalKorepta:
I see now why lint is not complaining. With nix changes https://github.com/redpanda-data/helm-charts/commit/7d851b941678a8a8002f7425d0520e89f7559655 we are using upstream chart-testing which does not lint correctly with current setup. I opened PR that is not being picked up.

### Reference

https://github.com/joejulian/chart-testing/commit/2cf55ae9e56f07f7af16f5282e133f95cd19a332

https://github.com/helm/chart-testing/pull/594

charts/redpanda/templates/_helpers.tpl
```diff
@@ -869,25 +869,79 @@ REDPANDA_SASL_USERNAME REDPANDA_SASL_PASSWORD REDPANDA_SASL_MECHANISM
...
+            {{/* iterate over the sorted keys of $config for deterministic output */}}
+            {{- range $i, $key := ($config | keys | sortAlpha) -}}
+                {{- $value := (get $config $key) -}}
+
+                {{/* Special case for cache size */}}
+                {{- if eq $key "cloud_storage_cache_size" -}}
+                    {{- $value = (include "SI-to-bytes" $value | int64) -}}
+                {{- end -}}
+
+                ,
```
RafalKorepta, line 933: I think there is problem with comma (`,`). After `RPK_CLOUD_STORAGE_ACCESS_KEY` env var there might be issue of missing comma.

chrisseto:
As mentioned in DMs. Merging this without the version bump or linting fix. We'll address those later.

PR #1057, opened 2024-02-27
Chris/helmtest

chrisseto:

PR #1056, opened 2024-02-27
Upgrade `pkg/helm`

chrisseto:
### This PR is stacked on top of others!

Please consider reviewing those PRs first. This message will be removed once prerequisite PRs are merged and this one is rebased.

* f126545423ce018f4821d3b7570d3f34f5a524a9 is from #1054
* 8798ce0f5cd72ee07b70f28ebe574207c2ae0a15 is from #1055

---

#### 63533ad5a94a070c019c65efb425a44969cbdae8 Upgrade `pkg/helm`

This commit upgrades to use a more standardized way of converting
structs into CLI flags ala `ToFlags`. It also implements a few more helm
function.

PR #1055, opened 2024-02-27
Add kubernetes client

chrisseto:
### This PR is stacked on top of others!

Please consider reviewing those PRs first. This message will be removed once prerequisite PRs are merged and this one is rebased.

* f126545423ce018f4821d3b7570d3f34f5a524a9 is from #1054

---

#### 8798ce0f5cd72ee07b70f28ebe574207c2ae0a15 Add kubernetes client

This commit implements a light weight Kubernetes in `pkg/kube`.
Said client is designed to mimic the UX of `kubectl` in a way that is
ergonomic within go.

It additionally comes with some experimental generic helper functions.

PR #1058, opened 2024-02-27
move redpanda tests to charts/redpanda

chrisseto:
### This PR is stacked on top of others!

Please consider reviewing those PRs first. This message will be removed once prerequisite PRs are merged and this one is rebased.

* f126545423ce018f4821d3b7570d3f34f5a524a9 is from #1054
* 8798ce0f5cd72ee07b70f28ebe574207c2ae0a15 is from #1055
* 63533ad5a94a070c019c65efb425a44969cbdae8 is from #1056
* fb35d875346f3a6799e8548c4361002f2c5b2dd8 is from #1057

---

#### be826d66ad33e74ed98769a181629155369d6b22 move redpanda tests to charts/redpanda

This commit moves the go based tests for the redpanda chart from
`main_test.go` to `charts/redpanda` to better match developer
expectations.

It additionally adds in partial go structs for modeling `values.yaml`
and implements an `rpk`-esque client to ease writing assertions against
cluster configuration values and the like.


charts/redpanda/values.go
```diff
@@ -0,0 +1,40 @@
+package redpanda
+
+type Values struct {
```
RafalKorepta, line 3: I wonder if `Values` structure could generate our `values.yaml` definition with `values.schema.json`. Or vice versa to keep all places in sync.

RafalKorepta (approved):

charts/redpanda/values.go
chrisseto: Absolutely! I'm not sure which direction the generation should go IE go -> yaml or yaml -> go? We'll want to ensure that documentation is generated appropriately for everything though.

PR #1071, opened 2024-03-02
Do not call rpk cluster import twice

RafalKorepta:
If cluster has sasl enabled that is realised by external Kubernetes secret the post-upgrade job (`-post-upgrade` name suffix) can clean configuration like superuser list. As the post-install and post-upgrade (`-configuration` name suffix) job is doing correct configuration drift correction this change will remove rpk cluster import call from post-upgrade job.


chrisseto (approved):
I thought very hard about removing this line but was worried it might cause something to break. 

LGTM if you're confident that it's okay to remove and tests pass!

PR #1073, opened 2024-03-04
redpanda: Add timeouts to curl exec probes

chrisseto:
Thanks to digging from Rafal, we've discovered that exec probes in kubernetes implicitly get an additional 2 minute timeout [1]. This results in abnormally long startup times for redpanda as each Pod's startup probe can hang for up to 2 minutes.

This commit adds a timeout of 5 seconds to each exec using curl via the `-m` flag which prevents the 2 minute hang time.

For posterity, 5 seconds was chosen arbitrarily.

[1]: https://github.com/kubernetes/kubernetes/blob/54f9807e1e84981b2053f4daf779f5ed19962144/pkg/kubelet/cri/remote/remote_runtime.go#L475-L478

### DIFF
--- a/charts/redpanda/templates/statefulset.yaml
+++ b/charts/redpanda/templates/statefulset.yaml
@@ -223,7 +223,7 @@ spec:
                 - -c
                 - |
                   set -e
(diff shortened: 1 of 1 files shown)


RafalKorepta (approved):
LGTM, if you want this to be released it would require chart version bump.

chrisseto:
This is more of a development QoL change. I think it's fine to have it ride along the next release :)

PR #1079, opened 2024-03-08
redpanda: implement snapshot testing

chrisseto:
This commit implements golden/snapshot based testing for the redpanda chart by leveraging all the test files in `ci` and placing the resultant YAML into `testdata`.

It includes a minor fix to `_configmap.tpl` that caused some non-determinism. For other sources, `--set` flags are specified to work around them.

### DIFF
--- a/charts/redpanda/chart_test.go
+++ b/charts/redpanda/chart_test.go
@@ -2,6 +2,7 @@ package redpanda
 
 import (
 	"os"
+	"os/exec"
 	"testing"
 
 	"github.com/redpanda-data/helm-charts/pkg/helm"
@@ -19,6 +20,11 @@ func TieredStorageStatic(t *testing.T) Values {
 	}
 
 	return Values{
+		Config: &Config{
+			Node: map[string]any{
+				"developer_mode": true,
+			},
+		},
 		Enterprise: &Enterprise{
 			License: os.Getenv("REDPANDA_LICENSE"),
 		},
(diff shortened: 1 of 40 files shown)


charts/redpanda/chart_test.go
```diff
@@ -74,6 +84,39 @@ func TieredStorageSecretRefs(t *testing.T, secret *corev1.Secret) Values {
 	}
 }
 
+func TestTemplate(t *testing.T) {
+	ctx := testutil.Context(t)
+	client, err := helm.New(helm.Options{ConfigHome: testutil.TempDir(t)})
```
RafalKorepta, line 89: If possible I would like to replace/inject the `charts/redpanda/Chart.lock` file with the following content

```
dependencies:
- name: console
  repository: https://charts.redpanda.com
  version: 0.7.21
- name: connectors
  repository: https://charts.redpanda.com
  version: 0.1.10
digest: sha256:020e143876af41c4cee1fe908bcd07151a4489115b0aa2100728a6ccc117671c
generated: "2024-03-11T15:50:33.503683+01:00"
```

I would recommend to perform `helm dep up` command before the for loop. That way tests would be deterministic.

chrisseto: Fixed by running `helm dep build` in this test. We don't want this test to perform updates for us just to ensure that `charts` is in sync with `Chart.lock`.

alejandroEsc (approved):
lgtm

PR #1085, opened 2024-03-13
Add security context for statefulset and post-upgrade/install jobs

tzahigro1:
This PR adds the option to configure "runAsNonRoot" and "allowPrivilegeEscalation" fields under "securityContext" for the statefulset , post-install-upgrade-job and post-upgrade job  inside the values file under statefulset.securityContext.allowPrivilegeEscalation and statefulset.securityContext.runAsNonRoot

### DIFF
--- a/.github/workflows/nightly_version_checks.yaml
+++ b/.github/workflows/nightly_version_checks.yaml
@@ -56,7 +56,7 @@ jobs:
           echo steps.chartversion.outputs.CHANGED
       - name: create PR
         if: steps.chartversion.outputs.CHANGED == 'true'
-        uses: peter-evans/create-pull-request@a4f52f8033a6168103c2538976c07b467e8163bc  # v6.0.1
+        uses: peter-evans/create-pull-request@70a41aba780001da0a30141984ae2a0c95d8704e  # v6.0.2
         with:
           title: ${{ steps.chartversion.outputs.TITLE }}
           commit-message: ${{ steps.chartversion.outputs.TITLE }}
--- a/.github/workflows/nightly_version_checks.yaml
+++ b/.github/workflows/nightly_version_checks.yaml
@@ -56,7 +56,7 @@ jobs:
           echo steps.chartversion.outputs.CHANGED
       - name: create PR
         if: steps.chartversion.outputs.CHANGED == 'true'
-        uses: peter-evans/create-pull-request@b1ddad2c994a25fbc81a28b3ec0e368bb2021c50  # v6.0.0
+        uses: peter-evans/create-pull-request@a4f52f8033a6168103c2538976c07b467e8163bc  # v6.0.1
         with:
           title: ${{ steps.chartversion.outputs.TITLE }}
           commit-message: ${{ steps.chartversion.outputs.TITLE }}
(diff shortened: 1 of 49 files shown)


RafalKorepta (approved):
Hi @tzahigro1,

I don't have problem with that implementation, but there are dedicated values for post jobs

https://github.com/redpanda-data/helm-charts/blob/90d3dbb6dac5e741e48c4b7a811a13fd862138bd/charts/redpanda/values.yaml#L568-L608

which could have that securityContext part. 

It would be good, as a follow up, to give users ability to change the securityContext per container. 

Could you bump the following semver? The patch version should suffice.
https://github.com/redpanda-data/helm-charts/blob/90d3dbb6dac5e741e48c4b7a811a13fd862138bd/charts/redpanda/Chart.yaml#L26

Could you signed CLA?

RafalKorepta:
I see that CLA is signed. I will bump Chart.yaml quickly.

tzahigro1:
> Hi @tzahigro1,
> 
> I don't have problem with that implementation, but there are dedicated values for post jobs
> 
> https://github.com/redpanda-data/helm-charts/blob/90d3dbb6dac5e741e48c4b7a811a13fd862138bd/charts/redpanda/values.yaml#L568-L608
> 
> which could have that securityContext part.
> 
> It would be good, as a follow up, to give users ability to change the securityContext per container.
> 
> Could you bump the following semver? The patch version should suffice.
> 
> https://github.com/redpanda-data/helm-charts/blob/90d3dbb6dac5e741e48c4b7a811a13fd862138bd/charts/redpanda/Chart.yaml#L26
> 
> Could you signed CLA?

Yes I saw that but upon looking at the previous implementation the "runAsUser" and "runAsGroup" values come from there so I figured its the place to add it. 
I agree with you that a better more preferred way is to add them to the values under each "job" , although it can conflict with the current implementation because I see that there are 3 resources using the same "container-security-context" values from helpers.tpl

tzahigro1:
for example in the existing _helpers.tpl
https://github.com/tzahigro1/redpanda-helm-charts/blob/b71a4578678e0587a27abaa30f55b152b9ae2d67/charts/redpanda/templates/_helpers.tpl#L477
You can see that "container-security-context" already takes the value from ".Values.statefulset.securityContext.runAsUser" and ".Values.statefulset.securityContext.runAsGroup" 
and it is used inside the post-upgrade job aswell as in the post-install-upgrade-job and statefulset.
I think this value from "_helpers.tpl" either need to be merged with the values inside "values.yaml" file or some other implementation.
@RafalKorepta Please let me know what you think

RafalKorepta:
@tzahigro1 I agree with you that this needs to be refactored. I will create follow up to address that.

RafalKorepta (commented):
Great to see that improvement! 

You can always extend the values.yaml with commented example and defines new fields in [`values.schema.json`](https://github.com/redpanda-data/helm-charts/blob/fb9de95981a0d4463194e39d3c3dbbf519d4e2b4/charts/redpanda/values.schema.json)

RafalKorepta (approved):

tzahigro1:
The CLA is bugged because I made commits from user without proper email address format , I re-wrote the commit history with my correct email. 
@RafalKorepta

EliranTurgeman:
@alejandroEsc , @chrisseto  please review :)

tzahigro1:
I see that the PR is merged but don't see any new chart released?
Can you please make sure the new chart is released and let us know.
@RafalKorepta @chrisseto

RafalKorepta:
@tzahigro1 The chart.yaml was not changed. Probably by rebasing the semver bump was missed. 

If CI passes for https://github.com/redpanda-data/helm-charts/pull/1087, then your changes would be included in new chart release.

PR #1091, opened 2024-03-18
upgrade to dyff for YAML diffing

chrisseto:
This commit upgrades `testutil.AssertGolden` to utilize the `dyff` tool
for diffing YAML files. `dyff` appropriately handles multi-document YAML
files (Like the output of `helm template`) and produces much more human
friendly outputs as they include the full path of the difference.


alejandroEsc (approved):
lgtm

Issue #1097, opened 2024-03-19
Chart-testing github action is flakey or broken


chrisseto:
The fork of `chart-testing` that we're using appears to be breaking but only on certain build steps: https://github.com/redpanda-data/helm-charts/actions/runs/8347611627/job/22854783792?pr=1089

chrisseto:
Hmm, the plot thickens. For some reason https://github.com/redpanda-data/helm-charts/pull/1089 successfully automerged but I can't trackdown a passing run of the actions in github's Actions tab. When you expand the details of the merge, it does indicate that there was a successful run however.

chrisseto:
AH. Got it. There's a collision of names now. Since there was gocode within the referenced PR it ran the non-Chart action which passed. Github saw that there was an action with the name "summary" and "lint" and decided that was good enough.

This further reinforces my desire to collapse down the workflow definitions to one pull_request action and then perform filtering on changes within the Job(s).

PR #1099, opened 2024-03-20
Add chart-testing and chart-releaser to flake.nix

chrisseto:
This commit adds the necessary nix files to install `chart-releaser`,
which is not available in nixpkgs, and joejuilian's fork of
`chart-testing`.

Doing so paves the way towards replacing unnecessary github actions
(anything that could be replaced with nix and a single shell command)
with nix to cut down on the duplication of dependencies.

This work was motivated by the sudden flakiness of installing github
actions. See also #1097.


RafalKorepta (approved):

chrisseto:
As this is just a change to flake.nix CI isn't being trigger. This was merged via https://github.com/redpanda-data/helm-charts/pull/1100 instead. Thanks to @RafalKorepta for the suggestion!

PR #1100, opened 2024-03-20
replace chart-testing and helm github actions with nix

### LINKED ISSUE #1097: Chart-testing github action is flakey or broken

chrisseto:
# Based on PR #1099

This commit removes all references to joejulian/chart-testing-action and
azure/setup-helm in favor of relying on nix to provide these packages.

There are a few side effects to this commit:
- All github actions are now run through `nix develop --impure --command
  bash`. This may interact strangely with the actions that install a
  dependency (ie kind) but would only result in running the version
  that's provided by nix.
- helm has been upgraded to `v3.14.2`

Fixes #1097.


chrisseto:
Looking good to me. All the failures are due to timeouts trying to install metallb from github or they're whining about the lack of a version bump.

RafalKorepta (approved):

PR #1102, opened 2024-03-20
actions: add minor workflow caching 

chrisseto:
This commit integrates github's cache action and DeterminateSystem's
magic nix cache action into our various workflows. These will cache helm
repositories, which helps reduce flakiness and the nix store which can
shave a couple minutes off all tasks.

Additionally this commit re-works the file filters of the pull_request
github action. Requiring that all paths match resulted in some
inconsistent rule sets. The matching is now back to "one or more" and
relies on being very specific rather than being broad with negations.

### DIFF
--- a/.github/workflows/pull_requests.yaml
+++ b/.github/workflows/pull_requests.yaml
@@ -20,6 +20,12 @@ on:
     branches:
       - "**"
 
+# Cancel old workflows whenever a PR is updated.
+# Thanks to https://stackoverflow.com/questions/66335225/how-to-cancel-previous-runs-in-the-pr-when-you-push-new-commitsupdate-the-curre
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 defaults:
   run:
     shell: nix develop --impure --command bash {0}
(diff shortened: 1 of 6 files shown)


chrisseto:
Looks like this shaves a solid ~2 minutes off each run due to caching the nix store!

I'll see if we can get a dedicated cachix instance setup as that'll speed us up a little bit more. Caching the go deps via github actions will speed up calling `go mod tidy` as well.

RafalKorepta:
I need to learn more but from
https://github.com/DeterminateSystems/magic-nix-cache-action/blob/a08d2ea91155518e6677c96e3111ac63813b9349/README.md?plain=1#L49-L51
there are 3 steps:

- nix-installer-action
- magic-nix-cache-action
- flake-checker-action

Why flake-checker-action is missing from our definition?

RafalKorepta:
Is it important warning? 
```
warning: error: unable to download 'http://127.0.0.1:37515/2hc9lg18zd6yabw9jqj0wy3s9kyvkzp0.narinfo': HTTP error 429

       response body:

       GitHub API error: API error (42[9](https://github.com/redpanda-data/helm-charts/actions/runs/8543311562/job/23407023403?pr=1102#step:5:10) Too Many Requests): StructuredApiError { message: "Request was blocked due to exceeding usage of resource 'Count' in namespace ''." }; retrying in 252 ms
```

https://github.com/redpanda-data/helm-charts/actions/runs/8543311562/job/23407023403?pr=1102#step:5:8

chrisseto:
> Why flake-checker-action is missing from our definition?

IMO there's just no need from it. The checks it performs aren't that useful to us. As long as `nix develop` works, which is tested by CI, we're good to go.

> Is it important warning?

Nope! It's just a rate limit from github that's [automatically handled by the action](https://github.com/DeterminateSystems/magic-nix-cache-action/blob/main/README.md#usage-notes)

TFTR!

PR #1108, opened 2024-03-22
redpanda: Migrating a subset of _helpers.tpl to golang

chrisseto:
Show casing what using gotohelm looks like and experimenting with some
dev / CI configurations.

### DIFF
--- a/.github/workflows/pull_requests_redpanda.yaml
+++ b/.github/workflows/pull_requests_redpanda.yaml
@@ -74,10 +74,6 @@ jobs:
         with:
           fetch-depth: 0
 
-      - run: |
-          git checkout main
-          git checkout -
-
       - name: compare redpanda values with main
         run: dyff --color=off -k between -s <(git show 'origin/main:charts/redpanda/values.yaml') charts/redpanda/values.yaml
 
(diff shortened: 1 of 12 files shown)


chrisseto:
The existence of `git checkout main && git checkout -` in our workflows are causing some fun problems with `nix develop`.  I'm not sure why those are present and further more it's vexing that they're still being run despite me having removed them within this PR.

chrisseto:
@alejandroEsc @RafalKorepta when you have a spare second, could you pull this branch and run `nix develop` and then `task charts:generate:redpanda && go test ./charts/redpanda` in the resultant subshell? Just want to double check that everything will work as expected on developer machines!

There will be some failures from `go test` as there's a bit of merge skew right now.

chrisseto:
Huh... This looks like the errors I was hitting before making `genpartial`. I'll try to repro locally. Thanks!

chrisseto:
No dice locally... I wonder if we'll want to switch to using `helm build` to get a built version of the chart within our tests. Just to reduce the overall moving parts?

RafalKorepta (approved):

PR #1112, opened 2024-03-26
Consolidation github workflows and run go test for PRs

chrisseto:
Prior to this commit, there existed a workflow file for each chart
within this repository. The introduction of go to the codebase caused
this pattern to become defunct as there would not be a great way to
integrate `go test` into these workflows without massive duplication and
oddly specific `-run` values.

This commit removes the individual workflows (with the exception of
`pull_requests_from_origin`) in favor of a single workflow that
delegates to chart specific workflows.

As linting is important for any change to the codebase and the speed
thereof is largely unaffected by the total number of files or tasks, all
linters are run irrespective of the total diff.

Additionally this commit adds `actionlint`, `go vet ./...`,
generated file linting, and `go test` as steps to CI.


.github/workflows/pull_requests.yaml
```diff
@@ -0,0 +1,132 @@
...
+          charts-redpanda:
+            - 'charts/redpanda/**'
+          charts-console:
+            - 'charts/console/**'
+          charts-connectors:
+            - 'charts/connectors/**'
+          charts-kminion:
+            - 'charts/kminion/**'
+          charts-operator:
+            - 'charts/operator/**'
```
RafalKorepta, line 65: If this is included then lines 31-33 should changed, right? There should not be negation in front of the path.

https://github.com/redpanda-data/helm-charts/pull/1112/files#diff-33cb58679656b227d577808dd88a6d7a9a915d63628459c99181089d3c5b9978R31-R33

RafalKorepta (approved):
LGTM

.github/workflows/pull_requests.yaml
chrisseto: Nice catch, fixed!

chrisseto:
Just a foot note here, there are some holes in the matching rules right now that could allow changes to the workflows and other dependent files to cause breakages in CI. I'm going to merge this as is just because the slowness and flakiness of everything would drag this initial change into oblivion.

I'll update the rules in https://github.com/redpanda-data/helm-charts/pull/1102

PR #1136, opened 2024-04-03
Fix merge function logic in sprig re-write

RafalKorepta:
Helmette merge function does not give the same result as helm. Currently multiple maps can be called.


alejandroEsc (approved):
lgtm, nice tests

pkg/gotohelm/helmette/sprig_test.go
```diff
@@ -34,3 +39,156 @@ func TestDefault(t *testing.T) {
...
+			},
+		}, dict["src1"].(map[string]any), dict["src2"].(map[string]any))
+
+		assert.Equal(t, changeAnyIntToFloat64(expected), ts)
+		assert.Equal(t, expected, dict["dst"])
+		assert.Equal(t, expected, answer)
+	})
+}
+
+func changeAnyIntToFloat64(input map[string]any) map[string]any {
```
chrisseto, line 164: Same nit as the other PR: can this instead use `json.Marshal/json.Unmarshal`? It might be worth making a `JSONRoundtrip(any) map[string]any` function as I suspect we'll be doing this a lot.

pkg/gotohelm/helmette/sprig_test.go
```diff
@@ -34,3 +39,156 @@ func TestDefault(t *testing.T) {
 	require.Equal(t, map[string]any{"default": true}, helmette.Default(map[string]any{"default": true}, map[string]any{}))
 	require.Equal(t, map[string]any{"default": false}, helmette.Default(map[string]any{"default": true}, map[string]any{"default": false}))
 }
+
+func TestMerge(t *testing.T) {
+	t.Run("matching_key", func(t *testing.T) {
```
chrisseto, line 44: super nit: How would you feel about re-writing this as a table driven test?

```go
testCases := []struct{
Name string
Inputs []map[string]any
Output map[string]any
}{
// Cases go here. Notes about what's being exercised can be added as comments here.
}

// Write your test code once
for _, tc := range testCases {
t.Run(tc.Name, func() { ... })
}
```

Doing so makes it a lot easier to add in regression tests and it makes it more clear that you're testing a variety of inputs against the same test code.

pkg/gotohelm/helmette/sprig.go
```diff
@@ -38,8 +39,12 @@ func Keys[K comparable, V any](m map[K]V) []K {
 }
 
 // Merge is a go equivalent of sprig's `merge`.
-func Merge[K comparable, V any](dst, src map[K]V) map[K]V {
-	maps.Copy(dst, src)
+func Merge[K comparable, V any](dst map[K]V, sources ...map[K]V) map[K]V {
```
chrisseto, line 42: Could we also add in a test case or two to the sprig test package in gotohelm to ensure that sprig works the same way as the transpiled version?

chrisseto (approved):
LGTM! No blocking comments.

pkg/gotohelm/helmette/sprig_test.go
RafalKorepta: Sorry, I cherry-pick from other commit and forgot to address that! I agree with you and I will push the change shortly

pkg/gotohelm/helmette/sprig.go
RafalKorepta: Ok, I will add that.

pkg/gotohelm/helmette/sprig_test.go
RafalKorepta: The test case borrowed from sprig will be harder to model in the table driven test fashion, but I will do my best.

chrisseto: Oh gotcha! You're welcome to leave it as is. Nits are always optional.

In theory, it might be fine to just import sprig and re-export their functions in helmette 🤔

RafalKorepta: I feel like test case borrowed from can not be easily added to the table driven tests. I will leave it as stand alone test case.

chrisseto (approved):

PR #1137, opened 2024-04-03
Assert that values does not change

RafalKorepta:
Tanspiler test suite can verify that input values has changed. That helps narrowing down badly written functions in any of the helm chart.

### DIFF
--- a/pkg/gotohelm/transpiler_test.go
+++ b/pkg/gotohelm/transpiler_test.go
@@ -22,7 +22,10 @@ import (
 
 type TestSpec struct {
 	Unsupported bool
-	Values      []map[string]any
+	// ValuesChanged when `false`, the default, will cause the test to assert that
+	// .Values has not be mutated by the chart. Set to `true` to disable.
+	ValuesChanged bool
+	Values        []map[string]any
 }
 
 var testSpecs = map[string]TestSpec{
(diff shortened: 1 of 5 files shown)


alejandroEsc (approved):
lgtm

chrisseto (commented):
I'm not sure if I fully understand what this assertion is checking. Are we asserting that `dot.Values` should not be mutated within helm charts? If so, I don't think this is a super reliable way to test or enforce that. There's a lot of copying of Values happening that might result in false negatives.

If calling `Merge` incorrectly resulted in unexpected mutations, we could instead enforce that `Merge` is always called with a new dictionary as the destination argument?

pkg/gotohelm/transpiler_test.go
```diff
@@ -21,16 +21,23 @@ import (
 )
 
 type TestSpec struct {
-	Unsupported bool
-	Values      []map[string]any
+	Unsupported   bool
+	ValuesChanged bool
```
chrisseto, line 25: Could you add a comment here describing what `ValuesChanged` does? Something like "ValuesChanged, when true, disable the default assertion that the global `.Values` field has not been mutated."

chrisseto (approved):
LGTM!

pkg/gotohelm/transpiler_test.go
RafalKorepta: Maybe I should rename the `AssertForValuesChanges`?

chrisseto: I generally like terse names with good documentation rather than long names :P. How about `NoAssertValuesMutation`? That works with the zero value but double negatives can be a bit confusing.

pkg/gotohelm/transpiler_test.go
```diff
@@ -22,7 +22,9 @@ import (
 
 type TestSpec struct {
 	Unsupported bool
-	Values      []map[string]any
+	// ValuesChanged when `true`, asserts that .Values has been mutated. Otherwise, it will assert that .Values was not mutated.
```
chrisseto, line 25:
```suggestion
	// ValuesChanged when `false`, the default, will cause the test to assert that .Values has not be mutated by the chart. Set to `true` to disable.
```

chrisseto (approved):
LGTM

PR #1145, opened 2024-04-04
fix: make automountServiceAccountToken configurable

stefan-caraiman:
Makes `automountServiceAccountToken` configurable

Defaults to true by default for both ServiceAccount and Deployment resources.


RafalKorepta (approved):
LGTM

PR #1151, opened 2024-04-04
redpanda: add support for overriding default flags

### LINKED ISSUE #1134: 🔹 Don't set duplicate CLI flags if provided by the end user
As it stands, if a user passes in an extra CLI flag to forward to redpanda that's already specified, it will be (incorrectly) provided twice.

This is most relevant in the context of `--memory` and `--reserve-memory`. In ~some~ most cases the built in memory calculations are tedious to understand and difficult to control. Instead, we should let users specify these flags directly and rely on their values rather than trying to force them through the charts memory calculations.

chrisseto:
#### 0c6068618df4a1b4bdacb6191df9d2f0bcd1a50f gotohelm: support delete and unset

This commits add support for transpiling the golang builtin `delete` and
sprig's `unset` helper.

#### 75d5a7848465aebe5f563acab856b47e026dc422 gotohelm: add support for concat and append

This commit adds support for sprig's `concat` helper and additionally
makes the golang builtin `append` work in all cases. Prior to this
commit `append` was limited to appending a single element, it now
supports multiple elements and spreading.

#### 690129a25d0f6eea86630e49e706fd0439ffea3e redpanda: add support for overriding default flags

Prior to this commit, users could not _directly_ control the value of
`--smp`, `--memory`, or `--reserve-memory`. This commit adds support for
overriding the defaults by specifying said flag in
`.statefulset.additionalRedpandaCmdFlags`.

For example, if a user passes `--smp=123` in, the chart's value for smp
will be ignored.

Fixes #1134

### DIFF
--- a/charts/redpanda/ci/30-additional-flags-override-novalues.yaml
+++ b/charts/redpanda/ci/30-additional-flags-override-novalues.yaml
@@ -1,0 +1,4 @@
+statefulset:
+  additionalRedpandaCmdFlags:
+    - "--smp=99"
+    - "--reserve-memory=9999"
--- a/charts/redpanda/configmap.tpl.go
+++ b/charts/redpanda/configmap.tpl.go
@@ -1,0 +1,64 @@
+// Licensed to the Apache Software Foundation (ASF) under one or more
+// contributor license agreements.  See the NOTICE file distributed with
+// this work for additional information regarding copyright ownership.
+// The ASF licenses this file to You under the Apache License, Version 2.0
+// (the "License"); you may not use this file except in compliance with
+// the License.  You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// +gotohelm:filename=_configmap.go.tpl
+package redpanda
+
+import (
+	"fmt"
+
+	"github.com/redpanda-data/helm-charts/pkg/gotohelm/helmette"
+)
+
+// RedpandaAdditionalStartFlags returns a string list of flags suitable for use
+// as `additional_start_flags`. User provided flags will override any of those
+// set by default.
+func RedpandaAdditionalStartFlags(dot *helmette.Dot, smp, memory, reserveMemory string) []string {
+	values := helmette.Unwrap[Values](dot.Values)
+
+	// All `additional_start_flags` that are set by the chart.
+	chartFlags := map[string]string{
+		"smp":               smp,
+		"memory":            fmt.Sprintf("%sM", memory),
+		"reserve-memory":    fmt.Sprintf("%sM", reserveMemory),
+		"default-log-level": values.Logging.LogLevel,
+	}
+
+	// If in developer_mode, don't set reserve-memory.
+	if values.Config.Node["developer_mode"] == true {
+		delete(chartFlags, "reserve-memory")
+	}
+
+	// Check to see if there are any flags overriding the defaults set by the
+	// chart.
+	for flag := range chartFlags {
+		for _, userFlag := range values.Statefulset.AdditionalRedpandaCmdFlags {
+			if helmette.RegexMatch(fmt.Sprintf("^--%s", flag), userFlag) {
+				delete(chartFlags, flag)
+			}
+		}
+	}
+
+	// Deterministically order out list and add in values supplied flags.
+	keys := helmette.Keys(chartFlags)
+	helmette.SortAlpha(keys)
+
+	flags := []string{}
+	for _, key := range keys {
+		flags = append(flags, fmt.Sprintf("--%s=%s", key, chartFlags[key]))
+	}
+
+	return append(flags, values.Statefulset.AdditionalRedpandaCmdFlags...)
+}
--- a/charts/redpanda/templates/_configmap.go.tpl
+++ b/charts/redpanda/templates/_configmap.go.tpl
@@ -1,0 +1,31 @@
+{{- /* Generated from "configmap.tpl.go" */ -}}
+
+{{- define "redpanda.RedpandaAdditionalStartFlags" -}}
+{{- $dot := (index .a 0) -}}
+{{- $smp := (index .a 1) -}}
+{{- $memory := (index .a 2) -}}
+{{- $reserveMemory := (index .a 3) -}}
+{{- range $_ := (list 1) -}}
+{{- $values := $dot.Values.AsMap -}}
+{{- $chartFlags := (dict "smp" $smp "memory" (printf "%sM" $memory) "reserve-memory" (printf "%sM" $reserveMemory) "default-log-level" $values.logging.logLevel ) -}}
+{{- if (eq (index $values.config.node "developer_mode") true) -}}
+{{- $_ := (unset $chartFlags "reserve-memory") -}}
+{{- end -}}
+{{- range $flag, $_ := $chartFlags -}}
+{{- range $_, $userFlag := $values.statefulset.additionalRedpandaCmdFlags -}}
+{{- if (regexMatch (printf "^--%s" $flag) $userFlag) -}}
+{{- $_ := (unset $chartFlags $flag) -}}
+{{- end -}}
+{{- end -}}
+{{- end -}}
+{{- $keys := (keys $chartFlags) -}}
+{{- $_ := (sortAlpha $keys) -}}
+{{- $flags := (list ) -}}
+{{- range $_, $key := $keys -}}
+{{- $flags = (mustAppend $flags (printf "--%s=%s" $key (index $chartFlags $key))) -}}
+{{- end -}}
+{{- (dict "r" (concat $flags $values.statefulset.additionalRedpandaCmdFlags)) | toJson -}}
+{{- break -}}
+{{- end -}}
+{{- end -}}
+
--- a/charts/redpanda/testdata/30-additional-flags-override-values.yaml.golden
+++ b/charts/redpanda/testdata/30-additional-flags-override-values.yaml.golden
@@ -15,70 +15,22 @@ spec:
   maxUnavailable: 1
   selector:
     matchLabels: 
-      app.kubernetes.io/component: redpanda-statefulset
-      app.kubernetes.io/instance: redpanda
       app.kubernetes.io/name: redpanda
+      app.kubernetes.io/instance: "redpanda"
+      app.kubernetes.io/component: redpanda-statefulset
       redpanda.com/poddisruptionbudget: redpanda
 ---
 # Source: redpanda/charts/console/templates/serviceaccount.yaml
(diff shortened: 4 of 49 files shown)


chrisseto:
Happy to split this up into multiple PRs if that's preferred!

pkg/gotohelm/testdata/src/example/main.go
```diff
@@ -39,6 +39,10 @@ func main() {
 			out = map[string]any{}
 		}
 
+		if err != nil {
+			err = fmt.Sprintf("%+v", err)
```
RafalKorepta, line 43: Why `fmt.Sprintf`?

pkg/gotohelm/transpiler_test.go
```diff
@@ -254,11 +254,10 @@ func (g *GoRunner) Render(ctx context.Context, dot *helmette.Dot) (map[string]an
...
 	select {
 	case res := <-g.outputCh:
-		var err error
-		if e, ok := res["err"]; ok && e != nil {
-			return nil, fmt.Errorf("%#v", e)
+		if err, ok := res["err"]; ok && err != nil {
+			return nil, fmt.Errorf("error from go code: %s", err)
 		}
-		return res["result"].(map[string]any), err
+		return res["result"].(map[string]any), nil
```
RafalKorepta, line 260: Good catch!

pkg/gotohelm/testdata/src/example/sprig/sprig.go
```diff
@@ -15,9 +15,26 @@ func Sprig() map[string]any {
...
 	}
 }
 
+func unset() []map[string]int {
+	m1 := map[string]int{"0": 0, "1": 1, "2": 2}
+	m2 := map[string]int{"0": 0, "1": 1, "2": 2}
+	m3 := map[string]int{"0": 0, "1": 1, "2": 2}
+	m4 := map[string]int{"0": 0, "1": 1, "2": 2}
+
+	delete(m2, "0")
```
RafalKorepta, line 37: I'm not sure how `delete` transpilation maps to `unset` helm template.

pkg/gotohelm/helmette/sprig.go
```diff
@@ -167,3 +167,8 @@ func MustFromJSON(data string) any {
 func Unset[K comparable, V any](d map[K]V, key K) {
 	delete(d, key)
 }
+
+// Concat is the go equivalent of sprig's `concat`.
+func Concat[T any](l1, l2 []T) []T {
```
RafalKorepta, line 172: Should it be 
```go
func Concat[T any](l1, l2 ...T) []T {
...
}
```

pkg/gotohelm/transpiler.go
```diff
@@ -597,6 +599,7 @@ func (t *Transpiler) transpileCallExpr(n *ast.CallExpr) Node {
 		"helmette.MustFromJSON": "mustFromJson",
 		"helmette.MustToJSON":   "mustToJson",
 		"helmette.RegexMatch":   "regexMatch",
+		"helmette.SortAlpha":    "sortAlpha",
```
RafalKorepta, line 602: Why `sortAlpha` is added to this commit?

charts/redpanda/configmap.tpl.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"fmt"
+
+	"github.com/redpanda-data/helm-charts/pkg/gotohelm/helmette"
+)
+
+// RedpandaAdditionalStartFlags returns a string list of flags suitable for use
+// as `additional_start_flags`. User provided flags will override any of those
+// set by default.
+func RedpandaAdditionalStartFlags(dot *helmette.Dot, smp, memory, reserveMemory string) []string {
```
RafalKorepta, line 28: Can you create go test for `RedpandaAdditionalStartFlags`?

charts/redpanda/configmap.tpl.go
```diff
@@ -0,0 +1,64 @@
...
+	for flag := range chartFlags {
+		for _, userFlag := range values.Statefulset.AdditionalRedpandaCmdFlags {
+			if helmette.RegexMatch(fmt.Sprintf("^--%s", flag), userFlag) {
+				delete(chartFlags, flag)
+			}
+		}
+	}
+
+	// Deterministically order out list and add in values supplied flags.
+	keys := helmette.Keys(chartFlags)
```
RafalKorepta, line 55: `Keys` is not implemented 

https://github.com/redpanda-data/helm-charts/blob/b728456fe2d490a84100bb5e6fb2937deed4aa15/pkg/gotohelm/helmette/sprig.go#L36-L39

charts/redpanda/configmap.tpl.go
RafalKorepta: NIT: Can smp, memory and reserveMemory be of type int or float?

charts/redpanda/testdata/30-additional-flags-override-values.yaml.golden
```diff
@@ -0,0 +1,1402 @@
+---
```
RafalKorepta, line 1: This file is too old. Please remote `charts/redpanda/Chart.lock` and re-run.

RafalKorepta: https://github.com/redpanda-data/helm-charts/actions/runs/8559264006/job/23455747231?pr=1151#step:5:293

charts/redpanda/configmap.tpl.go
```diff
@@ -0,0 +1,64 @@
...
+// set by default.
+func RedpandaAdditionalStartFlags(dot *helmette.Dot, smp, memory, reserveMemory string) []string {
+	values := helmette.Unwrap[Values](dot.Values)
+
+	// All `additional_start_flags` that are set by the chart.
+	chartFlags := map[string]string{
+		"smp":               smp,
+		"memory":            fmt.Sprintf("%sM", memory),
+		"reserve-memory":    fmt.Sprintf("%sM", reserveMemory),
+		"default-log-level": values.Logging.LogLevel,
```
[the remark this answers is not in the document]
chrisseto: The `logLevel` key is required by our JSON schema so there's no need to add a default here ;)

```
	LogLevel    string `json:"logLevel" jsonschema:"required,pattern=^(error|warn|info|debug|trace)$"`
```

Though the `logging` stanza isn't required... Might as well mark it as such given that we provide a default?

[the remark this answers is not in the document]
RafalKorepta: The values.yaml is not generated, so we need to make sure it is defined.

I'm fine leaving it as is.

charts/redpanda/templates/_configmap.go.tpl
```diff
@@ -0,0 +1,31 @@
+{{- /* Generated from "configmap.tpl.go" */ -}}
+
+{{- define "redpanda.RedpandaAdditionalStartFlags" -}}
```
alejandroEsc, line 3: Just realized what bothers me about these files. Are we not supposed to indent?

alejandroEsc (commented):
Just one question, still trying to get a hang of this.

pkg/gotohelm/testdata/src/example/sprig/sprig.go
chrisseto: Golang's `delete` and sprig's `unset` are equivalent functions, no? They both remove a key from a dictionary/map.

charts/redpanda/templates/_configmap.go.tpl
chrisseto: They're generated files, I didn't really intend for them to be human readable. Because everything is just JSON, there's no significance to the white space.

You're welcome to implement pretty printing, it might be a bit trickier than you think though.

charts/redpanda/configmap.tpl.go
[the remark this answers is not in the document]
chrisseto: > The values.yaml is not generated, so we need to make sure it is defined.

If the field was omitted in the values.yaml but required in the JSON schema, all our tests would start failing. It's not generated yet but there is an enforced structure :D

pkg/gotohelm/testdata/src/example/sprig/sprig.go
RafalKorepta: THX for explaining.

charts/redpanda/ci/30-additional-flags-override-novalues.yaml
```diff
@@ -0,0 +1,4 @@
+statefulset:
```
RafalKorepta, line 1: Could you add https://github.com/redpanda-data/helm-charts/blob/b14967ada018fc9be743665040a6191f187ed65a/.github/check-ci-files.sh#L14

`novalues.yaml` to the if statement? CI is failing at this moment with

```
verifying file: charts/redpanda/ci/30-additional-flags-override-novalues.yaml
- file does not match neither suffix '-values.yaml' nor -values.yaml.tpl'
```

https://github.com/redpanda-data/helm-charts/actions/runs/8602844060/job/23573322602?pr=1151#step:5:321

pkg/gotohelm/transpiler.go
```diff
@@ -650,26 +651,27 @@ func (t *Transpiler) transpileCallExpr(n *ast.CallExpr) Node {
...
+		"helmette.ToJSON":       "toJson",
+		"helmette.Tpl":          "tpl",
+		"helmette.Trunc":        "trunc",
+		"helmette.Unset":        "unset",
 		"helmette.Upper":          "upper",
-		"maps.Keys":               "keys",
-		"math.Floor":              "floor",
+		"maps.Keys":             "keys",
+		"math.Floor":            "floor",
```
RafalKorepta: NIT: go fmt was not executed against this file.

RafalKorepta (approved):
LGTM

CI just not pass ATM

charts/redpanda/testdata/99-none-existent-config-options-with-empty-values.yaml.golden
```diff
@@ -623,7 +622,7 @@ kind: Deployment
 metadata:
   name: redpanda-console
   labels:
-    helm.sh/chart: console-0.7.25
+    helm.sh/chart: console-0.7.24
```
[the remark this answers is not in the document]
chrisseto: Dang thought I fixed that. I guess the dep update command isn't doing what I was hoping for? I'll see if I can fix that.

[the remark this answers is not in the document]
chrisseto: Fixed for this PR, going to make a tracking issue for dealing with this. I think I'm leaning towards checking in Chart.lock but excluding it from our releases.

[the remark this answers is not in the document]
chrisseto: https://github.com/redpanda-data/helm-charts/issues/1166

pkg/gotohelm/testdata/src/example/main.go
chrisseto: The JSON serialization of error sometimes is pretty worthless. Printf turns it into a human readable string that we can then show in the tests. I had a out of bounds error happening for a bit but the JSON version of it didn't contain any usable information it was just {}. The sprintf version had the full stacktrace

PR #1152, opened 2024-04-04
Change the helmette merge behavior to not mutate parameters

RafalKorepta:
As the mergo library and helm chart mutates first dictionary (map) the transpiler and the spring wrapper changes that behavior. The first argument to merge function is created inline inside the wrapper function, so none of the arguments are changed.

### DIFF
--- a/pkg/gotohelm/transpiler.go
+++ b/pkg/gotohelm/transpiler.go
@@ -591,7 +591,6 @@ func (t *Transpiler) transpileCallExpr(n *ast.CallExpr) Node {
 		"helmette.Keys":         "keys",
 		"helmette.KindIs":       "kindIs",
 		"helmette.KindOf":       "kindOf",
-		"helmette.Merge":        "merge",
 		"helmette.MustFromJSON": "mustFromJson",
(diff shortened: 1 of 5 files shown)


pkg/gotohelm/transpiler.go
```diff
@@ -607,6 +607,10 @@ func (t *Transpiler) transpileCallExpr(n *ast.CallExpr) Node {
 	name := callee.Pkg().Name() + "." + callee.Name()
 
 	if tplFuncName, ok := funcMapping[name]; ok {
+		if tplFuncName == "merge" {
```
chrisseto, line 610: Could you push this change into the switch statement below? That's where all the mappings are for non-one-to-one functions are. See `helmette.Dig` for an example :D

chrisseto (approved):
LGTM! Just a suggestion on where to move the if statement to.

pkg/gotohelm/transpiler.go
RafalKorepta: Ok, I will do that.

PR #1153, opened 2024-04-04
Add statefulset pod template labels

### LINKED ISSUE #1147: 🫐 Allow to pass specific pod labels exclusively to Redpanda brokers
Pod labels that are only assigned to Redpanda pods. Otherwise, managed identity credentials are leaked to pods that don't need to access Blob Storage: 

Screenshot 2024-04-04 at 12 52 00 PM

### Why? 
We only need to set workload identity environment variables in Redpanda pods. 

> For applications using workload identity, it's required to add the label azure.workload.identity/use: "true" to the pod spec for AKS to move workload identity to a Fail Close scenario to provide a consistent and reliable behavior for pods that need to use workload identity. Otherwise the pods fail after they are restarted. 

https://learn.microsoft.com/en-us/azure/aks/workload-identity-overview?tabs=dotnet#pod-labels

RafalKorepta:
https://github.com/redpanda-data/helm-charts/pull/1153/commits/4f0d837e2400c2f83e08a59e42594db7a7552a05

#### Make commonLabels over default labels

Merge works that first dictionary/map key will not be overwritten.

https://github.com/redpanda-data/helm-charts/pull/1153/commits/498d5d189abad00ee14402ca08fb3fc7cd1cc31d

#### Allow adding StatefulSet Pod template labels

New podTemplateLabels are not added to label selector used by multiple resources:
* StatefulSet affinity
* StatefulSet selector
* Service
* PodDisruptionBudget

https://github.com/redpanda-data/helm-charts/pull/1153/commits/dd8ca4fa261da6eeeed49964f9ac6489e3c7c19f

Move StatefulSet Pod Template annotation to dedicated struct

To follow official StatefulSet schema the annotation and labels are gathered
together.

Fixes #1147

### DIFF
--- a/charts/redpanda/helpers.go
+++ b/charts/redpanda/helpers.go
@@ -62,8 +62,7 @@ func FullLabels(dot *helmette.Dot) map[string]string {
 		"app.kubernetes.io/component":  Name(dot),
 	}
 
-	// TODO Shouldn't our defaults take precedence over user provided labels?
-	return helmette.Merge(defaults, labels)
+	return helmette.Merge(labels, defaults)
 }
 
 // Create the name of the service account to use
--- a/charts/redpanda/helpers.go
+++ b/charts/redpanda/helpers.go
@@ -65,6 +65,61 @@ func FullLabels(dot *helmette.Dot) map[string]string {
 	return helmette.Merge(labels, defaults)
 }
 
+// StatefulSetPodLabelsSelector returns the label selector for the Redpanda StatefulSet.
+// If this helm release is an upgrade, the existing statefulset's label selector will be used as it's an immutable field.
+func StatefulSetPodLabelsSelector(dot *helmette.Dot, statefulSet map[string]any) map[string]string {
+	if dot.Release.IsUpgrade && statefulSet != nil {
+		existingStatefulSetLabelSelector := helmette.Dig(statefulSet, nil, "spec", "selector", "matchLabels")
+
+		if existingStatefulSetLabelSelector != nil {
+			return existingStatefulSetLabelSelector.(map[string]string)
+		}
+	}
+
+	values := helmette.Unwrap[Values](dot.Values)
+
+	commonLabels := map[string]string{}
+	if values.CommonLabels != nil {
+		commonLabels = values.CommonLabels
+	}
+
+	component := fmt.Sprintf("%s-statefulset",
+		strings.TrimSuffix(helmette.Trunc(51, Name(dot)), "-"))
+
+	defaults := map[string]string{
+		"app.kubernetes.io/component": component,
+		"app.kubernetes.io/instance":  dot.Release.Name,
+		"app.kubernetes.io/name":      Name(dot),
+	}
+
+	return helmette.Merge(commonLabels, defaults)
+}
+
+// StatefulSetPodLabels returns the label that includs label selector for the Redpanda PodTemplate.
+// If this helm release is an upgrade, the existing statefulset's pod template labels will be used as it's an immutable field.
+func StatefulSetPodLabels(dot *helmette.Dot, statefulSet map[string]any) map[string]string {
+	if dot.Release.IsUpgrade && statefulSet != nil {
+		existingStatefulSetPodTemplateLabels := helmette.Dig(statefulSet, nil, "spec", "template", "metadata", "labels")
+
+		if existingStatefulSetPodTemplateLabels != nil {
+			return existingStatefulSetPodTemplateLabels.(map[string]string)
+		}
+	}
+
+	values := helmette.Unwrap[Values](dot.Values)
+
+	statefulSetLabels := map[string]string{}
+	if values.Statefulset.PodTemplate.Labels != nil {
+		statefulSetLabels = values.Statefulset.PodTemplate.Labels
+	}
+
+	defults := map[string]string{
+		"redpanda.com/poddisruptionbudget": Fullname(dot),
+	}
+
+	return helmette.Merge(statefulSetLabels, StatefulSetPodLabelsSelector(dot, nil), defults)
+}
+
 // Create the name of the service account to use
 func ServiceAccountName(dot *helmette.Dot) string {
 	values := helmette.Unwrap[Values](dot.Values)
--- a/charts/redpanda/helpers.go
+++ b/charts/redpanda/helpers.go
@@ -120,6 +120,21 @@ func StatefulSetPodLabels(dot *helmette.Dot, statefulSet map[string]any) map[str
 	return helmette.Merge(statefulSetLabels, StatefulSetPodLabelsSelector(dot, nil), defults)
 }
 
+// StatefulSetPodAnnotations returns the annotation for the Redpanda PodTemplate.
+func StatefulSetPodAnnotations(dot *helmette.Dot, configMapChecksum string) map[string]string {
+	values := helmette.Unwrap[Values](dot.Values)
+
+	configMapChecksumAnnotation := map[string]string{
+		"config.redpanda.com/checksum": configMapChecksum,
+	}
+
+	if values.Statefulset.PodTemplate.Annotations != nil {
+		return helmette.Merge(values.Statefulset.PodTemplate.Annotations, configMapChecksumAnnotation)
+	}
+
+	return helmette.Merge(values.Statefulset.Annotations, configMapChecksumAnnotation)
+}
+
 // Create the name of the service account to use
 func ServiceAccountName(dot *helmette.Dot) string {
 	values := helmette.Unwrap[Values](dot.Values)
--- a/charts/redpanda/templates/_statefulset.tpl
+++ b/charts/redpanda/templates/_statefulset.tpl
@@ -15,23 +15,22 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */}}
 
+{{- define "statefulset-pod-labels-selector" -}}
+{{- /*
+  StatefulSets cannot change their selector. Use the existing one even if it's broken.
+  New installs will get better selectors.
+*/ -}}
+{{- $sts := lookup "apps/v1" "StatefulSet" .Release.Namespace (include "redpanda.fullname" .) -}}
+{{- get ((include "redpanda.StatefulSetPodLabelsSelector" (dict "a" (list . $sts))) | fromJson) "r" | toYaml }}
+{{- end -}}
+
 {{- define "statefulset-pod-labels" -}}
 {{- /*
   StatefulSets cannot change their selector. Use the existing one even if it's broken.
   New installs will get better selectors.
 */ -}}
 {{- $sts := lookup "apps/v1" "StatefulSet" .Release.Namespace (include "redpanda.fullname" .) -}}
-{{- $labels := dig "spec" "selector" "matchLabels" "" $sts -}}
-{{- if not (empty $labels) -}}
-{{ $labels | toYaml }}
-{{- else -}}
-app.kubernetes.io/name: {{ template "redpanda.name" . }}
-app.kubernetes.io/instance: {{ .Release.Name | quote }}
-app.kubernetes.io/component: {{ (include "redpanda.name" .) | trunc 51 }}-statefulset
-{{- with .Values.commonLabels }}
-{{ toYaml . }}
-{{- end }}
-{{- end -}}
+{{- get ((include "redpanda.StatefulSetPodLabels" (dict "a" (list . $sts))) | fromJson) "r" | toYaml }}
 {{- end -}}
 
 {{/*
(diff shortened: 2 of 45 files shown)


charts/redpanda/helpers.go
```diff
@@ -62,8 +62,46 @@ func FullLabels(dot *helmette.Dot) map[string]string {
 		"app.kubernetes.io/component":  Name(dot),
 	}
 
-	// TODO Shouldn't our defaults take precedence over user provided labels?
-	return helmette.Merge(defaults, labels)
+	return helmette.Merge(labels, defaults)
```
chrisseto, line 65: This _could_ be a breaking change and might be worth thinking a bit more about? It would also prevent us from being able to overwrite the chart version in our golden tests, right?

charts/redpanda/helpers.go
```diff
@@ -62,8 +62,46 @@ func FullLabels(dot *helmette.Dot) map[string]string {
...
+	}
+
+	values := helmette.Unwrap[Values](dot.Values)
+
+	commonLabels := map[string]string{}
+	if values.CommonLabels != nil {
+		commonLabels = values.CommonLabels
+	}
+
+	component := fmt.Sprintf("%s-statefulset",
```
chrisseto, line 80: nit: use the string helper function please

charts/redpanda/ci/31-overwrite-statefulset-pod-labels.yaml
```diff
@@ -0,0 +1,3 @@
+statefulset:
+  podTemplateLabels:
+    azure.workload.identity/use: "true"
```
chrisseto, line 3: suuuuuuuuuuuper nit: Please add a newline to the end of this file. If you open it in vim, save, and close it that'll do it.

charts/redpanda/helpers.go
```diff
@@ -62,8 +62,46 @@ func FullLabels(dot *helmette.Dot) map[string]string {
...
+	}
+
+	values := helmette.Unwrap[Values](dot.Values)
+
+	statefulSetLabels := map[string]string{}
+	if values.Statefulset.PodTemplateLabels != nil {
+		statefulSetLabels = values.Statefulset.PodTemplateLabels
+	}
+
+	return helmette.Merge(statefulSetLabels, StatefulSetPodLabelsSelector(dot, nil))
```
chrisseto, line 104: I need to think about this more but could passing `nil` here result in a broken StatefulSet?

charts/redpanda/templates/_statefulset.tpl
```diff
@@ -15,23 +15,24 @@ See the License for the specific language governing permissions and
...
+{{- get ((include "redpanda.StatefulSetPodLabelsSelector" (dict "a" (list . $labels))) | fromJson) "r" | toYaml }}
 {{- end -}}
+
+{{- define "statefulset-pod-labels" -}}
+{{- /*
+  StatefulSets cannot change their selector. Use the existing one even if it's broken.
+  New installs will get better selectors.
+*/ -}}
+{{- $sts := lookup "apps/v1" "StatefulSet" .Release.Namespace (include "redpanda.fullname" .) -}}
+{{- $labels := dig "spec" "template" "metadata" "labels" nil $sts -}}
```
chrisseto, line 34: If it would be easier, I _think_ you could just pass in the entire sts and then handle it as a fully types StatefulSet in Go world. That might also solve the `nil` problem above?

charts/redpanda/values.go
```diff
@@ -236,6 +236,7 @@ type Statefulset struct {
 	} `json:"updateStrategy" jsonschema:"required"`
 	AdditionalRedpandaCmdFlags []string          `json:"additionalRedpandaCmdFlags"`
 	Annotations                map[string]string `json:"annotations" jsonschema:"required"`
+	PodTemplateLabels          map[string]string `json:"podTemplateLabels"`
```
chrisseto, line 239: Should we nest this instead to make it look more like a statefulset? IE
```go
PodTemplate struct {
    Labels map[string]string
}
```

charts/redpanda/helpers.go
```diff
@@ -62,8 +62,46 @@ func FullLabels(dot *helmette.Dot) map[string]string {
 		"app.kubernetes.io/component":  Name(dot),
 	}
 
-	// TODO Shouldn't our defaults take precedence over user provided labels?
-	return helmette.Merge(defaults, labels)
+	return helmette.Merge(labels, defaults)
+}
+
+func StatefulSetPodLabelsSelector(dot *helmette.Dot, existingStatefulSetLabelSelector map[string]string) map[string]string {
```
chrisseto, line 68: nit: Add a short snippet of documentation here?

```
// StatefulSetPodLabelsSelector returns the label selector for the Redpanda StatefulSet.
// If this release is an upgrade, the existing statefulset's label selector will be used as it's an immutable field.


```

charts/redpanda/helpers.go
```diff
@@ -62,8 +62,46 @@ func FullLabels(dot *helmette.Dot) map[string]string {
...
+	if existingStatefulSetLabelSelector != nil {
```
chrisseto, line 69: nit: Should we add in a check for `IsUpgrade` here, more so for clarity than anything else?

charts/redpanda/templates/_statefulset.tpl
```diff
@@ -15,23 +15,24 @@ See the License for the specific language governing permissions and
...
+{{- $labels := dig "spec" "selector" "matchLabels" nil $sts -}}
+{{- get ((include "redpanda.StatefulSetPodLabelsSelector" (dict "a" (list . $labels))) | fromJson) "r" | toYaml }}
 {{- end -}}
+
+{{- define "statefulset-pod-labels" -}}
+{{- /*
+  StatefulSets cannot change their selector. Use the existing one even if it's broken.
+  New installs will get better selectors.
+*/ -}}
+{{- $sts := lookup "apps/v1" "StatefulSet" .Release.Namespace (include "redpanda.fullname" .) -}}
```
chrisseto, line 33: ooooooo, we'll want to add support for this in gotohelm soon.

chrisseto (approved):
Couple discussion points and nits, otherwise LGTM!

charts/redpanda/helpers.go
RafalKorepta: > This could be a breaking change and might be worth thinking a bit more about?

Yes, it would be breaking change. My intention was to give user ability to overwrite the defaults. Previously it was named `required`

https://github.com/redpanda-data/helm-charts/blob/bfb4e557ceadfff97a289febbaf5e25fe8b2cdd1/charts/redpanda/templates/_helpers.tpl#L51-L57

> It would also prevent us from being able to overwrite the chart version in our golden tests, right?

Merge works differently. The first argument in the parameters takes precedence. I think MergeOverwrite would work as you think. Currently user can not overwrite defaults map keys. 

https://helm.sh/docs/chart_template_guide/function_list/#mergeoverwrite-mustmergeoverwrite

https://github.com/Masterminds/sprig/blob/581758eb7d96ae4d113649668fa96acc74d46e7f/dict.go#L109-L117

https://github.com/darccio/mergo/blob/cde9f0ea26cccb1168ee3900cf8ca457bb928c3c/merge.go#L59

charts/redpanda/helpers.go
RafalKorepta: If you would like to use https://github.com/redpanda-data/helm-charts/blob/a0b1456c2274dcf85ea70f10af9b4e1a86846c1c/charts/redpanda/helpers.go#L174-L176 then it will not work as I need to trim not by 63 characters, but 51. 

Maybe you have different function in mind? If labels value can have more than 63, then I can use `cleanForK8s` function. Still it will be breaking change as current implementation uses 51 trim numer.

https://github.com/redpanda-data/helm-charts/blob/a0b1456c2274dcf85ea70f10af9b4e1a86846c1c/charts/redpanda/templates/_statefulset.tpl#L30

charts/redpanda/ci/31-overwrite-statefulset-pod-labels.yaml
RafalKorepta: Np. I will add new line.

charts/redpanda/helpers.go
RafalKorepta: I didn't see problem when performing `go test ./... -short`. My helm version

```
$ helm version
version.BuildInfo{Version:"v3.13.2", GitCommit:"2a2fb3b98829f1e0be6fb18af2f6599e0f4e8243", GitTreeState:"clean", GoVersion:"go1.21.4"}
```

charts/redpanda/templates/_statefulset.tpl
RafalKorepta: I can pass StatefulSet object. No problem.

charts/redpanda/templates/_statefulset.tpl
RafalKorepta: Yes, but I would like to change this in different PR.

charts/redpanda/helpers.go
RafalKorepta: Yes, I can add that.

charts/redpanda/helpers.go
RafalKorepta: Done

charts/redpanda/helpers.go
RafalKorepta: It seems I have some problem.

```
helm.go:84: [debug] template: redpanda/templates/statefulset.yaml:54:18: executing "redpanda/templates/statefulset.yaml" at <include "statefulset-pod-labels" .>: error calling include: template: redpanda/templates/_statefulset.tpl:33:10: executing "statefulset-pod-labels" at <include "redpanda.StatefulSetPodLabels" (dict "a" (list . $sts))>: error calling include: template: redpanda/templates/_helpers.go.tpl:90:64: executing "redpanda.StatefulSetPodLabels" at <include "redpanda.StatefulSetPodLabelsSelector" (dict "a" (list $dot nil))>: error calling include: template: redpanda/templates/_helpers.go.tpl:59:42: executing "redpanda.StatefulSetPodLabelsSelector" at <dig "spec" "selector" "matchLabels" nil $statefulSet>: error calling dig: interface conversion: interface {} is nil, not map[string]interface {}
```

charts/redpanda/helpers.go
chrisseto: Ooooh, 😵‍💫 Yeah I do not like sprig's argument ordering or examples.

alejandroEsc (commented):
Two questions about this approach.

charts/redpanda/helpers.go
RafalKorepta: Problem solved and the `chart_test.go` generated new `yaml` golden file `charts/redpanda/testdata/31-overwrite-statefulset-pod-labels-values.yaml.golden`.

charts/redpanda/values.go
RafalKorepta: The annotations are used only in Pod template definition 
https://github.com/redpanda-data/helm-charts/blob/a0b1456c2274dcf85ea70f10af9b4e1a86846c1c/charts/redpanda/templates/statefulset.yaml#L52-L57

not in StatefulSet object itself
https://github.com/redpanda-data/helm-charts/blob/a0b1456c2274dcf85ea70f10af9b4e1a86846c1c/charts/redpanda/templates/statefulset.yaml#L36-L43

charts/redpanda/templates/statefulset.yaml
```diff
@@ -43,7 +43,7 @@ metadata:
 {{- end }}
 spec:
   selector:
-    matchLabels: {{ (include "statefulset-pod-labels" .) | nindent 6 }}
+    matchLabels: {{ (include "statefulset-pod-labels-selector" .) | nindent 6 }}
   serviceName: {{ template "redpanda.servicename" . }}
   replicas: {{ .Values.statefulset.replicas | int64 }}
   updateStrategy:
```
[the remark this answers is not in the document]
RafalKorepta: It's the desire to move/chunk current definition of any object to go.

P.S. I realised that I can include `redpanda.com/poddisruptionbudget` to the `statefulset-pod-labels` template.

charts/redpanda/helpers.go
chrisseto: Aaaaaaaaaaaaaah, My bad I missed the number.

charts/redpanda/values.go
chrisseto: Oh boy that's confusing. Hmmm. Can you toss a comment on the Annotations field so we remember that it's for the pod template? I'd still go with some nesting here, we can tackle moving/deprecating the annotations field later.

Issue #1149, opened 2024-04-04
Update testdata with the release


RafalKorepta:
### What would you like to be added?

When release is happen the CI could auto commit changes to `charts/redpanda/testdata/` directory.

#### Reference 

https://github.com/redpanda-data/helm-charts/pull/1136/commits/1b507ff29c7cce070283e055ca460d76f1febcbb

### Why is this needed?

To not break CI when release happen.

chrisseto:
As part of this, we'll also want to make CI trigger `go-test` whenever the redpanda or console charts change so that CI will block version bumps if we forget to update anything.

RafalKorepta:
> Hmmm, maybe we just make go-test always run? It's pretty fast all things considered.

Yes, it is very quick with current setup.

> As part of this, we'll also want to make CI trigger go-test whenever the redpanda or console charts change so that CI will block version bumps if we forget to update anything.

It's a problem of running `helm dep build`. The changes to `Chart.lock` will be visible after official helm chart release.

PR #1163, opened 2024-04-08
Add automatic generation for golden files in release

RafalKorepta:
Nix will be used to define tools used during release process.

Redpanda orgnization wide bot will be used as an author for auto commits.

Successful CI run that creates auto commit.

CI - https://github.com/redpanda-data/helm-charts/actions/runs/8602431149/job/23571983449 Auto commit - https://github.com/redpanda-data/helm-charts/commit/243f2f39fc51ad44ecf685a3d00738561b1e68a8 Kustomization auto commit - https://github.com/redpanda-data/helm-charts/commit/5d0793e06918afd03ae8dfb747483915dfc27764

Above changes were possible thanks to this test commit - https://github.com/redpanda-data/helm-charts/commit/b8f9193c55a63ce56042c630907fc81567f35b93#diff-e426ed45842837026e10e66af23d9c7077e89eacbe6958ce7cb991130ad05ada

Fixes https://github.com/redpanda-data/helm-charts/issues/1149


.github/workflows/release.yaml
```diff
@@ -23,31 +23,60 @@ on:
...
+          github_access_token: ${{ secrets.GITHUB_TOKEN }}
+
+      # Cache the nix store.
+      - uses: DeterminateSystems/magic-nix-cache-action@v4
+
+      # Try to checkout a branch with the same name as provided in the dispatch
+      - uses: actions/checkout@v4 # v3.5.3
+        id: checkout
+        with:
+          ref: ${{ github.event.client_payload.branch }}
```
chrisseto, line 48: Can you provide an example of what the branch name would be here? Are we expecting this commit to go to `main`?

.github/workflows/release.yaml
```diff
@@ -23,31 +23,60 @@ on:
...
-      - name: Configure Git
+      - name: Update testdata
         run: |
-          git config user.name "$GITHUB_ACTOR"
-          git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
+          go test ./charts/redpanda -update -short
 
+      # Commit changes must be run before chart-releaser execution, because auto commit would include helm chart tar.gz artifact
+      - name: Commit changes for testdata golden files
+        uses: EndBug/add-and-commit@6e9dd5ddd54bb6b0c1d7b21824951061b9d0a2c0
```
chrisseto, line 72: Just my two cents, it might be better to just write a shell script that does this commit for us? It's one less dep for the project and we can run shell scripts locally to test them out. No need to change if this works though.

chrisseto (approved):
LGTM sans a quick question about the branch names.

.github/workflows/release.yaml
RafalKorepta: Because we will always run on main it will always be `main`

https://github.com/redpanda-data/helm-charts/blob/b14967ada018fc9be743665040a6191f187ed65a/.github/workflows/release.yaml#L22-L25

I took it from one of our internal repository example.

I can change it to hardcoded `main`.

chrisseto: AH, okay. Hard coded or just a quick comment at the end saying something like `branch will always be a branch from the list on line 24`

.github/workflows/release.yaml
RafalKorepta: I just borrowed this from console-enterpise repo.

I will change that to
```
          git config --global user.name 'vbotbuildovich'
          git config --global user.email 'vbotbuildovich@users.noreply.github.com'
          git config --global --add --bool push.autoSetupRemote true
          git add --all
          git commit -m "[Auto Commit] Update console-oss dependency"
          git push
```

PR #1169, opened 2024-04-09
connectors: Add envFrom configuration option to connectors chart

### LINKED ISSUE #1165: Connectors: Add Support for External Secrets via envFrom

- Enable support for passing a `envFrom` value in the Connectors Helm Chart
- Enable an env var configuration provider using `EnvVarConfigProvider`

### Why is this needed?

Currently, the helm chart is very limited in terms of passing external secrets to the pods. We can only directly specify values in the `values.yml` file, or bundle them inside the image itself. 

Additionally, there seems to be some support for directly using AWS secret manager, but this does not cover all use cases. 

Adding support for envFrom (documentation) we are able to pass Secret resources ([documentation](https://kubernetes.io/docs/concepts/configuration/secret/)) with values to add to the environment of the pods directly. Then, by enabling the EnvVarConfigProvider globally, we are able to use these secrets when configuring connector instances.

I am willing to contribute directly for this issue and will soon open a short PR implementing this suggestion.

antbz:
Closes #1165 

- Configures a environment variable configuration provider based on Kafka [EnvVarConfigProvider](https://kafka.apache.org/35/javadoc/org/apache/kafka/common/config/provider/EnvVarConfigProvider.html)
- Adds option to pass `envFrom` to `deployment.yaml` template
- Adds `deployment.extraEnvFrom` to `values.yaml`

### DIFF
--- a/charts/connectors/Chart.yaml
+++ b/charts/connectors/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 0.1.10
+version: 0.1.11
 
 # The app version is the default version of Redpanda Connectors to install.
 appVersion: v1.0.6
--- a/charts/connectors/README.md
+++ b/charts/connectors/README.md
@@ -3,7 +3,7 @@
 description: Find the default values and descriptions of settings in the Redpanda Connectors Helm chart.
 ---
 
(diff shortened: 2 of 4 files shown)


tomasz-sadura:
LGTM, thanks!

antbz:
@tomasz-sadura @RafalKorepta Thanks for the feedback! Updated the version as requested.

RafalKorepta (approved):

PR #1174, opened 2024-04-09
chart lock file committed and added to helmignore

alejandroEsc:
As per prescription from @chrisseto we are now committing the chart.lock file to help remove issues with golden tests.


chrisseto (approved):
LGTM! We might need to update the release automation before we merge this?

PR #1186, opened 2024-04-11
Fix release process with omitting Chart.lock unrelated updates

### LINKED ISSUE #1185: Fix release workflow
Something has gone wrong with the release github action. There are maaaaaaaaaaaaaaaaaany commits from the release bot. It seems that the workflow will trigger itself. For some reason it failed on the most recent run and that stopped the cycle. I didn't notice until I was looking through `git log`.

For now, I've manually disabled the action. It will have to be re-enabled to start running again. 

cc @RafalKorepta 

Screenshot 2024-04-10 at 15 22 32

I'm asking devprod/plateng to force push `main` to remove all the noise commits.

RafalKorepta:
Comparison between Chart.lock before and after `helm dep build` execution is required to prevent
circular dependency between Chart.lock update, which is included in release [Auto commit], and
next unnecessary release pipeline execution due to change in `generated` field. Allowed change
in Chart.lock is only when dependencies are different. Dependencies might change only when
Redpanda chart dependencies got new release.

Fixes #1185

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,9 @@ admin.conf
 # templated ci values
 charts/redpanda/ci/21-eks-tiered-storage-with-creds-values.yaml
 
+# Dep Chart Lock file
+Chart.lock
+
 charts/redpanda/templates/redpanda-license.yaml
 charts/redpanda/templates/external-service.yaml
 charts/redpanda/templates/external-tls-secret.yaml
--- a/charts/redpanda/chart_test.go
+++ b/charts/redpanda/chart_test.go
@@ -95,10 +95,26 @@ func TestTemplate(t *testing.T) {
 	client, err := helm.New(helm.Options{ConfigHome: testutil.TempDir(t)})
 	require.NoError(t, err)
 
+	lock, err := helm.GetChartLock("Chart.lock")
+	require.NoError(t, err)
+
 	// Chart deps are kept within ./charts as a tgz archive. Helm dep build
 	// will ensure that ./charts is in sync with Chart.lock.
 	_, err = exec.CommandContext(ctx, "helm", "dep", "build").CombinedOutput()
 	require.NoError(t, err, "failed to refresh helm dependencies")
+
+	newLock, err := helm.GetChartLock("Chart.lock")
+	require.NoError(t, err)
+
+	// Comparison between Chart.lock before and after `helm dep build` execution is required to prevent
+	// circular dependency between Chart.lock update, which is included in release [Auto commit], and
+	// next unnecessary release pipeline execution due to change in `generated` field. Allowed change
+	// in Chart.lock is only when dependencies are different. Dependencies might change only when
+	// Redpanda chart dependencies got new release.
+	if lock.Digest == newLock.Digest {
+		err = helm.UpdateChartLock(lock, "Chart.lock")
+		require.NoError(t, err)
+	}
 
 	values, err := os.ReadDir("./ci")
 	require.NoError(t, err)
--- a/pkg/helm/helm.go
+++ b/pkg/helm/helm.go
@@ -33,6 +33,45 @@ type Chart struct {
 	Description string `json:"description"`
 }
 
+// ChartLock is a helm lock file for dependencies.
+type ChartLock struct {
+	// Generated is the date the lock file was last generated.
+	Generated time.Time `json:"generated"`
+	// Digest is a hash of the dependencies in Chart.yaml.
+	Digest string `json:"digest"`
+	// Dependencies is the list of dependencies that this lock file has locked.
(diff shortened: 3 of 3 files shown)


.gitignore
```diff
@@ -19,6 +19,9 @@ admin.conf
 # templated ci values
 charts/redpanda/ci/21-eks-tiered-storage-with-creds-values.yaml
 
+# Dep Chart Lock file
+Chart.lock
```
alejandroEsc, line 23: So we ignore the file we can commit but then generate? Is this the intent?

RafalKorepta: I was testing if `git add --all` will omit redpanda `Chart.lock`, but turns out once you commit particular file git will track it regardless of the `.gitignore`. I can remove this, but do we want Chart.lock from other charts beside Redpanda?

RafalKorepta:
> Do we also need to do similarly to the other charts?

We don't have such golden tests in other charts, so the answer is no.

charts/redpanda/chart_test.go
```diff
@@ -95,11 +96,26 @@ func TestTemplate(t *testing.T) {
...
 	require.NoError(t, err, "failed to refresh helm dependencies")
 
+	newLock, err := helm.GetChartLock("Chart.lock")
+	require.NoError(t, err)
+
+	// Comparison between Chart.lock before and after `helm dep build` execution is required to prevent
+	// circular dependency between Chart.lock update, which is included in release [Auto commit], and
+	// next unnecessary release pipeline execution. Allowed change in Chart.lock is only when dependencies
+	// are different. Dependencies might change only when Redpanda chart dependencies got new release.
+	if assert.ObjectsAreEqual(lock.Dependencies, newLock.Dependencies) {
```
chrisseto, line 114: Should this be [`reflect.DeepEqual`](https://pkg.go.dev/reflect#DeepEqual)?

chrisseto (commented):
I wonder if we should instead just set the `generated` field of `Chart.lock` to a static time with `yq` and lint that the time is set correctly in CI. Anyone that runs `helm deps build` could generate some weird diffs.

RafalKorepta:
> I wonder if we should instead just set the generated field of `Chart.lock` to a static time with `yq` and lint that the time is set correctly in CI.

I would not recommend that path.

> Anyone that runs `helm deps build` could generate some weird diffs.

Yes, but I would leave that to any dev that need to maintain this repository.

charts/redpanda/chart_test.go
RafalKorepta: Yes, I just default to assert as this is go test :)

charts/redpanda/chart_test.go
```diff
@@ -95,11 +96,26 @@ func TestTemplate(t *testing.T) {
...
 	require.NoError(t, err, "failed to refresh helm dependencies")
 
+	newLock, err := helm.GetChartLock("Chart.lock")
+	require.NoError(t, err)
+
+	// Comparison between Chart.lock before and after `helm dep build` execution is required to prevent
+	// circular dependency between Chart.lock update, which is included in release [Auto commit], and
+	// next unnecessary release pipeline execution. Allowed change in Chart.lock is only when dependencies
+	// are different. Dependencies might change only when Redpanda chart dependencies got new release.
+	if reflect.DeepEqual(lock.Dependencies, newLock.Dependencies) {
```
chrisseto, line 114: nit: I think we could comparing .Digest here would be equivalent and slightly easier to reason about given it's just a string.

For future reference, it might be good to have this comment explicitly call out that `Generated` is the problem.

chrisseto (approved):

RafalKorepta:
> I wonder if we should instead just set the generated field of `Chart.lock` to a static time with `yq` and lint that the time is set correctly in CI.

From my perspective any invocation of `helm dep build` inside our test suite would bump generated timestamp in `charts/redpanda/Chart.lock`. We should always run `helm dep build` as downstream dependencies could change (new release) that could change the golden files output. If none of the Redpanda chart dependnecies changed, then our release process should accommodate that to not create endless loop. I chosen the closes place (that is not bash script) where `helm dep build` is invoked to handle a change in `Chart.lock`. 

The proposal of handling static time with `yq` is not that close to the developer workflow of calling `go test ./...`. The invocation would be codify in the Github Action Workflow which in my opinion creates friction in developer experience, because any one running `go test` would need to remember to exclude `Chart.lock` unrelated change from the commit. To add more `yq` might be additional dependency where go code uses the stdlib (less dependency).

charts/redpanda/chart_test.go
RafalKorepta: Thanks, I changed to `Digest`.

I added information about `generated` filed being to culprit in comment.

PR #1187, opened 2024-04-11
redpanda: include Chart.lock in released chart

chrisseto:
#### de30dc5305c2000f65cd38d1f90394e7ace075f6 redpanda: include Chart.lock in released chart

Prior to this commit we struggled to correctly managed `Chart.lock` as
it was believed that excluding it from the released chart tgz would
allow users to utilize their own version of console or connectors.

This is not the case. `Chart.lock` is merely the output of the version
negotiation done by `helm dep update` which is then used by `helm dep
build` to place the correct chart archives in to `charts/`. `charts/` is
then included within the released helm chart and is the only source for
subcharts[^1].

The attempt to helmignore `Chart.lock` resulted in `helm dep build`
constantly regenerating `Chart.lock` as helm would stop loading the
`Chart.lock` file[^2].

This commit removes `Chart.lock` from helm ignore and some code that
attempted to manage its accidental regeneration. Otherwise it serves as
documention for these choices.

[^1]: https://github.com/helm/helm/blob/14d0c13e9eefff5b4a1b511cf50643529692ec94/cmd/helm/install.go#L261-L290
[^2]: https://github.com/helm/helm/blob/14d0c13e9eefff5b4a1b511cf50643529692ec94/pkg/downloader/manager.go#L86

### DIFF
--- a/charts/redpanda/.helmignore
+++ b/charts/redpanda/.helmignore
@@ -26,6 +26,3 @@ README.md.gotmpl
 *.go
 testdata/
 ci/
-
-# include chart lock file
-Chart.lock
--- a/charts/redpanda/chart_test.go
+++ b/charts/redpanda/chart_test.go
@@ -2,7 +2,6 @@ package redpanda_test
 
 import (
 	"os"
-	"os/exec"
 	"testing"
 
 	"github.com/redpanda-data/helm-charts/charts/redpanda"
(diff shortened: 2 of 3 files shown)


RafalKorepta:
The problem is not within Chart.lock being in Redpanda helm release, but rather running `go test ./... -update` and auto committing golden files. That change will create circular dependency as follows:
* release might find changes to Redpanda helm chart dependencies, so that golden files needs to be updated
* due to auto commit the Github Action Workflow `release` will perform `go test ./... -update`, which will run `go dep build` which will update Chart.lock, so in the end `git add --all` will create new auto commit

chrisseto:
> The problem is not within Chart.lock being in Redpanda helm release, but rather running go test ./... -update and auto committing golden files. That change will create circular dependency as follows:

The circular dep is because `Chart.lock` is being helmignored. When it's not helm ignored `helm dep build` will _never_ result in a change to `Chart.lock` which breaks the cycle. To update the versions someone will have to run `helm dep upgrade`, which should be a distinct step from releasing a new chart IMO.

chrisseto:
Huh, the failure from go test is interesting. I wonder why it's only now showing up.

RafalKorepta (approved):
LGTM, but I feel like it was not working the way you are describing it. Chart.lock should be included in chart release and it is comitted to the repo. I feel like I'm missing something, but we can figure that out if it fails after merge

alejandroEsc:
was this expected:
```
--- FAIL: TestTemplate (0.18s)
    chart_test.go:102: 
        	Error Trace:	/home/runner/work/helm-charts/helm-charts/charts/redpanda/chart_test.go:102
        	Error:      	Received unexpected error:
        	            	exit status 1
        	Test:       	TestTemplate
        	Messages:   	failed to refresh helm dependencies:
        	            	Error: no repository definition for https://charts.redpanda.com,/ https://charts.redpanda.com./ Please add the missing repos via 'helm repo add'
```

PR #1188, opened 2024-04-11
Move golden file generation after chart release invocation

RafalKorepta:
When other than Redpanda chart is being released, then golden files where not updated.

## Ref

https://github.com/redpanda-data/helm-charts/actions/runs/8653258545/job/23728037041#step:8:17
https://github.com/redpanda-data/helm-charts/pull/1182
https://github.com/redpanda-data/helm-charts/pull/1182/commits/422acfdf04ca6062e2aa8feb16b712bdd84f15af

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -63,27 +63,25 @@ jobs:
         run: |
           helm repo add redpanda https://charts.redpanda.com
 
-      - name: Update testdata
-        run: |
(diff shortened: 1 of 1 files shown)


chrisseto (approved):
LGTM!

I think this ties into the whole conversation about `Chart.lock`. If I understand correctly, releasing another chart shouldn't result in any changes to the redpanda chart. Today it does due to Chart.lock being in helmignore.

Issue #1061, opened 2024-02-28
Run kube-linter against the output of `helm template`


chrisseto:
There are a handful of Kubernetes security best practices that we're not following right now. Some are accidental omissions and others may be bugs. We will need to start running [kube-linter](https://github.com/stackrox/kube-linter) against the results of `helm template` to ensure that we're compliant in most cases.
```[tasklist]
### Tasks
```

PR #1190, opened 2024-04-12
add kube-linter to task

alejandroEsc:
Fixes: https://github.com/redpanda-data/helm-charts/issues/1061

adding task file, will look to add action in next PR but first we should get an idea of what things to exclude, what things need to be fixed. Please run

task ci:kube-lint-template

this is preliminary and it justs tests the redpanda chart, we can add more templates later if we wish. For now i included hooks, but we can add the `--no-hooks` flag.

### DIFF
--- a/.kube-linter.yaml
+++ b/.kube-linter.yaml
@@ -1,0 +1,10 @@
+
+# Details on checks can be found on this page and the like: https://docs.kubelinter.io/#/generated/checks
+checks:
+  addAllBuiltIn: true
+  exclude:
+    - "unset-cpu-requirements" # Not every pod requires this, for example hooks, we can change opinion later
+    - "unset-memory-requirements" # Not every pod requires this, for example hooks, we can change opinion later
+    - "no-read-only-root-fs" # Some volume mounts require writes
+    - "required-annotation-email" # Not a requirement for out deployments
+    - "no-node-affinity" # Not every pod requires this
--- a/charts/redpanda/chart_test.go
+++ b/charts/redpanda/chart_test.go
@@ -1,7 +1,9 @@
 package redpanda_test
 
 import (
+	"bytes"
 	"os"
+	"os/exec"
 	"testing"
 
 	"github.com/redpanda-data/helm-charts/charts/redpanda"
@@ -122,6 +124,25 @@ func TestTemplate(t *testing.T) {
 				},
 			})
 			require.NoError(t, err)
+
+			// kube-lint template file
+			var stdout bytes.Buffer
+			var stderr bytes.Buffer
+			inputYaml := bytes.NewBuffer(out)
+
+			cmd := exec.CommandContext(ctx, "kube-linter", "lint", "-", "--format", "json")
+			cmd.Stdin = inputYaml
+			cmd.Stdout = &stdout
+			cmd.Stderr = &stderr
+
+			errKubeLinter := cmd.Run()
+			if errKubeLinter != nil && len(stderr.String()) > 0 {
+				t.Logf("kube-linter error(s) found for %q: \n%s\nstderr:\n%s", v.Name(), stdout.String(), stderr.String())
+			} else if errKubeLinter != nil {
+				t.Logf("kube-linter error(s) found for %q: \n%s", v.Name(), errKubeLinter)
+			}
+			// TODO: remove comment below and the logging above once we agree to linter
+			//require.NoError(t, errKubeLinter)
 
 			testutil.AssertGolden(t, testutil.YAML, "./testdata/"+v.Name()+".golden", out)
 		})
--- a/flake.nix
+++ b/flake.nix
@@ -90,6 +90,7 @@
               pkgs.kubectl
               pkgs.kubernetes-helm
               pkgs.kustomize
+              pkgs.kube-linter
               pkgs.yq # jq but for YAML
             ];
           };


alejandroEsc:
blocking until we have a good idea of what things to ignore.

Taskfile.yaml
```diff
@@ -144,6 +153,12 @@ tasks:
     status:
       - helm search repo -r '\vmetallb/metallb\v' | grep metallb
 
+  tool:install-kube-linter:
```
chrisseto, line 156: Let's install this with nix as it's how we're handing all other deps right now. There's a big ol' block of tools.  you can add [`pkgs.kube-linter`](https://search.nixos.org/packages?channel=23.11&show=kube-linter&from=0&size=50&sort=relevance&type=packages&query=kube-linter) and it'll just work ™️

Taskfile.yaml
```diff
@@ -71,6 +71,15 @@ tasks:
       - .github/check-ci-files.sh charts/redpanda/ci
       - go vet ./...
 
+  ci:kube-lint-template:
```
chrisseto, line 74: Could we instead run this via `go test`? Would love to add this as a second step to `TestTemplate`. Should be fine to run through `os/exec`. There's a JSON output if we want to parse it and that would allow us to manage the ignores in gocode as well.

chrisseto (commented):
Super excited to start seeing the output of this.

Taskfile.yaml
alejandroEsc: is it ok for me to leave this task file, or just remove it?

Taskfile.yaml
alejandroEsc: cool idea, will do.

Taskfile.yaml
chrisseto: In the interest of consistency, let's remove it. If `nix develop` isn't a pleasant experience there are myriad ways to otherwise provide access to them. nix ensures that our tools are pinned to the same version and work on most conceivable environments.

Taskfile.yaml
RafalKorepta: That step could be perform after `helm template`

https://github.com/redpanda-data/helm-charts/blob/6449ce3afcc6bb91607968accded1935ffab52e5/charts/redpanda/chart_test.go#L127-L140

charts/redpanda/chart_test.go
```diff
@@ -139,6 +139,17 @@ func TestTemplate(t *testing.T) {
...
 			require.NoError(t, err)
 
+			// kube-lint template file
+			err = os.WriteFile("temp-template.yaml", out, 0655)
+			require.NoError(t, err)
+
+			out, errKubeLinter := exec.CommandContext(ctx, "kube-linter", "lint", "temp-template.yaml").CombinedOutput()
+			if errKubeLinter != nil {
+				t.Logf("kube-linter error(s) found: \n%s", string(out))
+			}
```
RafalKorepta, line 149:
```suggestion
                         var stdout bytes.Buffer
	                 var stderr bytes.Buffer
                         inputYaml := bytes.NewBuffer(out)
                         
			cmd := exec.CommandContext(ctx, "kube-linter", "lint", "-", "--format", "json")
			cmd.Stdin = inputYaml
	                 cmd.Stdout = &stdout
	                 cmd.Stderr = &stderr
			
			stdout, stderr, errKubeLinter := cmd.Run()
			if errKubeLinter != nil {
				t.Logf("kube-linter error(s) found: \n%s\nstderr:\n%s", stdout.String(), stderr.String())
			}
```

Thanks to that change there will not be dandling `temp-template.yaml` file and stdout can be marshalled to json for later process.

Taskfile.yaml
RafalKorepta: I think this PR missing a removal of `ci:kube-lint-template` task target

charts/redpanda/chart_test.go
chrisseto: We could also pass in the path of the golden file to `kube-linter`. Stdin works just as well though!

alejandroEsc: cmd.Run has only one output? where are you getting 3?

Taskfile.yaml
alejandroEsc: No, that was intentional, i would like to run this manually but ok...  ill remove it

charts/redpanda/chart_test.go
```diff
@@ -139,6 +140,26 @@ func TestTemplate(t *testing.T) {
...
+			cmd := exec.CommandContext(ctx, "kube-linter", "lint", "-", "--format", "json")
+			cmd.Stdin = inputYaml
+			cmd.Stdout = &stdout
+			cmd.Stderr = &stderr
+
+			errKubeLinter := cmd.Run()
+			if errKubeLinter != nil {
+				t.Logf("kube-linter error(s) found for %q: \n%s\nstderr:\n%s", v.Name(), stdout.String(), stderr.String())
+			}
+			if errKubeLinter != nil {
```
chrisseto, line 157: duplicated error handling?

charts/redpanda/chart_test.go
```diff
@@ -139,6 +140,26 @@ func TestTemplate(t *testing.T) {
...
+			var stdout bytes.Buffer
+			var stderr bytes.Buffer
+			inputYaml := bytes.NewBuffer(out)
+
+			cmd := exec.CommandContext(ctx, "kube-linter", "lint", "-", "--format", "json")
+			cmd.Stdin = inputYaml
+			cmd.Stdout = &stdout
+			cmd.Stderr = &stderr
+
+			errKubeLinter := cmd.Run()
```
chrisseto, line 153: nit: in general, I prefer to use inline handling to avoid leaking errors into the larger scope.

```go
if err := ...; err != nil {
    // handle error here
}
```

That being said, I think the error handling here needs to be a bit weird. `kube-linter` could be erroring if it doesn't run correctly _or_ because it found errors. Given that we have the json output, we can pretty reasonably tell what's being reported:
```go
if errKubeLinter != nil && len(kubeLinterParsedJSONOutputErrorsField) > 0 {
 // Kube linter ran successfully but is reporting errors
} else if errKubeLinter != nil {
// Kube-linter did not run successfully.
} else {
// Kube-linter ran successfully and found no errors.
}

```

.kube-linter.yaml
```diff
@@ -0,0 +1,10 @@
+
+checks:
+  addAllBuiltIn: true
+  exclude:
```
chrisseto, line 4: Could we a few brief notes indicating why we've opted to exclude these?

I think `unset-*-requirements` is currently due to a bug that Rafal is actively fixing. The others I'm not super familiar with. If we're just disabling them to get passing results can we file issues for each ignored rule to eventual try and fix?

alejandroEsc: Sure, i mean i want this to be something we all participate on so feel free to add in another PR things you wish to expand on etc... I'll add my own thoughts but does not mean we have to keep those in if we do not want.

PR #1191, opened 2024-04-12
WIP: Proposal for new values shape

chrisseto:
Early working draft of my proposal on where we should move with our Values struct. Currently focused on redpanda.yaml and providing an "advanced mode" type of experience.

This PR is for discussion and will likely be splintered across many PRs once there's a general consensus.


charts/redpanda/configmap.tpl.go
```diff
@@ -62,3 +62,114 @@ func RedpandaAdditionalStartFlags(dot *helmette.Dot, smp, memory, reserveMemory
 
 	return append(flags, values.Statefulset.AdditionalRedpandaCmdFlags...)
 }
+
+// func RedpandaYAMLKafkaListeners(dot *helmette.Dot) []KafkaListener {
```
chrisseto, line 66: Was exploring the existing listener configurations. Will likely have a function that converts the existing shape into this structure and we'll convert our templates to use the new structure.

charts/redpanda/values_util.go
```diff
@@ -1,12 +1,15 @@
-//+gotohelm:ignore=true
+// +gotohelm:ignore=true
```
RafalKorepta, line 1: NIT: I wonder why my IDE is changing this every time 🤔

charts/redpanda/kafka_listener.go
```diff
@@ -0,0 +1,95 @@
...
+
+	// KeyFile maps to the `kafka_api_tls[*].key_file` field.
+	KeyFile FileSource
+
+	// CertFile maps to the `kafka_api_tls[*].cert_file` field.
+	CertFile FileSource
+
+	// TrustStoreFile maps to the `kafka_api_tls[*].truststore_file` field.
+	// +kubebuilder:default={"path": "/etc/ssl/certs/ca-certificates.crt"}
+	TrustStoreFile *FileSource
```
RafalKorepta, line 91: I wonder how users would control volume mount points.

Issue #1197, opened 2024-04-16
Bump supported Helm version to 3.10.0


ligfx:
### What happened?

[d91f5573d525e100240c0d21beda8b152e6674af](https://github.com/redpanda-data/helm-charts/commit/d91f5573d525e100240c0d21beda8b152e6674af) introduced some template code that uses `{{ break }}`, which doesn't exist until Go 1.18 / Helm 3.10.0

Running with older versions of Helm:

```
Error: parse error at (redpanda/templates/_helpers.go.tpl:7): function "break" not defined
helm.go:84: [debug] parse error at (redpanda/templates/_helpers.go.tpl:7): function "break" not defined
```

### What did you expect to happen?

.

### How can we reproduce it (as minimally and precisely as possible)?. Please include values file.

.

### Which are the affected charts?

Redpanda

### Chart Version(s)

Looks like it affects 5.7.36 and up:

<img width="955" alt="image" src="https://github.com/redpanda-data/helm-charts/assets/594093/62163457-bd29-4339-8ad3-6c42a5f75219">

### Cloud provider

.

PR #1199, opened 2024-04-16
Bump supported Helm version

RafalKorepta:
Fixes https://github.com/redpanda-data/helm-charts/issues/1197


chrisseto (approved):
Can we make this change across all of our charts for consistency?

It may be good to have a note about why we need this specific version. I'm not sure the best place to store that information though.

Looks like we could also use [`.Capabilities`](https://helm.sh/docs/chart_template_guide/builtin_objects/) to assert the version within our charts rather than relying on users to check themselves. We can push that off until later though!

PR #1201, opened 2024-04-16
Move git configuration to earlier step

RafalKorepta:
Due to error in https://github.com/redpanda-data/helm-charts/actions/runs/8708200936/job/23884992632#step:7:109 the git configuration is performed just after checkout.

Reference

```
Author identity unknown

*** Please tell me who you are.

Run

  git config --global user.email "you@example.com"
  git config --global user.name "Your Name"

to set your account's default identity.
Omit --global to set the identity only in this repository.

fatal: empty ident name (for <runner@fv-az1245-596.44uquw3knzoenloncamig1c14h.dx.internal.cloudapp.net>) not allowed
Error: exit status 128
Usage:
  cr index [flags]
```

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -58,6 +58,11 @@ jobs:
         with:
           token: ${{ secrets.ACTIONS_BOT_TOKEN }}
           fetch-depth: 0
+      - name: configure git
+        run: |
+          git config --global user.name 'vbotbuildovich'
+          git config --global user.email 'vbotbuildovich@users.noreply.github.com'
+          git config --global --add --bool push.autoSetupRemote true
 
       - name: Add Repositories for each chart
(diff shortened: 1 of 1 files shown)


chrisseto (approved):
LGTM. Can we remove the duplicated `git config` in `Build kustomization for operator`?

RafalKorepta:
> Can we remove the duplicated git config in Build kustomization for operator?

The `kustomization` release was created for anyone that does not want to use helm chart. 

https://github.com/redpanda-data/helm-charts/pull/663

I'm not sure if anyone is using this kustomization. Yet our documentation does not mention it, so it is uncharted territory (in my opinion). My take would be to remove the https://github.com/redpanda-data/helm-charts/tree/kustomize branch and the release process.

PR #1207, opened 2024-04-17
Replace Github workflow direct definition with taskfile goal and sync test cases

RafalKorepta:
Thanks to https://github.com/redpanda-data/helm-charts/pull/1101 test steps are already defined in taskfile.


chrisseto (approved):
LGTM! Thanks for the clean up.

PR #1211, opened 2024-04-17
introduce pkg/valuesutil

chrisseto:
This commit introduces `pkg/valuesutil` which contains a set of
utilities to work with helm values and jsonschemas. Most notably, it
includes the `Generate` function which can deterministically produce a
randomized output that is considered valid for a given json schema.

Future commits will utilize this utility to both fuzz and property test
the redpanda chart.

Examples of both fuzz testing and property testing with go's builtin
constructs are included in `fuzz_test.go`.

### DIFF
--- a/cmd/genschema/main.go
+++ b/cmd/genschema/main.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/invopop/jsonschema"
 	"github.com/redpanda-data/helm-charts/charts/redpanda"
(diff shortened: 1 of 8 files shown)


chrisseto:
I got a wee bit carried away on this one...

RafalKorepta (approved):

RafalKorepta (approved):

PR #1234, opened 2024-04-19
Create SI-to-bytes template equivalent in go code

RafalKorepta:
The SI to bytes template used in redpanda-memory and reserved-memory that translates memory from resources limits or requests.


pkg/gotohelm/helmette/shims.go
```diff
@@ -70,3 +73,25 @@ func Unwrap[T any](from Values) T {
...
+	switch {
+	case strings.HasSuffix(v, "b") || strings.HasSuffix(v, "B"):
+		return v[:len(v)-1]
+	case strings.HasSuffix(v, "K"):
+		v = fmt.Sprintf("%sk", v[:len(v)-1])
+	case strings.HasSuffix(v, "g"):
+		v = fmt.Sprintf("%sG", v[:len(v)-1])
+	case strings.HasSuffix(v, "m"):
+		v = fmt.Sprintf("%sM", v[:len(v)-1])
+	}
```
RafalKorepta, line 93: This is required to match `SI-to-bytes` capability, but I'm all for changing the Redpanda helm schema to match https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/quantity/

PR #1238, opened 2024-04-19
gotohelm: bootstrap _shims.tpl

chrisseto:
#### ac238d9d968ef0f26dcc45d4d510087e65c938ef gotohelm: bootstrap _shims.tpl

Previously, the `bootstrap` package had gotten accidentally checked in
with a good chunk of broken code and some misleading comments that might
have implied that it was actually the source of `_shims.tpl`.

This commit finishes the lofty work that was initially planned by
implementing all of `_shims.tpl` in go and using gotohelm to generate
`_shims.tpl`.

Doing so unearthed a few mistakes that had been over looked which have
now been fixed:
- type assertions actually perform type assertions now.
- dictionary testing and direct access return the zero value of the map
  as it does in go.
- type tests return the zero value of the type as it does in go.
- transpiled dereferences will panic when attempting to dereference `nil`.


chrisseto:
Talking with @RafalKorepta this morning inspired me to revisit the bootstrap package. I realized that it had been committed by accident. This commit worked through re implementing the _shims yaml file in go and actually completing the boot strap because I wanted to add in support for `ptr.Deref` in the most complicated way possible.

It might be nice to do something a bit more dynamic than copying bootstrap.yaml but I haven't come up with a clean way to do so just yet.

chrisseto:
I also realized that type assertions where going unasserted 😬

pkg/gotohelm/testdata/src/example/bootstrap/_shims.tpl
```diff
@@ -1,34 +1,80 @@
 {{- /* Generated from "" */ -}}
 
+{{- /* Generated from "_shims.go" */ -}}
```
RafalKorepta, line 3: Can the first line be deleted to not repeat `Generated from ...`?

chrisseto: Yes, I can update the "bundler" to no include that line. Good call!

RafalKorepta (approved):

RafalKorepta (approved):

PR #1241, opened 2024-04-22
gotohelm: fix and test AST rewrites

### LINKED ISSUE #1138: gotohelm: Write tests for AST rewrites
gotohelm has a two step transpilation process.
1. AST Rewrites to turn unsupported syntax into supported syntax
2. Node by node transpilation

Currently, only E2E tests exist. This works quite well but it's quite difficult to understand if and when AST rewriting goes wrong. We should make a set of tests that only exercise the AST rewrites so it's easier to understand what they do and vet that they're performing the writes correctly.

chrisseto:
#### 52a4f3777438ed25c6a61269f6823a64efc71fd4 gotohelm: fix and test AST rewrites

Previously, AST rewriting could fail in strange scenarios. In
particular, `_, _ := m[""]` would cause very strange errors complaining
about a misplaced `<`.

This was caused by `_`'s always having a nil type and `types.TypeString`
reporting `<nil>` when provided a nil type.

This commit fixes the issue by appropriately populating type information
when performing rewrites and utilizing the `Rhs` of assignments for type
information. It also adds a minimal set of golden tests which clarify
the exact behavior of the rewrites as well as providing a place to more
easily debug future issues. It also removes the now defunct `a` and `b`
test packages as they duplicated tests under ambiguous names.

Fixes #1138


pkg/gotohelm/testdata/src/example/astrewrites/astrewrites.rewritten.go
```diff
@@ -0,0 +1,99 @@
+//go:build rewrites
+package astrewrites
```
RafalKorepta, line 2: Where is the yaml representation of this file?

pkg/gotohelm/rewrite_test.go
```diff
@@ -0,0 +1,55 @@
...
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/redpanda-data/helm-charts/pkg/testutil"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/tools/go/packages"
+)
+
+func TestLoadPackages(t *testing.T) {
```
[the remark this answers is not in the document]
chrisseto: I'm not sure I understand the question? It's just `go test ./...`  and/or `go test ./... -update` like the existing gotohelm tests.

pkg/gotohelm/testdata/src/example/astrewrites/astrewrites.rewritten.go
chrisseto: This package is not currently being transpiled, it's just being used to showcase the AST rewriting. If you'd prefer, I could make all packages get transpiled and have their ASTs rewritten? It is a bit confusing to have everything live in the same place isn't it.

RafalKorepta: Ok, now I understand.

RafalKorepta (approved):

pkg/gotohelm/testdata/src/example/astrewrites/astrewrites.rewritten.go
chrisseto: I went ahead and update this commit to perform rewrites on all the test packages and to transpile this one. Transpiling the rewrites package actually found a tiny bug that's fixed in the bootstrap commit, so it's disabled for this commit but will be enable in the bootstrap commit 👍

PR #1244, opened 2024-04-23
redpanda: Allow specifying additional envvars to redpanda

### LINKED ISSUE #1243: 🫐 Allow providing additional environment variables to the redpanda container
It's currently not possible to pass additional environment variables to the redpanda container. Some users (including our cloud) will need the ability to do so to interact with various features of redpanda itself.

chrisseto:
This commit add support for passing additional environment variables to
the redpanda container via means of a strategic merge patch (kinda).

Rather than implementing strategic merge patch everywhere, this commit
sets the precedent that "overrides" should be exposed to end users via
following the shape of Kubernetes' primitives and following the merge
semantics as described by the strategic merge patch documentation.

Fixes #1243


c4milo (approved):
kosher

charts/redpanda/values.go
```diff
@@ -241,7 +263,7 @@ type Statefulset struct {
 	// Annotations are used only for `Statefulset.spec.template.metadata.annotations`. The StatefulSet does not have
 	// any dedicated annotation.
 	Annotations map[string]string `json:"annotations" jsonschema:"deprecated"`
-	PodTemplate PodTemplate       `json:"podTemplate"`
+	PodTemplate PodTemplate       `json:"podTemplate" jsonschema:"required"`
```
RafalKorepta, line 266: I don't understand why every field under `PodTemplate` including `PodTemplate` is now required field?

alejandroEsc (approved):
lgtm

charts/redpanda/values.go
chrisseto: The JSONSchema applies to the result of merging the users values and our values. By specifying required, we're really just ensuring that they can not be set to `null`.

RafalKorepta (approved):

PR #1245, opened 2024-04-23
Introduce improvements to the transpiler

RafalKorepta:
https://github.com/redpanda-data/helm-charts/commit/f216da637b33ac570bd14526dc5819600eb374cc

Wrap binary expression that have both `float32` type as arguments.

Binary expression can now handle `float32` arguments in both sides.

https://github.com/redpanda-data/helm-charts/commit/91a24f088fd16a8be31f0fc8606403ee4dff085a

Add `float64` literal to transpiler.

Binary expresion can now handle float arguemnts in both sides.

https://github.com/redpanda-data/helm-charts/commit/c03feb54dcc2ad7483f8254303f897d5dc560659

Fix `rune` handling when casting it to `string` via `printf`.

https://github.com/redpanda-data/helm-charts/commit/e9d7c7e2a6f279c77c9e2e78ce7987334895acef

Add `Float64` function to the `helmette` package

The `Atof` function in idiomatic go representation is implemented as
`ParseFloat`. The sprig is changing the naming convention to `float64`.
Similar to `Atoi` the `Float64` returns error in go representation, while
transpiler adding artificial `nil` to match double return values.

https://github.com/redpanda-data/helm-charts/commit/63c5f5106d73b24510b21c7fc88e0b6391b4280e

Add `atoi` function to the `helmette` package

The `Atoi` function in go representation returns an error. In the template
the error, if input is invalid, is not returned. To match the go
implementation the `nil` will be added to transpiled template code.

https://github.com/redpanda-data/helm-charts/commit/ef0cf2bfaee8e8a713acacbd45dbb2e050cda4c6

Add isolated regex test case in `helmette` test data

`MustRegexMatch` can only fail if pattern can not be compiled. This commit
tries to correct the implementation to match the one sprig.

https://github.com/redpanda-data/helm-charts/commit/2eafec2e258d44121b9aae1b17ff6ec08d5a8e4f

Fix slicing string and slice

The `substr` requires to have range represent as integers, but `sub` will
return int64 type. If both types of the `sub` arguments are equal the
`sub` result will be wrapped with type casting.

https://github.com/redpanda-data/helm-charts/commit/707c9d29609dc154305b9032af5a022b2e0c9f16

Change json schema and regexp for memory amount

The memory, under the hood, is Resource.Quantity that follows International
System of units. Resource.Quantity could be unmarshaled from string or integer.
To not complicate the values schema and `genschema` transpiler, by using
Resource.Quantity, the integer needs to be provided as string. The E (exa) or
Ei (exbi) does not work correctly within 64 bits unsigned integer, thats why it
is omitted.

### Reference

- https://physics.nist.gov/cuu/Units/binary.html
- https://www.nist.gov/pml/owm/metric-si-prefixes
- https://en.wikipedia.org/wiki/International_System_of_Units
- https://en.wikipedia.org/wiki/Binary_prefix


pkg/gotohelm/transpiler.go
```diff
@@ -526,50 +526,68 @@ func (t *Transpiler) transpileExpr(n ast.Expr) Node {
 		untyped := [3]string{"_", n.Op.String(), "_"}
 		typed := [3]string{t.typeOf(n.X).String(), n.Op.String(), t.typeOf(n.Y).String()}
 
+		f := func(op string) func(a, b Node) Node {
```
alejandroEsc, line 529: never seen this before, how does it work?

alejandroEsc (approved):
One question, nothing serious.

pkg/gotohelm/transpiler.go
RafalKorepta: Function wraps a closure that would be otherwise repeated in the map. That implementation was previously declared https://github.com/redpanda-data/helm-charts/blob/b7a118bedfd652769bfc8c0a5fda58c0aa397917/pkg/gotohelm/transpiler.go#L561-L564

pkg/gotohelm/transpiler.go
```diff
@@ -732,6 +761,22 @@ func (t *Transpiler) transpileCallExpr(n *ast.CallExpr) Node {
 	// Mappings that are not 1:1 and require some argument fiddling to make
 	// them match up as expected.
 	switch name {
+	case "helmette.Atoi":
+		return &BuiltInCall{
```
chrisseto, line 765: Maybe add a quick note here that `nil` is included to match the `(T, error)` return type of the `helmette` functions?

chrisseto (approved):
LGTM! Nothing blocking just a couple comment suggestions.

pkg/gotohelm/transpiler.go
RafalKorepta: Ok I will include explanation

PR #1250, opened 2024-04-25
[WIP] Convert servicemonitor

alejandroEsc:


charts/redpanda/servicemonitor.go
```diff
@@ -0,0 +1,97 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
+
+	"github.com/redpanda-data/helm-charts/pkg/gotohelm/helmette"
+)
+
+func ServiceMonitor(dot *helmette.Dot) *monitoringv1.ServiceMonitor {
+	values := helmette.Unwrap[Values](dot.Values)
+
+	if !ptr.Deref(values.Monitoring.Enabled, false) {
```
chrisseto, line 30: I just added support for this yesterday! I can pull it out of the PR it's in so we can get it in faster.

charts/redpanda/servicemonitor.go
```diff
@@ -0,0 +1,97 @@
...
+	monitorLabels := helmette.Merge(FullLabels(dot), values.Monitoring.Labels)
+
+	matchLabels := map[string]string{
+		"monitoring.redpanda.com/enabled": "true",
+		"app.kubernetes.io/name":          Name(dot),
+		"app.kubernetes.io/instance":      dot.Release.Name,
+	}
+
+	tlsConfigMap := values.Monitoring.TLSConfig
+	var tslConfig *monitoringv1.TLSConfig = nil
```
chrisseto, line 43: No need for `= nil`?
```suggestion
	var tslConfig *monitoringv1.TLSConfig
```

charts/redpanda/servicemonitor.go
```diff
@@ -0,0 +1,97 @@
...
+	if len(tlsConfigMap) > 0 {
+		tslConfig = &monitoringv1.TLSConfig{
+			SafeTLSConfig: monitoringv1.SafeTLSConfig{
+				InsecureSkipVerify: false,
+				// TODO, do we support this?
+				// .Values.monitoring.tlsConfig general enough
+			},
+			CAFile:   valueOrDefault(tlsConfigMap, "caFiles", ""),
```
chrisseto, line 51: Seems like this logic could be made easier by typing the `tlsConfig` as a `monitoringv1.TLSConfig`. Is there any reason to not do so?

charts/redpanda/values.go
```diff
@@ -157,6 +157,7 @@ type Monitoring struct {
 	ScrapeInterval string            `json:"scrapeInterval" jsonschema:"required,pattern=.*[smh]$"`
 	Labels         map[string]string `json:"labels"`
 	TLSConfig      map[string]any    `json:"tlsConfig"`
+	EnableHttp2    *bool             `json:"enabled" jsonschema:"required"`
```
chrisseto, line 160: The json name of this field feels _very_ misleading to me.

charts/redpanda/servicemonitor.go
alejandroEsc: that definitely would help, i asked, let me make that change and see what happens.

charts/redpanda/values.go
alejandroEsc: ahh thats me typing too fast and changing context, ill fix it.

chrisseto:
@alejandroEsc could you clean up the commit history? Looks like you've pulled in a few things from `main` that are muddying the diff.

alejandroEsc:
> @alejandroEsc could you clean up the commit history? Looks like you've pulled in a few things from `main` that are muddying the diff.

maybe i did, but i had not even rebased when those showed up, not sure what happened

charts/redpanda/values.go
```diff
@@ -168,10 +169,11 @@ type Logging struct {
 }
 
 type Monitoring struct {
-	Enabled        *bool             `json:"enabled" jsonschema:"required"`
-	ScrapeInterval string            `json:"scrapeInterval" jsonschema:"required,pattern=.*[smh]$"`
-	Labels         map[string]string `json:"labels"`
-	TLSConfig      map[string]any    `json:"tlsConfig"`
+	Enabled        *bool                   `json:"enabled" jsonschema:"required"`
+	ScrapeInterval monitoringv1.Duration   `json:"scrapeInterval" jsonschema:"required"`
```
chrisseto, line 173: I think we'll want to add a special case to `genschema` via the `Mapper` function to add the pattern back to the JSON schema.

charts/redpanda/values.go
```diff
@@ -168,10 +169,11 @@ type Logging struct {
 }
 
 type Monitoring struct {
-	Enabled        *bool             `json:"enabled" jsonschema:"required"`
-	ScrapeInterval string            `json:"scrapeInterval" jsonschema:"required,pattern=.*[smh]$"`
-	Labels         map[string]string `json:"labels"`
-	TLSConfig      map[string]any    `json:"tlsConfig"`
+	Enabled        *bool                   `json:"enabled" jsonschema:"required"`
```
chrisseto, line 172: This can be a `bool` instead of a pointer. It's marked as required and there's a default value in values.yaml.

charts/redpanda/servicemonitor.go
```diff
@@ -0,0 +1,79 @@
...
+			Labels:    monitorLabels,
+		},
+		Spec: monitoringv1.ServiceMonitorSpec{
+			Endpoints: []monitoringv1.Endpoint{
+				{
+					Interval:    values.Monitoring.ScrapeInterval,
+					Path:        "/public_metrics",
+					Port:        "admin",
+					EnableHttp2: values.Monitoring.EnableHttp2,
+					Scheme:      "https",
```
chrisseto, line 70: Should scheme be `http` if `tlsConfig` is nil?

alejandroEsc: its never really nil, if the user does not enter anything in tlsConfig one with insecureVerify will be created for the user.

cmd/genschema/main.go
```diff
@@ -51,6 +52,10 @@ func main() {
 						corev1.FSGroupChangeAlways,
 					},
 				}
+			case monitoringv1.Duration:
+				return &jsonschema.Schema{
+					Type: "string",
```
chrisseto, line 57: Can this also include a `Pattern` for duration strings? You can probably pull a good regex from the docs on `monitoringv1.Duration`.

charts/redpanda/servicemonitor.go
chrisseto: Isn't it possible to disable HTTPS/TLS on the admin endpoint though?

charts/redpanda/servicemonitor.go
```diff
@@ -0,0 +1,74 @@
...
+	matchLabels := map[string]string{
+		"monitoring.redpanda.com/enabled": "true",
+		"app.kubernetes.io/name":          Name(dot),
+		"app.kubernetes.io/instance":      dot.Release.Name,
+	}
+
+	tlsConfig := values.Monitoring.TLSConfig
+
+	if tlsConfig != nil {
+		tlsConfig.SafeTLSConfig.InsecureSkipVerify = false
```
chrisseto, line 43: This line is very confusing to me. The default of `InsecureSkipVerify` is false, why are we explicitly setting it here? Shouldn't we just accept the default or whatever the user has provided?

chrisseto (commented):
Ah, there's one missing part here 😓 Sorry there's not great documentation here.

You'll need to modify the original `servicemonitor.yaml` to instead call out to the go function. There's not currently support for marking a function as an "entrypoint" or anything like that. `service.nodeport.yaml` has an example of what that looks like.

Could we also add some more tests for this by creating some `novalues.yaml` files?

cmd/genschema/main.go
alejandroEsc: ahh ok, i was wondering how deep we should go with this.

charts/redpanda/servicemonitor.go
chrisseto: Ah, I didn't realize this would be keeping up the existing behavior. We can leave it as is for now. Can you add a note that this was pre-existing behavior and is likely less than ideal?

chrisseto (approved):
LGTM!

Looks like you'll need to run `go mod tidy` in the `testdata` directory and you've got an interesting test failure that I'm not immediately recognizing.

chrisseto:
closing in favor of https://github.com/redpanda-data/helm-charts/pull/1313

PR #1252, opened 2024-04-25
gotohelm: fully bootstrap _shims.tpl

chrisseto:
#### 63544e0500546c6c084d39957a72a1582da0b713 gotohelm: fully bootstrap _shims.tpl

Previously, gotohelm relied on `go test` to generate `_shims.tpl` as an
artifact which would then be hoisted into it's package by an external
script. The process was never quite right and required re-running the
entire process 1-3 times to actually get everything up to date.

This commit makes it possible to generate _shims.tpl on the fly by
embedding the bootstrap package's source into `gotohelm` and doing some
shenanigans with global variables. The result removes the possibility of
_shims.tpl ever being out of date and additionally makes the bootstrap
package much more pleasant to work within as it exists in the same go
module as the rest of the helm-charts.


alejandroEsc (approved):
nice bootstrapping conceptually helps me understand what is going on better.

PR #1255, opened 2024-04-26
nix: remove nix builds of ./cmd tools

chrisseto:
#### e19ff60350b45a864ab835375f9643ae1e5f31af nix: remove nix builds of ./cmd tools

Previously, nix was being utilized to build `genparital`, `genschema`,
and `gotohelm`. Doing so allowed us to leverage the nix magic cache in
CI and made it easy to work with our tooling in a `nix develop` shell.

Due to the heavy development on said tooling, nix ended up being more of
a burden during local development. `nix run` incurred a small startup
overhead, built packages slower than `go run`, and required too much
fiddling with source files and vendor hashes to be worthwhile.

This commit replaces all instances of `nix run .#` with their equivalent
`go run`.


alejandroEsc (approved):
lgtm

PR #1257, opened 2024-04-29
Add new managed identity configs for Azure Tiered Storage

JakeSCahill:


RafalKorepta (approved):

chrisseto (approved):
LGTM, though as @RafalKorepta has pointed out there's some generated files that need to be updated.

PR #1259, opened 2024-04-29
redpanda: add security context to fs-validator

chrisseto:
Prior to this commit, the fs-validator was the only initcontainer that
did not have securityContext specified on it. This commit adds it so the
redpanda STS renders as expected.


RafalKorepta (approved):
LGTM, but there are more init containers without security context.

P.S. This improvement would not be released until Chart.yaml version is not bumped.

c4milo (approved):

RafalKorepta (approved):

PR #1281, opened 2024-05-15
gh: Added Jira issue syncing workflows

michael-redpanda:
Added workflows that will sync GH issues to Jira


chrisseto (approved):
LGTM just need to add some `"`'s around instances of bash variable substitutions to pass CI.

PR #1290, opened 2024-05-20
pkg/helm: run `helm template` directly instead of via subprocess

chrisseto:
#### b705eabf81fbfd72148c08d08932e8976b02cebb pkg/helm: run `helm template` directly instead of via subprocess

This commit modifies helm.Client.Template to run via depending on Helm as a
library rather than run through a subprocess. This provides a greater degree of
control over the helm execution environment but also reduces the overhead of
running `helm template` many times.

Note: The actual performance gain of this commit hasn't truly been measured.
The bulk of the time was wasted in `deepCopy` calls, this commit was required
to appropriately profile the execution of `helm template`.


pkg/helm/helm.go
```diff
@@ -239,32 +262,95 @@ type TemplateOptions struct {
...
+	// fake KubeClient.
+	client := action.NewInstall(c.config)
+
+	client.APIVersions = chartutil.DefaultVersionSet
+	client.ClientOnly = true
+	client.DryRun = true
+	client.GenerateName = opts.GenerateName
+	client.Namespace = opts.Namespace
+	client.ReleaseName = opts.Name
+	client.Replace = true // Skips "name" checks.
```
c4milo, line 279: are we sure we want this default value in production?

chrisseto: This is borrowed directly from helm itself. `helm template` needs to run in an isolated environment. Rather than having that logic baked in, they opted to use this as a work around. I'll add a back link and a note saying as much. I could have sworn that I had a few links peppered in there already but looks pretty scant.

RafalKorepta (approved):

PR #1296, opened 2024-05-22
gotohelm: add support for helm's `lookup` function

chrisseto:
This commit adds support for the `lookup` function from helm. It does so by injecting a Kubernetes API client into the `helmette.Dot` struct to avoid relying on global variables. This is a divergence from helm itself but is thoroughly documented as such.

Unlike helm's `lookup`, gotohelm returns `nil, false` in the case of not found errors rather than an empty dictionary.


pkg/gotohelm/transpiler_test.go
```diff
@@ -101,16 +129,19 @@ func TestTranspile(t *testing.T) {
...
 
-	go func() {
-		require.NoError(t, runner.Run(ctx))
-	}()
+	runner := NewGoRunner(t, td)
 
 	for _, pkg := range pkgs {
 		pkg := pkg
 		t.Run(pkg.Name, func(t *testing.T) {
+			ctx := testutil.Context(t)
```
RafalKorepta, line 143: NIT: Why context is created within loop? Could there be single context that is capped by the go test framework?

pkg/gotohelm/transpiler_test.go
```diff
@@ -285,11 +342,26 @@ func NewGoRunner(root string) *GoRunner {
...
 	)
 
-	return &GoRunner{
+	runner := &GoRunner{
 		cmd:      cmd,
 		inputCh:  make(chan *helmette.Dot),
 		outputCh: make(chan map[string]any),
 	}
+
+	ctx, cancel := context.WithCancel(context.Background())
```
RafalKorepta, line 351: NIT: Could the context be based on context created in https://github.com/redpanda-data/helm-charts/blob/102a2faf6294c9092a9ffc9f39c05a204da66b20/pkg/gotohelm/transpiler_test.go#L143 ?

RafalKorepta (approved):

pkg/gotohelm/transpiler_test.go
chrisseto: Hoisted.

pkg/gotohelm/transpiler_test.go
chrisseto: This is a weird case. It's done this way to ensure that any errors get appropriately piped to the `testing.T` so there aren't silent failures. You could plumb another context in but it would be doing the same job as the `t.Cleanup` call but without the correct error propagation.

PR #1298, opened 2024-05-22
redpanda: move golden files into txtars

chrisseto:
Previously, all golden files where kept as individual files in the `testdata` directory. This would result in very noisy diffs on Github upon most changes to the helm chart.

This commit adds functionality to run golden tests using a `txtar` archive which is a plain text "tar" archive. This limits diffs to instead be to a small set of very large files but preserves the ability to inspect diffs in plain text.


chrisseto:
Hmmmm. So this is interesting. The txtar diff is now too long to be displayed on github which is certainly a major drawback. We could potentially chunk the txtar file so it doesn't exceed github's limits? WDYT @RafalKorepta? Worth it or should we stick with flat files?

RafalKorepta (approved):
I'm not sure what limits for which dimension (bytes or line number) we would hit. We can try and revert if it would not work.

chrisseto:
Hmmmmmm. Let's hold off on merging this for a minute. I keep having trouble getting the txtars to update for some reason. I thought I had fixed the issue but it keeps coming back. Going to move this into a draft state and revisit it later.

chrisseto:
Ah, might have figure it out. I think I messed up a loop and wasn't capturing a value correctly. Will still revisit this a bit later though because I found the issue in a conflict with a different patch.

PR #1305, opened 2024-05-22
Handle for statement in transpiler

RafalKorepta:
Current go template does not have for statement iteration. To work around
the range loop using `until` buildin function. As the `until` counts from 0,
the `for` loop needs to have integer iterator that starts from 0.


chrisseto (approved):
LGTM! Thanks for extracting this one!

PR #1307, opened 2024-05-23
Fix great or equal and less or equal binary expression

RafalKorepta:
Go template supports great or equal and less or equal with different name 'ge' and 'le' respectively.

Reference

* https://pkg.go.dev/text/template
* https://github.com/golang/go/blob/62711d61e5ba3a6c42bc57a28c2402dd8669261f/src/text/template/doc.go#L380-L394
* https://github.com/golang/go/blob/62711d61e5ba3a6c42bc57a28c2402dd8669261f/src/text/template/funcs.go#L55-L61

### DIFF
--- a/pkg/gotohelm/testdata/src/example/syntax/syntax.go
+++ b/pkg/gotohelm/testdata/src/example/syntax/syntax.go
@@ -28,9 +28,6 @@ func Syntax() map[string]any {
 	_ = len([]int{})
 	_ = len(map[string]any{})
 
-	// BinaryExprs
-	binaryExprs()
-
 	// ParenExpr
 	_ = (true)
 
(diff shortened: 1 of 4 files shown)


pkg/gotohelm/testdata/src/example/syntax/syntax.go
```diff
@@ -133,4 +131,15 @@ func binaryExprs() {
 	_ = map[string]any{} != nil
 	// TODO strings
 	// TODO floats
+
+	result := []bool{}
```
chrisseto, line 135: super nit: I think it's a bit preferable to make one big literal:
```go
results := []bool{
  1 > 2,
  1 < 2,
  1 >= 2,
// etc
}
```

chrisseto (approved):
LGTM!

PR #1308, opened 2024-05-23
Add first function from helm chart engine

RafalKorepta:
To handle Redpanda configuration in go code the `first` function needs to be included to helmette as part of the sprig definition.

Negative numbers are recognized as `ast.UnaryExpr` with sub token. Now they can be transpiled.

P.S. The `[]byte` array can be represented only in Go, but not in template.

### DIFF
--- a/pkg/gotohelm/helmette/sprig.go
+++ b/pkg/gotohelm/helmette/sprig.go
@@ -14,6 +14,20 @@ import (
 	"golang.org/x/exp/maps"
 )
 
+// First function can not return `T` as sprig implementation
+// will return nil if array is of the size 0.
+//
+// # Reference
+// https://github.com/Masterminds/sprig/blob/581758eb7d96ae4d113649668fa96acc74d46e7f/list.go#L161-L163
+// +gotohelm:builtin=first
+func First[T any](x []T) any {
+	if len(x) == 0 {
+		return nil
+	}
+
+	return x[0]
+}
+
 // TrimPrefix is the go equivalent of sprig's `trimPrefix`
 // +gotohelm:builtin=trimPrefix
 func TrimPrefix(prefix, s string) string {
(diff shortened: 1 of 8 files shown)


pkg/gotohelm/transpiler.go
```diff
@@ -674,6 +674,10 @@ func (t *Transpiler) transpileExpr(n ast.Expr) Node {
 		case token.AND:
 			// Can't take addresses in templates so just return the variable.
 			return t.transpileExpr(n.X)
+		case token.SUB:
```
chrisseto, line 677: This change seems unrelated? Bundling multiple commits into a PR is fine by me but I will ask for tests :D

pkg/gotohelm/helmette/sprig.go
```diff
@@ -14,6 +14,15 @@ import (
 	"golang.org/x/exp/maps"
 )
 
+// +gotohelm:builtin=first
+func First[T any](x []T) any {
+	if len(x) == 0 {
+		return nil
```
chrisseto, line 20: ~Does this work? `nil` would be an invalid return if `T` is any primitive (string, int). I would expect the go compiler to yell at you.~ OH, you're returning `any`.

Depending on what sprig does internally, it might be preferable to make this `func[T any]([]T) T`? The syntax is a bit weird to make it work though
```go
if len(x) == 0 {
  var zero T
  return zero
}


```

chrisseto (approved):
LGTM, just a quick Q about sprigs behavior and a stray change.

pkg/gotohelm/helmette/sprig.go
RafalKorepta: That was my initial implementation, but then go code gives different output. Please see

https://github.com/Masterminds/sprig/blob/581758eb7d96ae4d113649668fa96acc74d46e7f/list.go#L163

To match implementation I would leave it as is, WDYT?

chrisseto: Ah. Hm. That's a tricky one. Mind adding that link as a comment for future reference? It might be painful to deal with `any` but we can also make improvements in the future if need be.

RafalKorepta: Yes, I will add comment.

PR #1315, opened 2024-05-23
Add trimSuffix function to helmette

RafalKorepta:
`trimSuffix` function is required for Redpanda configuration conversion.

Reference

https://github.com/redpanda-data/helm-charts/blob/9d0f4fcf9929c8c8574d8cd5137da1bd7b7e94e3/charts/redpanda/templates/_helpers.tpl#L169

### DIFF
--- a/pkg/gotohelm/helmette/sprig.go
+++ b/pkg/gotohelm/helmette/sprig.go
@@ -41,10 +41,22 @@ func First[T any](x []T) any {
 	return x[0]
 }
 
+// TrimSuffix is the go equivalent of sprig's `trimSuffix`
+// +gotohelm:builtin=trimSuffix
+func TrimSuffix(suffix, s string) string {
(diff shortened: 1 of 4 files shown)


pkg/gotohelm/helmette/sprig.go
```diff
@@ -12,8 +12,19 @@ import (
 
 	"github.com/imdario/mergo"
 	"golang.org/x/exp/maps"
+	"gopkg.in/yaml.v3"
 )
 
+// ToYAML is the go equivalent of sprig's `toYaml`.
```
chrisseto, line 18: super nit: apparently `toYaml` is from helm not sprig. Sprig only provides toJSON.

chrisseto (approved):
Looks like `toYaml` slipped into here but still LGTM!

PR #1327, opened 2024-05-29
Further indent podAntiAffinity matchLabels for soft anti-affinity

kyle-render:
Currently, it ends up on the same depth as `matchLabels:` in my testing running `helm template`:

```
        podAntiAffinity:
          preferredDuringSchedulingIgnoredDuringExecution:
          - weight: 100
            podAffinityTerm:
              topologyKey: kubernetes.io/hostname
              labelSelector:
                matchLabels:
                app.kubernetes.io/component: redpanda-statefulset
                app.kubernetes.io/instance: redpanda
```

Which kubernetes seems to accept and interpret correctly, but when I run this yaml through any post-processing (eg `yq`) it "fixes" `matchLabels:` not having a value and sets it to `null`, causing the labels to be interpreted as keys under `labelSelector` instead, which kubernetes rejects as not matching the sts spec.

### DIFF
--- a/charts/redpanda/templates/_statefulset.tpl
+++ b/charts/redpanda/templates/_statefulset.tpl
@@ -91,7 +91,7 @@ podAntiAffinity:
     podAffinityTerm:
       topologyKey: {{ .Values.statefulset.podAntiAffinity.topologyKey }}
       labelSelector:
-        matchLabels: {{ include "statefulset-pod-labels-selector" . | nindent 8 }}
+        matchLabels: {{ include "statefulset-pod-labels-selector" . | nindent 10 }}
   {{- else if eq .Values.statefulset.podAntiAffinity.type "custom" -}}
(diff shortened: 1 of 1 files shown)


kyle-render:
> Strange that yq is less permissive than Kubernetes' (go's) YAML parser.

Yeah, this one is fascinating for sure. `yq` reads it fine, but makes a different assumption about the ambiguous indentation and assumes this means `matchLabels: null`, which is what comes out when you run this through `yq`.

I'll try to update the tests!

kyle-render:
Hm, the tests all appear to have the correct (2 space) indentation already, and this makes them 4 spaced. Could be something on my side? I'll try to dig in more.

chrisseto:
Just as a quick double check, which version of helm and the chart have you been working with?

chrisseto:
Ah, gotcha! I wrote up an issue to get some tests on this matter as I'm guessing there are other cases of this floating around. Thanks for the fix and for bringing this to our attention!

PR #1338, opened 2024-06-06
Handle instance methods with pointer receiver

### LINKED ISSUE #1329: gotohelm: add support for instance methods
gotohelm does not currently support transpiling instance methods. As we're starting to creep over the ~50% margin, we're starting to notice that having a lot of functions that all accept `helmette.Dot` as the only argument isn't exactly that easy to maintain or test.

We _should_ be able to get an MVP in place that turns instance methods into namespaced `define` blocks that accept the receiver as the first argument. For simplicity sake, we should ban non-pointer receivers to start. We can likely add support by just `deepCopy`'ing the receiver.

Chicken scratch implementation ideas:
```go
type Values struct {
}

func (v *Values) SomeMethod() bool {
    return true
}
```

would transpile to

```
{{ define "chart.Values.SomeMethod" }}
{{ $v := /* Pop the first argument */ }}
{{ end }}

/* Usage */
{{ include "chart.Values.SomeMethod" (dict "a" (list $values) | toJSON) }}
```

JIRA Link: [K8S-242](https://redpandadata.atlassian.net/browse/K8S-242)

RafalKorepta:
Instance methods are transpiled into templates that receives as a first argument the instance of the struct. Any function call that have pointer receiver would be translated to template `include` statement with one argument which is struct itself.

There is special case for Values.AsMap method as it is not defined as seperated template, but rather it is build in feature of the helm engine.

In future improvements instance methods transpilation could support arguments.

Fixes #1329

### DIFF
--- a/pkg/gotohelm/transpiler.go
+++ b/pkg/gotohelm/transpiler.go
@@ -148,8 +148,20 @@ func (t *Transpiler) transpileFile(f *ast.File) *File {
 		}
 
 		funcDirectives := parseDirectives(fn.Doc.Text())
+		// To not clash with the same method name in the same package
+		// which can be declared in multiple struct the package and function
+		// name could be separated with the name of the struct type.
+		// package example
+		//
+		// func FunExample() {} => {{- define "example.FunExample" -}}
+		//
+		// func (e *Example) MethodExample() {} => {{- define "example.Example.MethodExample" -}}
 		if _, ok := funcDirectives["name"]; !ok {
-			funcDirectives["name"] = fn.Name.String()
+			funName := fn.Name.String()
+			if fn.Recv != nil {
+				funName = fmt.Sprintf("%s.%s", baseTypeName(fn.Recv.List[0].Type), funName)
+			}
+			funcDirectives["name"] = funName
 		}
 
 		if _, ok := funcDirectives["sprig"]; ok {
@@ -157,6 +169,14 @@ func (t *Transpiler) transpileFile(f *ast.File) *File {
 		}
 
 		var params []Node
+		if fn.Recv != nil {
+			for _, param := range fn.Recv.List {
+				for _, name := range param.Names {
+					params = append(params, t.transpileExpr(name))
+				}
+			}
+		}
+
 		for _, param := range fn.Type.Params.List {
 			for _, name := range param.Names {
 				params = append(params, t.transpileExpr(name))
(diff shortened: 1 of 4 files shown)


pkg/gotohelm/transpiler.go
```diff
@@ -620,8 +631,8 @@ func (t *Transpiler) transpileExpr(n ast.Expr) Node {
 			for _, f := range t.getFields(typ.Underlying().(*types.Struct)) {
 				if f.Field.Name() == n.Sel.Name {
 					return &Selector{
-						Expr:  t.transpileExpr(n.X),
-						Field: f.JSONName(),
+						Expr:    t.transpileExpr(n.X),
```
chrisseto, line 634: We need to get CI to start running `go fmt` or `gofumpt` 😬

pkg/gotohelm/transpiler.go
```diff
@@ -858,15 +869,28 @@ func (t *Transpiler) transpileCallExpr(n *ast.CallExpr) Node {
 	}
 
 	// Method call.
-	if callee.Type().(*types.Signature).Recv() != nil {
+	if r := callee.Type().(*types.Signature).Recv(); r != nil {
```
chrisseto, line 872: I think it might be better to move this below the really big switch statement as that would let us more easily special case `Values.AsMap`.

In the numerics PR (I think) there's a change that turns the `id` variable into something like: `github.com/redpanda-data/helm-charts/pkg/gotohelm/helmette.(*Values).AsMap`. That (roughly) matches the way that `reflect` shows functions as strings and is preferable IMO. WDYT?

pkg/gotohelm/transpiler.go
```diff
@@ -858,15 +869,28 @@ func (t *Transpiler) transpileCallExpr(n *ast.CallExpr) Node {
...
+				// In helm the AsMap method could be used for helm Values
+				return t.transpileExpr(n.Fun)
+			}
+		}
+
+		if typeName, ok := r.Type().(*types.Pointer); ok {
+			if baseTypeName, ok := typeName.Elem().(*types.Named); ok {
+				return &Call{
+					FuncName:  fmt.Sprintf("%s.%s.%s", t.Package.Name, baseTypeName.Obj().Name(), callee.Name()),
+					Arguments: []Node{t.transpileExpr(n.Fun.(*ast.SelectorExpr).X)},
```
chrisseto, line 889: Can you copy the original comment about method calls to be above this line? 

Here's a slightly more clear version of it:

```
// Method calls come in as a "top level" CallExpr where .Fun is the
// selector up to that call. e.g. `Foo.Bar.Baz()` will be a `CallExpr`.
// It's `.Fun` is a `SelectorExpr` where `.X` is `Foo.Bar`, the receiver,
// and `.Sel` is `Baz`, the method name.
```

chrisseto (approved):
LGTM just a cuple notes!

pkg/gotohelm/transpiler.go
```diff
@@ -148,7 +148,11 @@ func (t *Transpiler) transpileFile(f *ast.File) *File {
 
 		funcDirectives := parseDirectives(fn.Doc.Text())
 		if _, ok := funcDirectives["name"]; !ok {
-			funcDirectives["name"] = fn.Name.String()
+			if fn.Recv != nil {
+				funcDirectives["name"] = fmt.Sprintf("%s.%s", fn.Recv.List[0].Type.(*ast.StarExpr).X.(*ast.Ident), fn.Name.String())
```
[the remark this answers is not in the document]
RafalKorepta: > This cast is pretty brittle. It might be better to use typeOf here to get the function type as it'll be a bit easier to resolve the receiver to a named type.

I totally agree it's brittle. In the next PR it is a bit improved https://github.com/redpanda-data/helm-charts/pull/1343/files#diff-5fd16cbd6eecffe68e88c97a779873344d516695142c55bb3a83e55653253df8R151-R157

> We can keep it as is for now but if it ever causes us grief, we should go down the type route rather than the AST route. 

I could improve this in this spot, but I would need more guidance about usage of `typeOf`.

> Also worth noting that the name directive is going to behave a bit strangely for methods as it will result in the method not being namespaced under the type name. It's not likely to be a problem but might be worth a quick note?

I agree to add a note.

pkg/gotohelm/transpiler.go
RafalKorepta: > That (roughly) matches the way that reflect shows functions as strings and is preferable IMO. WDYT?

I agree. Should I wait to rebase on top of https://github.com/redpanda-data/helm-charts/pull/1332?

chrisseto (approved):

PR #1342, opened 2024-06-07
Add support for instance methods with parameters

RafalKorepta:
Handle method parameters in transpiler.

Based on https://github.com/redpanda-data/helm-charts/pull/1338


pkg/gotohelm/transpiler.go
```diff
@@ -160,18 +160,18 @@ func (t *Transpiler) transpileFile(f *ast.File) *File {
...
-			}
-		}
 		if fn.Recv != nil {
 			for _, param := range fn.Recv.List {
 				for _, name := range param.Names {
 					params = append(params, t.transpileExpr(name))
 				}
 			}
 		}
+		for _, param := range fn.Type.Params.List {
```
chrisseto, line 170: super nit: Add a new line before the `for` loop. The code feels "cramped" otherwise.

You can ignore my comment about the ordering in the previous PR :D

chrisseto (approved):
LGTM

PR #1343, opened 2024-06-07
Add support for instance methods without pointer receiver

### LINKED ISSUE #1329: gotohelm: add support for instance methods

RafalKorepta:
Handle named receiver methods that can not mutate object.

Based on:
* https://github.com/redpanda-data/helm-charts/pull/1338
* https://github.com/redpanda-data/helm-charts/pull/1342

Fixes #1329

### DIFF
--- a/pkg/gotohelm/transpiler.go
+++ b/pkg/gotohelm/transpiler.go
@@ -947,20 +947,37 @@ func (t *Transpiler) transpileCallExpr(n *ast.CallExpr) Node {
 	if callee.Pkg().Path() == t.Package.PkgPath {
 		// Method call.
 		if r := callee.Type().(*types.Signature).Recv(); r != nil {
-			if typeName, ok := r.Type().(*types.Pointer); ok {
-				if baseTypeName, ok := typeName.Elem().(*types.Named); ok {
-					return &Call{
-						FuncName: fmt.Sprintf("%s.%s.%s", t.Package.Name, baseTypeName.Obj().Name(), callee.Name()),
-						// Method calls come in as a "top level" CallExpr where .Fun is the
-						// selector up to that call. e.g. `Foo.Bar.Baz()` will be a `CallExpr`.
-						// It's `.Fun` is a `SelectorExpr` where `.X` is `Foo.Bar`, the receiver,
-						// and `.Sel` is `Baz`, the method name.
-						Arguments: append([]Node{t.transpileExpr(n.Fun.(*ast.SelectorExpr).X)}, args...),
-					}
+			typ := r.Type()
+
+			mutable := false
+			switch typ.(type) {
+			case *types.Pointer:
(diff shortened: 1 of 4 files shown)


pkg/gotohelm/transpiler.go
```diff
@@ -885,6 +889,11 @@ func (t *Transpiler) transpileCallExpr(n *ast.CallExpr) Node {
 					Arguments: append([]Node{t.transpileExpr(n.Fun.(*ast.SelectorExpr).X)}, args...),
 				}
 			}
+		} else if typeName, ok := r.Type().(*types.Named); ok {
```
chrisseto, line 892: I should probably just make a helper for this but I've been writing this type of check as:

```go
typ := r.Type()

// "unwrap" the pointer type.
switch typ.(type) {
case *types.Pointer:
typ = typ.Elem()
}

if _, ok := typ.(TypeYouActuallyWant); ok {
// ...
}
```

pkg/gotohelm/transpiler.go
```diff
@@ -885,6 +889,11 @@ func (t *Transpiler) transpileCallExpr(n *ast.CallExpr) Node {
...
+			return &Call{
+				FuncName:  fmt.Sprintf("%s.%s.%s", t.Package.Name, typeName.Obj().Name(), callee.Name()),
+				Arguments: append([]Node{t.transpileExpr(n.Fun.(*ast.SelectorExpr).X)}, args...),
```
chrisseto, line 895: In the case where the receiver is _not_ a pointer. I think we'll need to wrap the receiver parameter in `deepCopy` or something similar.

For example: https://go.dev/play/p/WXUgEvJhCLJ

Let's also add a test for structs mutating themselves in methods so there are no surprises later down the road.

RafalKorepta: Agree, I will add test.

chrisseto (approved):

PR #1347, opened 2024-06-12
redpanda: convert `service.loadbalancer.yaml` to go

### LINKED ISSUE #1215: redpanda: Convert `service.loadbalancer.yaml` to go code
JIRA Link: [K8S-161](https://redpandadata.atlassian.net/browse/K8S-161)

[K8S-161]: https://redpandadata.atlassian.net/browse/K8S-161?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

chrisseto:
Fixes #1215


charts/redpanda/service.loadbalancer.go
```diff
@@ -0,0 +1,175 @@
...
+
+			fallbackPorts := append(listener.AdvertisedPorts, values.Listeners.Admin.Port)
+
+			ports = append(ports, corev1.ServicePort{
+				Name:     fmt.Sprintf("admin-%s", name),
+				Protocol: corev1.ProtocolTCP,
+				// TODO this diverges from all the other cases that use
+				// listener.Port. Is this a bug in the original template?
+				TargetPort: intstr.FromInt32(values.Listeners.Admin.Port),
+				Port:       ptr.Deref(listener.NodePort, fallbackPorts[0]),
```
RafalKorepta, line 100: I wonder how `values.Listeners.Admin.External` could be deterministic with ranging and always returning the same port. 

Would there be a case where running helm template in a loop without changing values input would generate different loadbalancer manifest?

charts/redpanda/service.loadbalancer.go
```diff
@@ -0,0 +1,175 @@
...
+			annotations[k] = v
+		}
+
+		if externalDNS.Enabled {
+			prefix := podname
+			if len(values.External.Addresses) > i {
+				prefix = values.External.Addresses[i]
+			}
+
+			address := fmt.Sprintf("%s.%s", prefix, helmette.Tpl(*values.External.Domain, dot))
```
RafalKorepta, line 71: The `helmette.Tpl` is not implemented, so that go implementation is incomplete.

RafalKorepta (approved):
LGTM

charts/redpanda/service.loadbalancer.go
chrisseto: Added an implementation of `Tpl`!

charts/redpanda/service.loadbalancer.go
chrisseto: Ah, I'm actually going to punt on this change for now. The nodeport service works the same way and templates' `range`s are actually deterministic.

PR #1353, opened 2024-06-13
redpanda: partially migrate STS volumes/mounts

chrisseto:
This commit migrates the easiest of the redpanda volumes and volume mounts into go code. This change is in preparation of an upcoming change to the `truststore_file` that will allow specifying a ConfigMap key to be mounted.


RafalKorepta (approved):
LGTM

PR #1268, opened 2024-05-05
Move config map to go

RafalKorepta:
#### 75f28bb55a11672c1a6b9acb26e342533dad4a03 Generate values to helm template

Methods improves readability of Redpanda configuration.

`IsAccessKeyReferenceValid` and `IsSecretKeyReferenceValid` functions are not used.

#### e0fad8c2258a8fc14205cce7276863ff2206f761 Fix `for` loops using selector expression

Previous implementation would panic as type assertion was hardcoded for `ast.Ident`.

#### 6df252f0843f945c9a5798dfbd5c4e60f1862e2d Add Monitoring coreos.com custom resources to the scheme

When Prometheus Redpanda helm chart dependency is enabled, then DecodeYAMLFrom
is not working due to missing ServiceMonitor custom resource in scheme.
```
=== CONT  TestConfigMap/14-prometheus-no-tls-values.yaml
    chart_test.go:182:
        	Error Trace:	/Users/rafalkorepta/workspace/redpanda/helm-charts/charts/redpanda/chart_test.go:182
        	Error:      	Received unexpected error:
        	            	no kind "ServiceMonitor" is registered for version "monitoring.coreos.com/v1" in scheme "pkg/runtime/scheme.go:100"
        	Test:       	TestConfigMap/14-prometheus-no-tls-values.yaml
    --- FAIL: TestConfigMap/14-prometheus-no-tls-values.yaml (0.08s)
```

#### 97a80f2869c96a4a609b98fecfeb3552e40292f8 Remove old config map implementation

Update Statefulset config map hash calculation

#### 9ef78e3c73c156f17b8051534dd1c0f06a5da340 Handle overprovisioned differently from the desire implementation

Changes to the input values are anti-pattern and relay on order of execution.
Our test shows that overprovioned are incorrectly set when CPU is bellowed
`1000m`.

Reference
(overproviosned is set to false)[https://github.com/redpanda-data/helm-charts/blob/5e30ecbaf15da44691402f6128194d6639652845/charts/redpanda/testdata/ci/22-gke-tiered-storage-with-creds-values.yaml.tpl.golden#L481]
https://github.com/redpanda-data/helm-charts/blob/5e30ecbaf15da44691402f6128194d6639652845/charts/redpanda/ci/22-gke-tiered-storage-with-creds-values.yaml.tpl#L33
[...]

PR #1357, opened 2024-06-14
redpanda: configmap refactors

chrisseto:
These two commits encapsulate some of the more difficult checks of `Enabled` fields and migrates most config generation onto the listener structs. This is in preparation for allowing customization of the `truststore_file`.


RafalKorepta:
I incorporated all improvements from this PR into https://github.com/redpanda-data/helm-charts/pull/1268

PR #1358, opened 2024-06-17
Handle overprovisioned flag based on CPU values

RafalKorepta:
Overprovisioned till Redpanda chart version 5.8.8 should be
set to `true` when Statefulset CPU request value is bellow 1000 mili cores.

Function `redpanda-smp`, that should overwrite `overprovisioned` flag,
was not called before setting `overprovisioned` flag.

Reference

redpanda-smp template - https://github.com/redpanda-data/helm-charts/blob/5f287d45a3bda2763896840e505fb3de82b968b6/charts/redpanda/templates/_helpers.tpl#L187
redpanda-smp template invocation - https://github.com/redpanda-data/helm-charts/blob/5f287d45a3bda2763896840e505fb3de82b968b6/charts/redpanda/templates/_configmap.tpl#L610
overprovisioned flag - https://github.com/redpanda-data/helm-charts/blob/5f287d45a3bda2763896840e505fb3de82b968b6/charts/redpanda/templates/_configmap.tpl#L607

### DIFF
--- a/charts/redpanda/chart_test.go
+++ b/charts/redpanda/chart_test.go
@@ -4,6 +4,7 @@ import (
 	"maps"
 	"os"
 	"path/filepath"
+	"regexp"
 	"testing"
 
 	"github.com/redpanda-data/helm-charts/charts/redpanda"
@@ -201,6 +202,16 @@ func TestConfigMap(t *testing.T) {
 
 			newConf, err := extractRedpandaConfigsFromConfigMap(manifests)
 			require.NoError(t, err)
+
+			// Overprovisioned till Redpanda chart version 5.8.8 should be
+			// set to `true` when Statefulset CPU request value is bellow 1000 mili cores.
+			// Function `redpanda-smp`, that should overwrite `overprovisioned` flag,
+			// was not called before setting `overprovisioned` flag.
+			// redpanda-smp template - https://github.com/redpanda-data/helm-charts/blob/5f287d45a3bda2763896840e505fb3de82b968b6/charts/redpanda/templates/_helpers.tpl#L187
+			// redpanda-smp template invocation - https://github.com/redpanda-data/helm-charts/blob/5f287d45a3bda2763896840e505fb3de82b968b6/charts/redpanda/templates/_configmap.tpl#L610
+			// overprovisioned flag - https://github.com/redpanda-data/helm-charts/blob/5f287d45a3bda2763896840e505fb3de82b968b6/charts/redpanda/templates/_configmap.tpl#L607
+			rex := regexp.MustCompile("\"overprovisioned\":(true|false)")
+			newConf.redpanda = rex.ReplaceAllString(newConf.redpanda, "\"overprovisioned\":false")
 
 			require.JSONEq(t, oldConf.redpanda, newConf.redpanda)
 			require.JSONEq(t, oldConf.bootstrap, newConf.bootstrap)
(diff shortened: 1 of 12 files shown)


charts/redpanda/chart_test.go
```diff
@@ -238,6 +239,9 @@ func extractRedpandaConfigsFromConfigMap(manifests []byte) (*configmapRepresenta
 				}
 				result.redpanda = string(jsonR)
 
+				rex := regexp.MustCompile("\"overprovisioned\":(true|false)")
```
chrisseto, line 242: nit: This is kinda hidden right now and using regex to change up JSON always feels a bit sketchy. What about moving this up to line 205 and making it something like:

```go
// If this path is deeper, use the `unstructured` package from k8s to access it
// Explain the old bug here. Now that it's considered fixed, we'll ignore this diff by pushing the correct value into the old conf
oldConf.redpanda["overprovisioned"] = newConf.redpanda["overprovisioned"]
```

Placing this right before the assertion will help group up all the logic that we start to accumulate like this.

You could also compute a JSON diff and remove the `overprovisioned` field from it in a similar way.

charts/redpanda/values.go
```diff
@@ -345,18 +345,10 @@ type RedpandaResources struct {
...
-	//if rr.CPU.Overprovisioned == nil {
-	//	return false
-	//}
-	//
-	//if int(rr.RedpandaCoresInMillis()) < 1000 {
-	//	return true
-	//}
-	//
-	//return *rr.CPU.Overprovisioned
+	if int(rr.RedpandaCoresInMillis()) < 1000 {
```
chrisseto, line 348: There's a really interesting issue from @StephanDollberg on what the correct way to set this is. Curious if he (currently) has any suggestions on what we should be doing here.

chrisseto (commented):
Could you explain the original issue a bit more? Either in the commit message or as a comment somewhere. I don't think I fully understand why this is a fix?

charts/redpanda/values.go
StephanDollberg: I am not sure I understand the full PR but this change only seems to be concerned with setups where we have less than one 1 core/1000 millicpus assigned to Redpanda?

In that case using overprovisioned unconditionally is the right thing to do.

charts/redpanda/chart_test.go
RafalKorepta: > What about moving this up to line 205
>
> If this path is deeper, use the `unstructured` package from k8s to access it

I can not use on manifests as `---` (YAML delimiter) is unrecognized by the unstructured unmarshall function.

> Placing this right before the assertion will help group up all the logic that we start to accumulate like this.

I agree.

> You could also compute a JSON diff and remove the overprovisioned field from it in a similar way.

Which JSON diff function/library?

charts/redpanda/values.go
chrisseto: That's exactly what it is, I just didn't understand at the time 😓 

Thank you for confirming! I was going to seek further guidance on the `--overprovisioned` flag but we'll do that at a later time.

charts/redpanda/chart_test.go
chrisseto: > I can not use on manifests as --- (YAML delimiter) is unrecognized by the unstructured unmarshall function.

Wouldn't that be a non issue around line 205 as you won't be working with manifests (multi-document YAMLs) but with the parsed out configmap values?

RafalKorepta: `newConf.redpanda` holds a raw redpanda config not K8S ConfigMap. When I'm unmarshall raw redpanda config I get the error:
```
Received unexpected error:
Object 'Kind' is missing in ...
```

chrisseto (approved):

PR #1359, opened 2024-06-17
Handle secret reference check as method

RafalKorepta:
Based on https://github.com/redpanda-data/helm-charts/pull/1268

#### 529cffb0439e889db532719849fe8eda5382ae9c Handle access and secret key reference as methods

### DIFF
--- a/charts/redpanda/values.go
+++ b/charts/redpanda/values.go
@@ -1363,6 +1363,12 @@ type SecretRef struct {
 	Name             string `json:"name"`
 }
 
+// IsValid confirms whether EnvVarSource could be built from
+// SecretRef.
+func (sr *SecretRef) IsValid() bool {
+	return sr != nil && !helmette.Empty(sr.Key) && !helmette.Empty(sr.Name)
+}
+
 type TieredStorageCredentials struct {
 	ConfigurationKey string     `json:"configurationKey" jsonschema:"deprecated"`
 	Key              string     `json:"key" jsonschema:"deprecated"`
--- a/charts/redpanda/post_install_upgrade_job.tpl.go
+++ b/charts/redpanda/post_install_upgrade_job.tpl.go
@@ -103,11 +103,7 @@ func addCloudStorageAccessKey(tieredStorageConfig TieredStorageConfig, values Va
 				Value: v.(string),
 			},
 		}
-		// TODO change this to the following representation when struct function transpilation would work
-		//} else if values.Storage.Tiered.CredentialsSecretRef.IsAccessKeyReferenceValid() {
-	} else if ak := values.Storage.Tiered.CredentialsSecretRef.AccessKey; ak != nil &&
-		!helmette.Empty(ak.Name) &&
-		!helmette.Empty(ak.Key) {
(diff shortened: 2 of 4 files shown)


charts/redpanda/values.go
```diff
@@ -1363,6 +1363,10 @@ type SecretRef struct {
 	Name             string `json:"name"`
 }
 
+func (sr *SecretRef) IsReferenceValid() bool {
+	return sr != nil && !helmette.Empty(sr.Key) && !helmette.Empty(sr.Name)
```
chrisseto, line 1367: nit: prefer direct string comparisons and/or using `ptr.Deref` over `helmette.Empty`. The former makes the typing of `.Key` and `.Name` much more clear and confident in itself.

charts/redpanda/values.go
```diff
@@ -1363,6 +1363,10 @@ type SecretRef struct {
 	Name             string `json:"name"`
 }
 
+func (sr *SecretRef) IsReferenceValid() bool {
```
chrisseto, line 1366: nit: Because the type already has `Ref` in the name, `IsReferenceValid` has a bit of a "stutter" to it. `SecretRef.IsValid` is just as clear. 

Could you also add a quick comment denoting what `Is[Reference]Valid` means?

chrisseto (approved):

charts/redpanda/values.go
RafalKorepta: I found that transpiled code would wrongly handled cases where secret ref is not provided and key is not empty string, but rather nil. 

`testdata/ci/23-aks-tiered-storage-with-creds-values.yaml.tpl.golden` had the following change

```diff
diff --git a/charts/redpanda/testdata/ci/23-aks-tiered-storage-with-creds-values.yaml.tpl.golden b/charts/redpanda/testdata/ci/23-aks-tiered-storage-with-creds-values.yaml.tpl.golden
index b2244521..02480a60 100644
--- a/charts/redpanda/testdata/ci/23-aks-tiered-storage-with-creds-values.yaml.tpl.golden
+++ b/charts/redpanda/testdata/ci/23-aks-tiered-storage-with-creds-values.yaml.tpl.golden
@@ -1393,6 +1393,11 @@ spec:
           value: ${REDPANDA_SAMPLE_LICENSE}
         - name: RPK_CLOUD_STORAGE_AZURE_SHARED_KEY
           value: ${TEST_AZURE_SHARED_KEY}
+        - name: RPK_CLOUD_STORAGE_ACCESS_KEY
+          valueFrom:
+            secretKeyRef:
+              key: null
+              name: null
         - name: RPK_CLOUD_STORAGE_AZURE_CONTAINER
           value: ${TEST_STORAGE_CONTAINER}
         - name: RPK_CLOUD_STORAGE_AZURE_STORAGE_ACCOUNT
```

I can not compare string to nil as our declaration does not declares pointer to string.

RafalKorepta: The `SecretRef` is not nil 😮‍💨 as the `ConfigurationKey` is set to `cloud_storage_access_key`.

charts/redpanda/values.go
RafalKorepta: ok

PR #1364, opened 2024-06-17
gotohelm: add partial support for `resource.Quantity`

chrisseto:
This commit adds partial support for Kubernetes' `resource.Quantity` type into gotohelm. The supported methods are:
- `resource.MustParse`
- `resource.(*Quantity).AsInt64`
- `resource.(*Quantity).Value`
- `resource.(*Quantity).DeepCopy`

It is also possible to include a `resource.Quantity` type on a Values struct and interact with it as you would any other `resource.Quantity` provided that the value is a numeric or a valid `resource.Quantity` string.

### DIFF
--- a/pkg/gotohelm/internal/bootstrap/bootstrap.go
+++ b/pkg/gotohelm/internal/bootstrap/bootstrap.go
@@ -22,6 +22,22 @@ import (
 	"fmt"
 )
 
+const (
+	// For reference: https://physics.nist.gov/cuu/Units/binary.html
+	milli = 0.001
+	kilo  = 1000
+	mega  = kilo * kilo
+	giga  = kilo * kilo * kilo
+	terra = kilo * kilo * kilo * kilo
+	peta  = kilo * kilo * kilo * kilo * kilo
+
+	kibi = 1024
+	mebi = kibi * kibi
+	gibi = kibi * kibi * kibi
+	tebi = kibi * kibi * kibi * kibi
+	pebi = kibi * kibi * kibi * kibi * kibi
+)
+
 // typeatest is the implementation of the go syntax `_, _ := m.(t)`.
 func typetest(typ string, value, zero any) []any {
 	if TypeIs(typ, value) {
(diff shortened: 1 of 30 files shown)


chrisseto:
This commit is a bit old at this point. I need to revisit it to make sure that everything lines up as I recall considering dropping support for `AsInt64` as `Value` seems to serve it's purpose but better. I don't foresee anything major changing though.

pkg/gotohelm/internal/bootstrap/bootstrap.go
```diff
@@ -129,3 +146,125 @@ func asintegral(value any) (any, bool) {
...
 	return 0, false
 }
+
+func parseResource(repr any) (float64, float64) {
+	if TypeIs("float64", repr) {
+		return Float64(repr), 1
+	}
+
+	if !TypeIs("string", repr) {
+		panic(fmt.Sprintf("invalid Quantity expected string or float64 got: %T", repr))
```
RafalKorepta, line 156: NIT:
```suggestion
		panic(fmt.Sprintf("invalid Quantity expected string or float64 got: %T (%v)", repr, repr))
```

RafalKorepta (approved):
LGTM

pkg/gotohelm/transpiler.go
```diff
@@ -878,7 +878,25 @@ func (t *Transpiler) transpileCallExpr(n *ast.CallExpr) Node {
...
 
 	id := callee.Pkg().Path() + "." + callee.Name()
+
+	// TODO explain/figure out the difference here. It matters a lot.
 	signature := t.typeOf(n.Fun).(*types.Signature)
+	// signature := callee.Type().(*types.Signature)
+
+	var reciever Node
+	if recv := callee.Type().(*types.Signature).Recv(); recv != nil {
+		reciever = t.transpileExpr(n.Fun.(*ast.SelectorExpr).X)
```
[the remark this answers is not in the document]
chrisseto: There's one edge case that I don't think we'll ever encounter. I've added a note!

pkg/gotohelm/internal/bootstrap/bootstrap.go
chrisseto: Done!

RafalKorepta (approved):

PR #1366, opened 2024-06-18
gotohelm: support go's append semantics

chrisseto:
Prior to this commit our gotohelmified go code was littered with slice declarations in the form `_ := T[]{}` due to sprig's `mustAppend` function not supporting appends to `nil` slices as go's does.

This commit corrects the semantics of transpiling `append` to now match go's semantics by use of `default` and `concat` instead of `mustAppend`. It also updates a few occurences of `_ := T[]{}` to the more idiomatic `var _ T[]`


RafalKorepta (approved):
LGTM

PR #1368, opened 2024-06-20
buildkite: retrieve secrets from aws sm

andrewhsu:
fixes https://redpandadata.atlassian.net/browse/PESDLC-1527

buildkite currently reads secrets from the s3 secrets bucket into env vars for use by the pipeline. these secrets were copied into aws secretsmanager and this PR updates the pipeline so it uses the [seek-oss/aws-sm](https://github.com/seek-oss/aws-sm-buildkite-plugin) plugin to read the secrets from aws secretsmanager for the https://buildkite.com/redpanda/helm-charts pipeline. there is no functionality change because the env vars are kept the same name and values.

### DIFF
--- a/.buildkite/pipeline.yaml
+++ b/.buildkite/pipeline.yaml
@@ -1,3 +1,21 @@
+---
+plugins:
+  - seek-oss/aws-sm#v2.3.2: &aws-sm-plugin
+      json-to-env:
+        - json-key: .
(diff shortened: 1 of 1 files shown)


RafalKorepta (approved):
Does the K8S team have access to AWS Secret Manager to self manage new secrets that might be added in future?

andrewhsu:
@RafalKorepta fyi i've posted you a link in jira

PR #1370, opened 2024-06-20
redpanda: convert secrets.yaml to go

jan-g:
This includes a fix to a long-standing bug: the curl arguments for an admin listener that doesn't require a client cert named the wrong file.

This replicates blank lines in generated shell scripts with high fidelity to the helm originals, but that can be relaxed once the conversion is complete.


charts/redpanda/secrets.go
```diff
@@ -0,0 +1,731 @@
...
+	"fmt"
+	"github.com/redpanda-data/helm-charts/pkg/gotohelm/helmette"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
+	"strings"
+)
+
+func Secrets(dot *helmette.Dot) []*corev1.Secret {
+	secrets := []*corev1.Secret{}
```
chrisseto, line 29: nit: This is now supported
```suggestion
	var secrets []*corev1.Secret
```

charts/redpanda/secrets.go
```diff
@@ -0,0 +1,731 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("%s-config-watcher", Fullname(dot)),
+			Namespace: dot.Release.Namespace,
+			Labels:    FullLabels(dot),
+		},
+		Type:       corev1.SecretTypeOpaque,
+		StringData: map[string]string{},
+	}
+
+	saslUserSh := []string{}
```
chrisseto, line 228: same nit:
```suggestion
	var saslUserSh []string
```

charts/redpanda/secrets.go
```diff
@@ -0,0 +1,731 @@
...
+			`# Configure bootstrap`,
+			`## Not used for Redpanda v22.3.0+`,
+			`rpk --config "${CONFIG}" redpanda config set redpanda.node_id "${POD_ORDINAL}"`,
+			`if [ "${POD_ORDINAL}" = "0" ]; then`,
+			`	rpk --config "${CONFIG}" redpanda config set redpanda.seed_servers '[]' --format yaml`,
+			`fi`,
+		)
+	}
+
+	{ // Handle kafka listener
```
chrisseto, line 465: 😬 So the scoping guarantees of blocks do not extend to gotohelm. It might be best to factor these out into different functions just to be safe?

I see what you mean about the pain of dealing with all the slightly differently typed listeners now.

charts/redpanda/secrets.go
```diff
@@ -0,0 +1,731 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// +gotohelm:filename=secrets.go.tpl
+package redpanda
+
+import (
+	"fmt"
```
chrisseto, line 20: I'm guessing that gofumpt is what's failing CI as this needs a newline after it.

chrisseto (approved):
LGTM and I'm blown away that you managed to get the whitespace to line up perfectly.

To be on the safe side, I think it would be good to pull some of the `{}` blocks into functions as I wouldn't want a latent divergence to bite us in certain cases as the test coverage isn't incredible right now.

charts/redpanda/secrets.go
jan-g: This manages to happen every time. Need to examine my ide settings - it should be fixing this for me and instead it's reliably "fixing" it :-/

charts/redpanda/secrets.go
jan-g: Yeah, I wanted to get this working first - the blocks are there as a visual guide as much as anything else.

The bulk of the noise in here comes from the support for templating in advertised names, which I think we need to keep - there's a contract there that we need to confirm is working.

jan-g: Given that these are functions rather than methods, I think we might be able to refactor this to write that function once. I'll put up something to address the nits then look at that.

RafalKorepta (approved):

charts/redpanda/secrets.go
```diff
@@ -0,0 +1,759 @@
...
+	return snippet
+}
+
+// The following from _helpers.tpm
+
+func adminInternalTLSEnabled(dot *helmette.Dot) bool {
+	values := helmette.Unwrap[Values](dot.Values)
+
+	listener := values.Listeners.Admin
+	enabled := ptr.Deref(listener.TLS.Enabled, values.TLS.Enabled)
```
RafalKorepta, line 669: NIT: There should be option to 
```suggestion
	enabled := values.Listeners.Admin.TLS.IsEnabled(&values.TLS)
```

RafalKorepta: Sorry wrong place

charts/redpanda/secrets.go
```diff
@@ -0,0 +1,759 @@
...
+	hasCert := listener.TLS.Cert != ""
+	return enabled && hasCert
+}
+
+func adminTLSCurlFlags(dot *helmette.Dot) string {
+	values := helmette.Unwrap[Values](dot.Values)
+
+	if !adminInternalTLSEnabled(dot) {
```
RafalKorepta, line 677:
```suggestion
	if !values.Listeners.Admin.TLS.IsEnabled(&values.TLS) {
```

charts/redpanda/secrets.go
```diff
@@ -0,0 +1,759 @@
...
+				"port":    port,
+			}
+		}
+	}
+	return host
+}
+
+// adminInternalHTTPProtocol was admin-http-protocol
+func adminInternalHTTPProtocol(dot *helmette.Dot) string {
+	if adminInternalTLSEnabled(dot) {
```
RafalKorepta, line 729:
```suggestion
	if values.Listeners.Admin.TLS.IsEnabled(&values.TLS) {
```

charts/redpanda/secrets.go
jan-g: I'll get these now.

PR #1373, opened 2024-06-21
redpanda: allow specifying `truststore_file`

### LINKED ISSUE #1339: 🫐 🐛 We are unable to correctly implement mTLS because it doesn't seem we can pass our own truststore files to Redpanda listeners
### What happened?

The Helm chart nor the operator CR seem to give us a way to configure listeners' truststores. The current configurations seem to only support 1 CA cert, which is configured to that of the certificate and key configured in the listener. For Redpanda Cloud, we need to be able to provide one or more public CA certs that Redpanda must trust and use to authenticate client connections. 

I was expecting to be able to mount our own truststore files, containing one or more chained public CA certs. 

Try to configure mTLS with two or more CAs public certs. 

_No response_

Redpanda, Operator

### Chart Version(s)

<details>

```console
$ helm -n <redpanda-release-namespace> list 
# paste output here
```

</details>

### Cloud provider

<details>
Azure, GCP, AWS
</details>

JIRA Link: [K8S-253](https://redpandadata.atlassian.net/browse/K8S-253)

[K8S-253]: https://redpandadata.atlassian.net/browse/K8S-253?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

chrisseto:
Prior to this commit the `truststore_file` of a given listener could only be set by using the confusingly named `caEnabled` field on certificates. Furthermore `caEnabled` was limited to a the `ca.crt` key of the certificate's secret and was set on a per certificate basis.

This commit allows explicit control of `truststore_file` on a per listener basis by adding a `truststore` field to listeners' `tls` field. Similar to Kubernetes' `corev1.EnvVarSource`, `truststore` accepts either a `configMapKeyRef` or a `secretKeyRef`.

Fixes #1339

### DIFF
--- a/charts/redpanda/chart_template_test.go
+++ b/charts/redpanda/chart_template_test.go
@@ -10,6 +10,7 @@ import (
 	"testing"
 
 	certmanagerv1 "github.com/cert-manager/cert-manager/pkg/apis/certmanager/v1"
+	jsoniter "github.com/json-iterator/go"
 	"github.com/redpanda-data/helm-charts/charts/redpanda"
 	"github.com/redpanda-data/helm-charts/pkg/helm"
 	"github.com/redpanda-data/helm-charts/pkg/kube"
@@ -41,6 +42,7 @@ func TestTemplate(t *testing.T) {
 	cases := CIGoldenTestCases(t)
 	cases = append(cases, VersionGoldenTestsCases(t)...)
 	cases = append(cases, DisableCertmanagerIntegration(t)...)
+	cases = append(cases, CertTrustStoreCases(t)...)
 
 	for _, tc := range cases {
 		tc := tc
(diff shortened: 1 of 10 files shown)


charts/redpanda/statefulset.go
```diff
@@ -212,13 +216,21 @@ func StatefulSetVolumes(dot *helmette.Dot) []corev1.Volume {
 		},
 	}...)
 
+	if len(values.Listeners.TrustStores(&values.TLS)) > 0 {
+		volumes = append(
+			volumes,
+			values.Listeners.TrustStoreVolume(&values.TLS),
```
RafalKorepta, line 222: Would that change work?
```suggestion
	if vols := values.Listeners.TrustStores(&values.TLS); len(vols) > 0 {
		volumes = append(
			volumes,
			vols,
```

charts/redpanda/chart_template_test.go
```diff
@@ -294,6 +296,171 @@ tls:
 	}
 }
 
+func CertTrustStoreCases(t *testing.T) []TemplateTestCase {
+	// truststores is a map of listener type to map of listener name to truststore_file. ({"admin": {"internal": "ca.crt"}}).
+	assertTrustStores := func(t *testing.T, manifests []byte, truststores map[string]map[string]string) {
+		cm, err := extractRedpandaConfigsFromConfigMap(manifests)
```
RafalKorepta, line 302: `extractRedpandaConfigsFromConfigMap` is missing from this PR.

charts/redpanda/statefulset.go
chrisseto: There's only a single volume that's returned so I couldn't do it that way (and it's two different method calls). I've change it so that the truststorevolume returns a pointer that's nil when no truststores are configured.

charts/redpanda/chart_template_test.go
chrisseto: Fixed!

RafalKorepta (approved):
LGTM. Do you want to bump Chart.yaml to release this improvement?

chrisseto:
I'll cut a release in a different PR!

PR #1374, opened 2024-06-24
redpanda: convert post-install-upgrade-job.yaml to go

jan-g:
This includes a new `helmette.MergeTo`, as follows:

We are largely dealing with types from three domains: the input values, the desired k8s structures (both of these have relatively sophisticated typing; although these types might in some cases be roughly isomorphic but not identical) and the raw-json-structs of Helm's templates.

Because we're coming from a templated world, the use of `merge` is common to fold overrides onto defaults.

We provide a type-aware `MergeTo`, parameterised by the target type. This permits go to be written in a type-safe fashion, but still permits the kind of structure-merging common in code that begins life as the translation of a Helm template. Under the hood, it has to round-trip into `mergo`-compatible raw values, but we do our best to hide this from the caller.


charts/redpanda/post_install_upgrade_job.go
```diff
@@ -0,0 +1,200 @@
...
+					Containers: []corev1.Container{
+						{
+							Name:      fmt.Sprintf("%s-post-install", Name(dot)),
+							Image:     fmt.Sprintf("%s:%s", values.Image.Repository, Tag(dot)),
+							Env:       PostInstallUpgradeEnvironmentVariables(dot),
+							Command:   []string{"bash", "-c"},
+							Args:      []string{},
+							Resources: postInstallJobResources(values.PostInstallJob.Resources),
+							// XXX the following merge is probably the wrong way around: it doesn't permit the overriding of defaults
+							SecurityContext: ptr.To(helmette.MergeTo[corev1.SecurityContext](ContainerSecurityContext(dot), helmette.Default(corev1.SecurityContext{}, values.PostInstallJob.SecurityContext))),
```
jan-g, line 89: Question for reviewer: this merge is arguably the wrong way around, since the "defaults" will overrider the job-specific values. Given that this field was originally missing from values.go, I question whether this was actually being used in the wild.

I'm leaning towards fixing up the ordering of the MergeTo arguments here, but I'd like a second opinion before I make that change in this PR.

charts/redpanda/values.go
```diff
@@ -484,9 +484,10 @@ type PostInstallJob struct {
 	Affinity  map[string]any `json:"affinity"`
 
 	// Fields that are in values.yaml but not in values.schema.json.
-	// Enabled     bool              `json:"enabled"`
-	// Labels      map[string]string `json:"labels"`
-	// Annotations map[string]string `json:"annotations"`
+	Enabled         bool                   `json:"enabled"`
+	Labels          map[string]string      `json:"labels"`
+	Annotations     map[string]string      `json:"annotations"`
+	SecurityContext corev1.SecurityContext `json:"securityContext"`
```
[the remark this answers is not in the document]
RafalKorepta: > do we have a way to audit the use of these potentially vestigial fields?

No, there is no way to check what values are used in the wild.

> which this PR currently doesn't fix - should it?

In the new commit that would have dedicated values input to test it out it can go in this PR or in new. I'm open to any outcome. 

>  I'm also not 100% convinced that additional security context settings can provide anything that this script could conceivably require in order to operate (outside of pathological scenarios).

I feel like access to `/tmp/` file system and network connectivity should be an requirement. Other security settings could be more restrictive.

charts/redpanda/post_install_upgrade_job.go
```diff
@@ -0,0 +1,200 @@
...
+	affinity.PodAntiAffinity = ptr.To(helmette.Unwrap[corev1.PodAntiAffinity](helmette.Default(map[string]any{}, values.Affinity.PodAntiAffinity)))
+	return &affinity
+}
+
+func tolerations(dot *helmette.Dot) []corev1.Toleration {
+	values := helmette.Unwrap[Values](dot.Values)
+
+	var result []corev1.Toleration
+	for _, t := range values.Tolerations {
+		result = append(result, helmette.Unwrap[corev1.Toleration](helmette.Default(map[string]any{}, t)))
```
RafalKorepta, line 183: This code transpiles to 

```
{{- $result = (concat (default (list ) $result) (list (default (dict ) $t).AsMap)) -}}
```

An my initial reaction would be how to test a case where `values.Tolerations` have nil map included in the list.

charts/redpanda/post_install_upgrade_job.go
```diff
@@ -0,0 +1,200 @@
...
+					SecurityContext:  PodSecurityContext(dot),
+					ImagePullSecrets: pullSecrets(dot),
+					Containers: []corev1.Container{
+						{
+							Name:      fmt.Sprintf("%s-post-install", Name(dot)),
+							Image:     fmt.Sprintf("%s:%s", values.Image.Repository, Tag(dot)),
+							Env:       PostInstallUpgradeEnvironmentVariables(dot),
+							Command:   []string{"bash", "-c"},
+							Args:      []string{},
+							Resources: postInstallJobResources(values.PostInstallJob.Resources),
```
RafalKorepta, line 87: I just wonder why not:

```suggestion
							Resources: helmette.MergeTo[corev1.ResourceRequirements](resources),
```

charts/redpanda/post_install_upgrade_job.go
```diff
@@ -0,0 +1,200 @@
...
+			APIVersion: "batch/v1",
+			Kind:       "Job",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("%s-configuration", Fullname(dot)),
+			Namespace: dot.Release.Namespace,
+			Labels: helmette.Merge(
+				FullLabels(dot),
+				helmette.Default(map[string]string{}, values.PostInstallJob.Labels),
+			),
```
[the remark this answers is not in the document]
jan-g: I'll fix all of these and put up a test case covering them.

charts/redpanda/post_install_upgrade_job.go
jan-g: Yeah, given how trivial that function became once I had MergeTo I'll do this.

RafalKorepta (commented):
I would like to approve this PR once you create new ci test case that would try to change PodSpec.Affinity and PodSpec.Tolerations and postInstallJobResources. As far as I saw the SecurityContext is correctly propagated within `34-security-contexts-novalues.yaml.golden` generated file.

charts/redpanda/post_install_upgrade_job.go
jan-g: What would that mean? I think we should look to move towards a situation where there's a strong schema wrapped around `values` - Tolerations moving from `[]map[string]any` (ie, abrogating responsibility of checking the type to the user) to `[]corev1.Toleration`. Shifting toward the situation where illegal values are unrepresentable seems the right way to go.

charts/redpanda/post_install_upgrade_job.go
jan-g: ... actually there's something amiss here. That refactor leaves the resulting template failing when values.postInstallJob.resources is empty.

charts/redpanda/post_install_upgrade_job.go
RafalKorepta: I agree with you. Could we change the schema
https://github.com/redpanda-data/helm-charts/blob/9410e53301116e3aa5c6b6bc6cd49e51096ca86e/charts/redpanda/values.schema.json#L2487-L2492
to not allow nil object in the array?

charts/redpanda/post_install_upgrade_job.go
```diff
@@ -0,0 +1,200 @@
...
+		},
+		Spec: batchv1.JobSpec{
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					GenerateName: fmt.Sprintf("%s-post-", dot.Release.Name),
+					Labels: helmette.Merge(
+						map[string]string{
+							"app.kubernetes.io/name":      Name(dot),
+							"app.kubernetes.io/instance":  dot.Release.Name,
+							"app.kubernetes.io/component": fmt.Sprintf("%s-post-install", Name(dot)[:50]),
```
chrisseto, line 68: Fun fact, printf has builtin support for truncating `%[50]s-post-install`. Either way works for me.

charts/redpanda/post_install_upgrade_job.go
```diff
@@ -0,0 +1,200 @@
...
+	values := helmette.Unwrap[Values](dot.Values)
+
+	var result []corev1.Toleration
+	for _, t := range values.Tolerations {
+		result = append(result, helmette.Unwrap[corev1.Toleration](helmette.Default(map[string]any{}, t)))
+	}
+	return result
+}
+
+func postInstallJobResources(resources JobResources) corev1.ResourceRequirements {
```
chrisseto, line 188: I have this solved in a commit I'll put up soon. I just added the corev1.ResourceRequirements to the values type.

charts/redpanda/post_install_upgrade_job.go
jan-g: Fixed up in secnod commit.

charts/redpanda/post_install_upgrade_job.go
jan-g: ah-ha, I'll fix that up.

pkg/gotohelm/helmette/sprig.go
```diff
@@ -160,6 +160,30 @@ func Merge[K comparable, V any](sources ...map[K]V) map[K]V {
...
+	for _, src := range sources {
+		// Turn it into something json-like
+		roundTrip := FromJSON(ToJSON(src))
+		if err := mergo.Merge(&dst, roundTrip); err != nil {
+			panic(fmt.Errorf("cannot merge roundtripped element: %w", err))
+		}
+	}
+	// The following could be replaced with Unwrap[T] at the point it handles
+	// things like resource.Quantity correctly
+	var result T
```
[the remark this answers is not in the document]
jan-g: done anyway

charts/redpanda/post_install_upgrade_job.go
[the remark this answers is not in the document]
jan-g: Change of plan: I've currently left this as-is. I think a general task to take an overview of overrides and ensure they're working correctly is better done over the setup as a whole rather than just here.

RafalKorepta (approved):

PR #1354, opened 2024-06-14
genpartial: refactor to type traversal

chrisseto:
This commit refactors `genpartial` to instead traverse types via the go type system instead of traversing the AST. This strategy fixes a handful of bugs that previously went unnoticed and otherwise makes the package a bit more approachable.

Fixes #1129

PR #1376, opened 2024-06-24
redpanda: use `resource.Quantity`

chrisseto:
This commit upgrades all redpanda values types to use the `resource.Quantity` struct where we had previously been using other types due to a lack of compatibility.

It also adds a new `UnmarshalInto` function to the `helmette` package that allows working around go's type system in cases where an `any` needs to be upgraded into a `resource.Quantity`.


chrisseto:
The partial values in this PR is going to be incorrect until https://github.com/redpanda-data/helm-charts/pull/1354 gets (fixed and) merged

RafalKorepta (approved):

PR #1380, opened 2024-06-25
fix up secrets.go

jan-g:
The `tmplVals` code appears to be entirely dead code - the only previous call to `tpl` is computing `externalAdvertiseAddress`, which is performed using `.` = `$`. The template expansion of external hostnames is done using the root context (which is preserved in `.` in this case).

### References

https://github.com/redpanda-data/helm-charts/pull/425
https://github.com/helm/helm/issues/4166#issuecomment-497430905
https://github.com/redpanda-data/helm-charts/pull/410

### DIFF
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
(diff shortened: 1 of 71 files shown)


jan-g:
Marked as ready for review, but ideally we need a test case of the templating of an external name. I think @RafalKorepta might have an example.

RafalKorepta (approved):

PR #1382, opened 2024-06-25
redpanda: migrate cert-issuers.yaml to go

jan-g:


chrisseto (approved):
Looks like there's a some tests that need to be updated but otherwise LGTM!

RafalKorepta (approved):

PR #1386, opened 2024-06-26
gotohelm: fix return from nested range

jan-g:


jan-g:
Already landed with Rafal's associated PR.

PR #1402, opened 2024-06-27
Console fix up with mutual TLS

RafalKorepta:
This PR is based on https://github.com/redpanda-data/helm-charts/pull/1386 to solve early return problem in go template.

Client certificates does not relay on secret reference, that's why checks are removed.

Add integration test that setup mTLS for all listeners (Kafka, Admin API, HTTP Proxy, Schema Registry, RPC). 

Verification in rpk client is not always done using `rpk`. `curl` should be replaced with port-forward and simple go based http client Get/Post/Delete implementation.

Fix `ClientAuthRequired` check as `Dig` function should start from `listeners` object.

Expose, in Redpanda values, Console config and chunk of Connectors values. 

Mount all client certificates (ca, certificate and key) in Console deployment.

Move Admin API configuration from env vars to ConfigMap. 

#### Reference

https://docs.redpanda.com/current/develop/http-proxy/

https://docs.redpanda.com/current/manage/schema-reg/schema-reg-api/


charts/redpanda/certs.go
```diff
@@ -89,7 +89,7 @@ func ClientCerts(dot *helmette.Dot) []certmanagerv1.Certificate {
 		panic(fmt.Sprintf("Certificate %q referenced but not defined", name))
 	}
 
-	if helmette.Empty(data.SecretRef) || !ClientAuthRequired(dot) {
+	if !ClientAuthRequired(dot) {
```
chrisseto, line 92: I was a bit worried about this. We'll need to ask Azure (cc @c4milo or @ncole ) how they'd like to handle this. 
For mTLS to work, the chart needs to be able to get a client certificate for either console or RPK (@RafalKorepta which is it?). This currently works through the cert manager integration but there's not really a fallback. If y'all can mint a cert we can find a field to add a secretRef to or we can talk through relying on the cert-manger integration either partially or fully.

charts/redpanda/client_test.go
```diff
@@ -45,3 +46,494 @@ func (c *Client) ClusterConfig(ctx context.Context) (redpanda.ClusterConfig, err
...
+
+	var event map[string]any
+	if err = json.Unmarshal(out.Bytes(), &event); err != nil {
+		return nil, err
+	}
+
+	return event, nil
+}
+
+func (c *Client) GetClusterHealth(ctx context.Context, dot *helmette.Dot) (map[string]any, error) {
```
chrisseto, line 121: I don't love the idea of using `*helmette.Dot` here. If there's a bug in some of the Values functions that would surface here and potentially obscure issues from our tests, specifically in the case of TLS configurations.

I think it would be a more robust testing strategy to instead have the Client accept the TLS certificates (extracted from the cluster if using the cert-manger integration) and then use those to actually dial into the various listeners of the Pod, piggy backing on `kubectl port-forward` to actually make the connection. Then we can rely on client libraries instead of having to template out CLI commands.

For an initial implementation thought, let's roll with it!

charts/redpanda/console.tpl.go
```diff
@@ -0,0 +1,119 @@
...
+				"urls": []string{
+					fmt.Sprintf("%s://%s:%d", schema, InternalDomain(dot), values.Listeners.Admin.Port),
+				},
+				"tls": values.Listeners.Admin.ConsoleTLS(&values.TLS),
+			},
+		},
+	}
+
+	if values.Connectors.Enabled {
+		// TODO Do not cal Dig with dot.Values as restPort that is defined in connectors helm chart is not
```
chrisseto, line 70: I can think of ~2 ways we could handle this.
1) Have a helmette function that runs a "raw" template command (Though it'll be quite difficult to get that to work reliably in go but perhaps that's okay. `helmette.Raw("(include connectors.serviceName $var)")`
2) Teach gotohelm how to handle subcharts by resolving the name to the appropriate call but not actually transpile the underlying call.

Given that we'll need to convert the connectors and console chart, I'm leaning towards 2 but 1 would certainly be faster.

charts/redpanda/console.tpl.go
```diff
@@ -0,0 +1,119 @@
...
+
+		port := helmette.Dig(dot.Values.AsMap(), 8083, "connectors", "connectors", "restPort")
+		p, ok := helmette.AsIntegral[int](port)
+		if !ok {
+			return c
+		}
+
+		connectorsURL := fmt.Sprintf("http://%s.%s.svc.%s:%d",
+			ConnectorsFullName(dot),
+			dot.Release.Namespace,
```
chrisseto, line 83: hmmm, for sub charting is there a separate release generated or is technically the same?

charts/redpanda/helpers.go
```diff
@@ -175,7 +175,7 @@ func TLSEnabled(dot *helmette.Dot) bool {
 func ClientAuthRequired(dot *helmette.Dot) bool {
 	listeners := []string{"kafka", "admin", "schemaRegistry", "rpc", "http"}
 	for _, listener := range listeners {
-		required := helmette.Dig(dot.Values.AsMap(), false, listener, "tls", "requireClientAuth")
+		required := helmette.Dig(dot.Values.AsMap(), false, "listeners", listener, "tls", "requireClientAuth")
```
chrisseto, line 178: Oh dear this has been wrong for a while. I wonder how this didn't show up sooner? I'm guessing there's not too many tests that actually use this function.....

chrisseto (commented):
Not a full review will follow up on the test. Seems like we'll want to solve the subcharting problem soon.

charts/redpanda/certs.go
ncole: I'm not sure the context on this, but I think you're saying for mTLS we need a client cert.  We are already creating a self-signed CA *AND* a client cert, so we can provide that ref somehow for Console (or RPK, whichever it is @RafalKorepta)?

In fact, in the other providers, if mTLS is enabled, we add a self-signed CA and a cert for kminion to use so we have an existing precedent for this.

RafalKorepta: > For mTLS to work, the chart needs to be able to get a client certificate for either console or RPK (@RafalKorepta which is it?).

In this particular case, client certificate is used for both RPK and console. They will reference and mount the same certificate. 

> This currently works through the cert manager integration but there's not really a fallback. If y'all can mint a cert we can find a field to add a secretRef to or we can talk through relying on the cert-manger integration either partially or fully.

I agree, that this check needs to accommodate a use case when certificate is provided. I will change that check.

charts/redpanda/client_test.go
RafalKorepta: I will add certificate to the Client.

charts/redpanda/console.tpl.go
RafalKorepta: As far as I understand it's the same.

charts/redpanda/helpers.go
RafalKorepta: I don't know if any test is asserting that Client Auth is configured.

charts/redpanda/certs.go
chrisseto: > I agree, that this check needs to accommodate a use case when certificate is provided. I will change that check.

Let's handle that in a follow up PR/commit?

charts/redpanda/testdata/ci/01-default-values.yaml.golden
```diff
@@ -654,23 +661,14 @@ spec:
         - name: kafka-default-cert
           secret:
             defaultMode: 272
-            items:
```
chrisseto, line 657: huh. Do you know where these diffs are stemming from? It seems incorrect for these to be removed. Could they just be out of date?

charts/redpanda/testdata/ci/01-default-values.yaml.golden
```diff
@@ -490,13 +489,21 @@ data:
...
           enabled: true
+          insecureSkipTlsVerify: false
+          keyFilepath: ""
         urls:
         - https://redpanda-0.redpanda.default.svc.cluster.local.:8081
         - https://redpanda-1.redpanda.default.svc.cluster.local.:8081
         - https://redpanda-2.redpanda.default.svc.cluster.local.:8081
       tls:
+        caFilepath: /mnt/cert/kafka/default/ca.crt
+        certFilepath: ""
```
chrisseto, line 503: Are empty strings valid values of `certFilepath` and `keyFilepath`?

charts/redpanda/testdata/ci/01-default-values.yaml.golden
RafalKorepta: I didn't want transiple the schema, admin and kafka volume definition in the following console deployement:
* https://github.com/redpanda-data/helm-charts/pull/1402/files#diff-0b45ecf3a48bcb79531c1cf2372a743fa93c58bc1ee5193a798e9decd52bf7e4L245
* https://github.com/redpanda-data/helm-charts/pull/1402/files#diff-0b45ecf3a48bcb79531c1cf2372a743fa93c58bc1ee5193a798e9decd52bf7e4L213
* https://github.com/redpanda-data/helm-charts/pull/1402/files#diff-0b45ecf3a48bcb79531c1cf2372a743fa93c58bc1ee5193a798e9decd52bf7e4L185

Regardless of the certification secret reference all will be mounted to console pod.

RafalKorepta:
I added new CI execution that using kind cluster is performing new mTLS test cases
https://github.com/redpanda-data/helm-charts/actions/runs/9864729172/job/27240338010?pr=1402#step:8:1111
https://github.com/redpanda-data/helm-charts/actions/runs/9864729172/job/27240338010?pr=1402#step:8:1116

P.S. I don't understand why `REDPANDA_LICENSE` is not found, so that `tiered-storage-secrets` is skipped. 
https://github.com/redpanda-data/helm-charts/actions/runs/9864729172/job/27240338010?pr=1402#step:8:1039

pkg/kube/ctl.go
```diff
@@ -154,3 +158,80 @@ func (c *Ctl) Exec(ctx context.Context, pod *corev1.Pod, opts ExecOptions) error
...
+	if err != nil {
+		return nil, nil, errors.WithStack(err)
+	}
+
+	stopChan := make(chan struct{})
+	readyChan := make(chan struct{})
+
+	dialer := spdy.NewDialer(upgrader, &http.Client{Transport: transport}, "POST", req.URL())
+
+	var ports []string
```
chrisseto, line 194: I'm not a huge fan of relying on the "portforwarding" abstraction because it binds to local ports for no reason which adds a bunch of extra goroutines and indirection but it's much easier to get started.

We can rip the guts out of https://github.com/kubernetes/client-go/blob/592d891671b2a09e5f81781b28ebe078d8115e41/tools/portforward/portforward.go#L335 to wrap this up into a Dialer interface that'll place nice with go's networking constructs which is a lot easier to utilize with other client libraries (http.RoundTripper for example).

That's a bigger change for later. Just leaving so notes for ourselves.

pkg/helm/helmtest/helmtest.go
```diff
@@ -73,15 +75,26 @@ func (e *NamespacedEnv) Ctl() *kube.Ctl {
...
-func (e *NamespacedEnv) Install(chart string, opts helm.InstallOptions) helm.Release {
-	require.Zero(e.t, opts.Name, ".Name may not be specified")
-	require.Zero(e.t, opts.Namespace, ".Namespace may not be specified")
+func (e *NamespacedEnv) Install(ctx context.Context, chart string, opts helm.InstallOptions) helm.Release {
 	require.False(e.t, opts.CreateNamespace, ".CreateNamespace may not be specified")
 
 	opts.Namespace = e.namespace.Name
 
-	release, err := e.env.helm.Install(context.Background(), chart, opts)
+	// Every time, Redpanda chart is installed, the next deployment would fail
```
chrisseto, line 83: I would argue that this mutex should be used in the actual tests (Or we should disable the node port service in our tests). This struct is completely safe for concurrent use, so let's not limit it's abilities because a user thereof is not.

pkg/helm/helm.go
```diff
@@ -257,6 +257,16 @@ func (c *Client) Install(ctx context.Context, chart string, opts InstallOptions)
...
 }
 
+func (c *Client) Uninstall(ctx context.Context, release Release) error {
+	args := []string{"uninstall", release.Name, fmt.Sprintf("--namespace=%s", release.Namespace), "--wait", "--cascade=foreground"}
+
+	_, _, err := c.runHelm(ctx, args...)
+	if err != nil {
+		return err
+	}
+	return err
```
chrisseto, line 267: nit: `return nil` to be explicit. I'd expect staticlint to yell at you for this.

charts/redpanda/testdata/ci/32-statefulset-podspec-novalues.yaml.golden
```diff
@@ -654,23 +672,14 @@ spec:
         - name: kafka-default-cert
           secret:
             defaultMode: 272
-            items:
```
chrisseto, line 657: Still not sure where this change is coming from and it doesn't seem like it's correct/desirable.

chrisseto (commented):
This is quite a large PR again 😅 Let's try to get the for loop fix merged externally so we can cut down on the diff a bit.

I'm seeing some concerning looking diffs in the golden files. Do you think they're expected?

charts/redpanda/chart_test.go
```diff
@@ -13,11 +26,227 @@ import (
...
+	"go.uber.org/zap/zapcore"
+	"go.uber.org/zap/zapio"
+	"go.uber.org/zap/zaptest"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
 )
 
+func kafkaListenerTest(t *testing.T, ctx context.Context, rpk Client) {
```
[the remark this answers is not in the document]
RafalKorepta: I agree, but I follow tiered storage function which have the same order (first declaration then usage).

https://github.com/redpanda-data/helm-charts/blob/7648b58b9e02e3feebc4b7c797724c3728fd738b/charts/redpanda/chart_test.go#L21-L94

I will move all of them below `TestChart` function

charts/redpanda/chart_test.go
```diff
@@ -138,6 +373,155 @@ func TestChart(t *testing.T) {
...
+			schemaRegistryListenerTest(t, ctx, rpk)
+		})
+
+		t.Run("http-proxy-listener", func(t *testing.T) {
+			httpProxyListenerTest(t, ctx, rpk)
+		})
+	})
+}
+
+func convertPartialToValues(partial *redpanda.PartialValues) (*helmette.Values, error) {
```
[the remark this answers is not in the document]
RafalKorepta: Ok, I will use it then.

charts/redpanda/chart_test.go
```diff
@@ -228,3 +612,142 @@ func TestLabels(t *testing.T) {
 		}
 	}
 }
+
+func createCertificates(chartReleaseName, chartReleaseNamespace string) (ca, serverPublic, serverPrivate, clientPublic, clientPrivate []byte, err error) {
```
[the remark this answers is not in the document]
RafalKorepta: Agree, I will move it to `tlsgenerator` package.

pkg/helm/helmtest/helmtest.go
RafalKorepta: Ok, I will remove mutex and external connectivity properties in values (node ports).

pkg/helm/helm.go
RafalKorepta: Ok, will do that.

charts/redpanda/testdata/ci/32-statefulset-podspec-novalues.yaml.golden
RafalKorepta: It is desirable as per https://github.com/redpanda-data/helm-charts/pull/1402#discussion_r1661911665

I didn't want to list all items explicitly, but rather mount secret and relay on the (implicit) fact that TLS secrets generated by cert-manager and our test suite will have always ca.crt, tls.crt and tls.key.

charts/redpanda/chart_test.go
```diff
@@ -138,6 +373,155 @@ func TestChart(t *testing.T) {
...
+			},
+		}
+
+		cleanup, err := rpk.ExposeRedpandaCluster(ctx, dot, w, wErr)
+		if cleanup != nil {
+			t.Cleanup(cleanup)
+		}
+		require.NoError(t, err)
+
+		t.Run("kafka-listener", func(t *testing.T) {
```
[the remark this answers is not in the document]
chrisseto: aaaaaaah, that's a good point! Other options won't include the diffs if there are failures.

charts/redpanda/testdata/ci/32-statefulset-podspec-novalues.yaml.golden
chrisseto: SGTM! I missed and misunderstood your first comment.

chrisseto (approved):

PR #1404, opened 2024-07-02
Adding readiness probe timeout to the Redpanda stateful set

pfcurtis:


chrisseto (approved):

RafalKorepta:
Just for information that because of Chart.yaml didn't have version bump it is not available atm.

pfcurtis:
It can wait for the next version bump. I have a private chart in use at the moment.

PR #1407, opened 2024-07-03
Fix any Node configuration that is of numerical type

RafalKorepta:
For example crash_loop_limit can change value type from integer to string while template engine converts it due to `toYaml` function usage.

### Reference

https://docs.redpanda.com/current/reference/properties/broker-properties/#crash_loop_limit

### DIFF
--- a/charts/redpanda/chart_test.go
+++ b/charts/redpanda/chart_test.go
@@ -3,7 +3,6 @@ package redpanda_test
 import (
 	"maps"
 	"os"
-	"path/filepath"
 	"testing"
 
 	"github.com/redpanda-data/helm-charts/charts/redpanda"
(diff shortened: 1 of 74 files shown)


chrisseto (approved):
LGTM

How did this make it passed chart-testing in CI?

RafalKorepta:
> How did this make it passed chart-testing in CI?

Maybe Redpanda is not strict about it. Redpanda apparently could work with string conversion to integer (I verified that manually).

I found out by yaml.Unmarshall using RPK go struct that represent Redpanda Node Configuration.

chrisseto:
oooooooh, fascinating. I remember stumbling across the C++ YAML parsing and noting how different it was. Seems like it's more permissive.

PR #1412, opened 2024-07-09
redpanda: move statefulset containers to go

jan-g:


RafalKorepta (approved):
I wonder if we could release this change to cloudsmith or other artifactory to test this with operator crd. 
LGTM

pkg/gotohelm/helmette/shims.go
```diff
@@ -144,6 +147,17 @@ func UnmarshalInto[T any](value any) T {
 	return t
 }
 
+// UnmarshalYaml fills in the type requested
+// +gotohelm:builtin=fromYamlArray
```
chrisseto, line 151: Only concern here is that we might want to wrap this in an actual shim. Helm's implementation of `fromYamlArray` will catch errors and stuff the stringified error into the returned array. That might be acceptable for now as the chart won't be able to apply invalid objects but if this is ever used to unmarshal a `[]string`, we're gonna have a bad time.

https://github.com/helm/helm/blob/4278ada158a91c7476b96a673280d87b4668aa78/pkg/engine/funcs.go#L106-L119

chrisseto (approved):

pkg/gotohelm/helmette/shims.go
jan-g: for posterity: there's nothing really dissimilar to the situation if you supply a values.yaml with:

```
...
  extraVolumeMounts: |-
    - "oops, a string!"
```

... the older chart doesn't catch this until the point it attempts to upload it to k8s

PR #1417, opened 2024-07-11
redpanda: move statefulset.yaml to go

jan-g:


jan-g:
Needs (a) tidying up the cruft; (b) regenerate the test values; (c) investigate the source of the differences between the config checksum values in the old and the new.

charts/redpanda/statefulset.go
```diff
@@ -847,3 +886,442 @@ func templateToContainers(dot *helmette.Dot, template string) []corev1.Container
...
+					PriorityClassName:             values.Statefulset.PriorityClassName,
+					Tolerations:                   statefulSetTolerations(dot),
+				},
+			},
+			VolumeClaimTemplates: nil, //TODO
+		},
+	}
+
+	// TopologySpreadConstraints
+	// Was protected with this: semverCompare ">=1.16-0" .Capabilities.KubeVersion.GitVersion
```
chrisseto, line 940: I think it's fairly safe to drop support for 1.16. Turns out the minimum required version defined by Chart.yaml is 1.21. So very safe to drop actually.

pkg/gotohelm/helmette/sprig.go
```diff
@@ -400,3 +402,9 @@ func Join[T any](sep string, s []T) string {
 	}
 	return out
 }
+
+// +gotohelm:builtin=sha256sum
+func Sha256Sum(input string) string {
+	hash := sha256.Sum256([]byte(input))
+	return hex.EncodeToString(hash[:])
```
chrisseto, line 409: Remind me the purpose of `[:]`? I was under the impression that it's functionally a no-op or there's something tricky about strings vs slices vs arrays.

charts/redpanda/statefulset.go
jan-g: In which case I'l drop the "Capabilities" bit :-)

pkg/gotohelm/helmette/sprig.go
jan-g: Sum256 returns a fixed-length array; EncodeToSting wants a slice.

charts/redpanda/statefulset.go
jan-g: (done)

chrisseto (approved):

PR #1418, opened 2024-07-11
console: add initial draft of Values struct

chrisseto:
This commit adds the initial go version of the console helm chart's values.yaml. The typing is primarily based off `values.yaml`, what can be cleaned from the chart, and the json schema. A test has been added to ensure that `values.yaml` and the struct are YAML/JSON compatible.

A few minor changes have been made to the `values.yaml` file to ease the typing consistency.

- `topologySpreadConstraints` is treated as a slice in the chart but defaults to an empty dict. It is now an empty slice.
- `livenessProbe.initialDelaySeconds` defaults to 0 but the underlying Kubernetes type specifies `omitempty`. It's defaulting has been moved to runtime to match the marshaling behavior.


chrisseto:
Note to me: Need to update the partial values and there's a tiny bug in genpartial that mucks up tags if `omitempty` is specified in the target struct 🤔

RafalKorepta (approved):

PR #1428, opened 2024-07-17
redpanda: handle cert = "" 

chrisseto:
Prior to this commit, the redpanda chart would fail to render if the admin's
internal TLS set cert to "" due to the order of operations in
post_upgrade_job.go. The regression appeared during the conversion to go as
there was previously no error handling if an invalid certificate was
referenced.

This failure surfaced during an attempted release of the operator as a helm
test for the operator chart did exactly this. (See
https://github.com/redpanda-data/helm-charts/pull/1422)

This commit corrects the issue by checking TLS.IsEnabled _before_ calling
MustGet which will only trigger a failure if TLS is enabled and references an
invalid certificate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,16 @@
 
 ## Redpanda Chart
 
-### [Unreleased](https://github.com/redpanda-data/helm-charts/releases/tag/redpanda-FILLMEIN) - YYYY-MM-DD
+### [Unreleased](https://github.com/redpanda-data/helm-charts/releases/tag/redpanda-5.8.13) - YYYY-MM-DD
+#### Added
+#### Changed
+#### Fixed
+* Fixed a regression where `post_upgrade_job` would fail if TLS on the admin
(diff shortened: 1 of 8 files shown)


chrisseto:
Dang. Just realized the operator test won't pass until there's a release of the redpanda chart with this commit...

CHANGELOG.md
```diff
@@ -2,7 +2,16 @@
 
 ## Redpanda Chart
 
-### [Unreleased](https://github.com/redpanda-data/helm-charts/releases/tag/redpanda-FILLMEIN) - YYYY-MM-DD
+### [Unreleased](https://github.com/redpanda-data/helm-charts/releases/tag/redpanda-5.8.13) - YYYY-MM-DD
```
RafalKorepta, line 5: NIT:

Why link of Unreleased points to `https://github.com/redpanda-data/helm-charts/releases/tag/redpanda-5.8.13`?

chrisseto: Whoops, that's a mistake on my behalf! I'll fix it the next time I update the notes.

PR #1430, opened 2024-07-19
connectors: set up values.go

jan-g:


charts/connectors/values.go
```diff
@@ -0,0 +1,242 @@
...
+type Logging struct {
+	Level string `json:"level"`
+}
+
+type Monitoring struct {
+	Enabled           bool              `json:"enabled"`
+	ScrapeInterval    Duration          `json:"scrapeInterval"`
+	Labels            map[string]string `json:"labels"`
+	Annotations       map[string]string `json:"annotations"`
+	NamespaceSelector struct {
```
chrisseto, line 123: nit: there's a type for this in the `monitoringv1` package that can be found in the imports of redpanda's values.go. The struct is otherwise correct though!

charts/connectors/values.go
```diff
@@ -0,0 +1,242 @@
...
+		Any bool `json:"any"`
+	} `json:"namespaceSelector"`
+}
+
+type Duration time.Duration
```
chrisseto, line 128: I think `metav1.Duration` will work for this type instead of a custom one?

charts/redpanda/ci/09-initcontainers-resources-values.yaml
```diff
@@ -73,13 +73,13 @@ statefulset:
...
           mountPath: /fake/lifecycle
     extraInitContainers: |-
-      - name: "test-init-container"
-        image: "mintel/docker-alpine-bash-curl-jq:latest"
-        command: [ "/bin/bash", "-c" ]
-        args:
-          - |
-            set -xe
-            echo "Hello World!"
+    - name: "test-init-container"
```
chrisseto, line 76: What lead to this being reformatted? 🤔

charts/connectors/values.go
```diff
@@ -0,0 +1,242 @@
...
+		MaxUnavailable int32 `json:"maxUnavailable"`
+	} `json:"budget"`
+	Annotations    map[string]string `json:"annotations"`
+	LivenessProbe  *corev1.Probe     `json:"livenessProbe,omitempty"`
+	ReadinessProbe *corev1.Probe     `json:"readinessProbe,omitempty"`
+	ExtraEnv       []struct {
+		Name  string `json:"name"`
+		Value string `json:"value"`
+	} `json:"extraEnv"`
+	ExtraEnvFrom []struct {
```
chrisseto, line 189: nit: `[]corev1.EnvFromSource` (I think?) works here though it's a bit more permissive.

charts/connectors/values.go
```diff
@@ -0,0 +1,242 @@
...
+	UpdateStrategy struct {
+		Type string `json:"type"`
+	} `json:"updateStrategy"`
+	Budget struct {
+		MaxUnavailable int32 `json:"maxUnavailable"`
+	} `json:"budget"`
+	Annotations    map[string]string `json:"annotations"`
+	LivenessProbe  *corev1.Probe     `json:"livenessProbe,omitempty"`
+	ReadinessProbe *corev1.Probe     `json:"readinessProbe,omitempty"`
+	ExtraEnv       []struct {
```
chrisseto, line 185: nit: `[]corev1.EnvVar`

chrisseto (approved):
I think you'll need a few `omitempty`s. As for the lack of roundtripping of the resources in `values.yaml` it should be pretty safe for us to change the default values, especially if the old defaults remain acceptable.

charts/redpanda/ci/09-initcontainers-resources-values.yaml
jan-g: that's a glitch; I was checking a bug report. I'll drop this.

charts/connectors/values.go
jan-g: These two get massaged into much the same thing, but sure.

PR #1439, opened 2024-07-22
console: convert secret.yaml to go

### LINKED ISSUE #1398: console: convert `secret.yaml` to go code
JIRA Link: [K8S-280](https://redpandadata.atlassian.net/browse/K8S-280)

chrisseto:
This commit converts secret.yaml to go. Notably, it opts to use `StringData` instead of `Data`. The `Data` field relies on json's marshaling behavior of base64 encoding `[]byte`s, functionality that's not presently available in gotohelm. Attempts to support this behavior would likely be quite involved or result in divergences between go and helm as there's no way to produce a byte slice in helm.

Fixes #1398


jan-g (approved):
I quite like this approach.

PR #1456, opened 2024-07-30
redpanda: fix image.repository in nightly task

chrisseto:
`image.repository` was previously missed. It is now correctly set to redpanda-unstable.


chapmanc (approved):
LGTM

PR #1459, opened 2024-08-01
Remove status field from resources

wdonne:
A Helm chart is not supposed to generate status fields in the resources. Although it may not change the behaviour in this case, it will keep the resource "out of sync" in GitOps tools.


chrisseto:
Hey @wdonne! This approach is not going to work. Our charts are generated from a go code base. The generation (run in CI) will immediately overwrite your changes.

Our charts follow the JSON/YAML marshaling semantic of Go and therefore Kubernetes' types itself. The Status field is always included due to it go's zero value system but the Kubernetes API ignores any changes to it as it's a sub-resource. Can you provide a bit more information about what GitOps tool you're using the misbehavior you're experiencing?

wdonne:
Hi @chrisseto ,

I understand that my changes won't work.

The GitOps tool I use is ArgoCD. It supports Helm charts. Since a rendered Helm chart is one big YAML file with all resources in it, it will be considered as the desired state. If the desired state contains non-empty status fields, then those resources will be out of sync immediately because the status is not supposed to be managed by ArgoCD but by the controller that owns the resource.

chrisseto:
Interesting! I'd almost argue that's a bug in ArgoCD but we're really the exception here. I'll see what we can do.

wdonne:
You could indeed say it is also a bug in ArgoCD because it never makes sense to compare the status fields.

chrisseto:
See https://github.com/redpanda-data/helm-charts/pull/1462

PR #1462, opened 2024-08-02
*: unset `status` and `creationTimestamp` before rendering

### LINKED ISSUE #1458: PodDisruptionBudget generated with status field
### What happened?

When rendering the Helm chart the PodDisruptionBudget resource is generated with a status field.

No status field. Only a controller should produce that.

<details>

```console
$ helm helm template redpanda redpanda-data/redpanda
# paste output here
---
# Source: redpanda/templates/poddisruptionbudget.yaml
apiVersion: policy/v1
kind: PodDisruptionBudget
metadata:
  creationTimestamp: null
  labels:
    app.kubernetes.io/component: redpanda
    app.kubernetes.io/instance: redpanda
    app.kubernetes.io/managed-by: Helm
    helm.sh/chart: redpanda-5.8.13
  name: redpanda
  namespace: default
spec:
  maxUnavailable: 1
  selector:
    matchLabels:
      app.kubernetes.io/component: redpanda-statefulset
      app.kubernetes.io/instance: redpanda
      redpanda.com/poddisruptionbudget: redpanda
status:
  currentHealthy: 0
  desiredHealthy: 0
  disruptionsAllowed: 0
  expectedPods: 0
```

</details>

_No response_

Redpanda

### Chart Version(s)

<details>

```console
5.8.13
```

</details>

### Cloud provider

<details>
NA
</details>

JIRA Link: [K8S-308](https://redpandadata.atlassian.net/browse/K8S-308)

[K8S-308]: https://redpandadata.atlassian.net/browse/K8S-308?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

chrisseto:
Previously our helm charts would render out Kubernetes objects the same way that calling `json.Marshal` would. This causes our charts to include the zero value for most `.Status` fields which is generally excluded from distributed manifests. It's been reported that ArgoCD will attempt to reconcile `.status` if specified and constantly report the resource as being out of sync.

An invariant of gotohelm is that it produces the same output as the provided go package. Rather than attempting to break this invariant, this commit bundles a new helper into `_shims.tpl`, `_shims.render-manifest`, for rendering the results of gotohelm templates which will remove `status` and `creationTimestamp` before rendering the object to YAML.

This commit also renames all `.go.tpl` files to begin with an `_` as is expected of partial templates.

Fixes #1458


RafalKorepta (approved):
LGTM, Thank you for the cleanup in Redpanda chart generated file names.

PR #1466, opened 2024-08-05
Fix volume mount without volume

RafalKorepta:
When internal communication to Admin API is configured to have mutual TLS configuration the volumes and volume mounts need to have additional mtls-client certificate included. That certificate could be generated in volume mounts if internal Admin API has disabled mutual TLS, but any other internal listener has it enabled.

### DIFF
--- a/charts/redpanda/chart_template_test.go
+++ b/charts/redpanda/chart_template_test.go
@@ -19,6 +19,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"golang.org/x/tools/txtar"
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/util/jsonpath"
 	"k8s.io/utils/ptr"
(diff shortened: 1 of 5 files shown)


charts/redpanda/chart_template_test.go
```diff
@@ -118,6 +123,36 @@ func TestTemplate(t *testing.T) {
 	}
 }
 
+func AssertVolumeMountsVerification(t *testing.T, manifests []byte) {
+	objs, err := kube.DecodeYAML(manifests, redpanda.Scheme)
+	require.NoError(t, err)
+
+	for _, obj := range objs {
+		sts, ok := obj.(*v1.StatefulSet)
```
chrisseto, line 131: I think we'll be doing ourselves a lot of favors if we make this work for more than just `StatefulSets`. If you think that's too much work for now can this be renamed to showcase that it's statefulset specific? Something like `ASSERT-STATEFULSET-VOLUME-MOUNTS-VERIFICATION`.

We could recursively crawl the members of a struct looking for PodTemplates using `reflect` 😅

charts/redpanda/chart_template_test.go
```diff
@@ -18,6 +18,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"golang.org/x/tools/txtar"
+	v1 "k8s.io/api/apps/v1"
```
chrisseto, line 21: nit: For consistency let's import this as `appsv1`

chrisseto (approved):

charts/redpanda/chart_template_test.go
RafalKorepta: ok

charts/redpanda/chart_template_test.go
RafalKorepta: I changed the annotation. I will leave enhancing to find PodTemplates using `reflect` for later :)

PR #1467, opened 2024-08-05
Assert all volumes are used

RafalKorepta:
Remove from the StatefulSet `-fs-validator` if it is not used.


charts/redpanda/chart_template_test.go
```diff
@@ -118,6 +123,37 @@ func TestTemplate(t *testing.T) {
 	}
 }
 
+func AssertAllVolumesAreUsed(t *testing.T, manifests []byte) {
```
chrisseto, line 126: Is this the same assertion as https://github.com/redpanda-data/helm-charts/pull/1466/files#diff-9816759f5552bf9796d055b4dd5bb21fe21633582b33349ba828ac5d4092ffe2R126 ?

It isn't technically an error to have an extra volume that's not referenced in a volume mount (though it is likely bad practice :P )

chrisseto (commented):
Can you explain the relation between this PR and #1466 ?

I also don't understand why you say this is a breaking change. Wouldn't this just be a small "bug" fix?

charts/redpanda/chart_template_test.go
RafalKorepta: > It isn't technically an error to have an extra volume that's not referenced in a volume mount (though it is likely bad practice :P )

Yes, I agree that it's not an error.

> Is this the same assertion as https://github.com/redpanda-data/helm-charts/pull/1466/files#diff-9816759f5552bf9796d055b4dd5bb21fe21633582b33349ba828ac5d4092ffe2R126 ?

That assertion is other way around. In this change we are making sure that all volumes has at least one corresponding volume mount.

In the other PR it is all volume mounts have at least one corresponding volume.

> I also don't understand why you say this is a breaking change. Wouldn't this just be a small "bug" fix?

It is breaking change as StatefulSet does not support changes to the volumes. 

I will bump redpanda chart version to show you it will error out during chart testing execution.

RafalKorepta:
I should stand corrected. My mental model always think that changing anything other than few fields would end up in the following error. Apparently that's not the case as volumes is part of template which could be changed. 
```
Forbidden: updates to statefulset spec for fields other than 'replicas', 'ordinals', 'template', 'updateStrategy', 'persistentVolumeClaimRetentionPolicy' and 'minReadySeconds' are forbidden
```

My intuition and my memory is tainted by the fact I was fighting over changes in `volumeClaimTemplates` in 
https://github.com/redpanda-data/redpanda-operator/blob/923dc4c618b664b2c2ba1fc1f5d9877d18b5d343/src/go/k8s/pkg/resources/statefulset_update.go#L544-L548
https://github.com/redpanda-data/redpanda/pull/2728

chrisseto (approved):
LGTM!

PR #1472, opened 2024-08-06
Add: appProtocol as an option in listeners admin configurations

lemonprogis:
Hello 👋 

When exposing the admin API using Istio as this port isn't made available since it's not an http default name. 
image

When I edit the service and add appProtocol: http the port becomes available with in istio.
image
now it's available.
image

Since the listeners section is configurable, I added an additional optional key under `listeners.admin`, but gives the ability to allow instrumentation to pickup that it's not just raw tcp per [K8s Service documentation](https://kubernetes.io/docs/concepts/services-networking/service/#application-protocol)

[Istio Docs around automatic detection](https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/)


RafalKorepta:
Could you run the following command?
```
nix develop -c task ci:lint
```

RafalKorepta (approved):

charts/redpanda/values.go
```diff
@@ -1010,9 +1010,10 @@ func (t *ExternalTLS) IsEnabled(i *InternalTLS, tls *TLS) bool {
 }
 
 type AdminListeners struct {
-	External ExternalListeners[AdminExternal] `json:"external"`
-	Port     int32                            `json:"port" jsonschema:"required"`
-	TLS      InternalTLS                      `json:"tls" jsonschema:"required"`
+	External    ExternalListeners[AdminExternal] `json:"external"`
+	Port        int32                            `json:"port" jsonschema:"required"`
+	AppProtocol *string                          `json:"appProtocol" jsonschema:"optional"`
```
RafalKorepta, line 1015: Should this change to
```suggestion
	AppProtocol *string                          `json:"appProtocol,omitempty" jsonschema:"optional"`
```

RafalKorepta (changes requested):

charts/redpanda/service_internal.go
```diff
@@ -38,14 +38,26 @@ func ServiceInternal(dot *helmette.Dot) *corev1.Service {
...
-	ports = append(ports, corev1.ServicePort{
-		Name:       "admin",
-		Protocol:   "TCP",
-		Port:       values.Listeners.Admin.Port,
-		TargetPort: intstr.FromInt32(values.Listeners.Admin.Port),
-	})
+
+	appProtocol := values.Listeners.Admin.AppProtocol
+
+	if appProtocol != nil {
```
chrisseto, line 45: There's no need for this check. As `values.Listeners.Admin.AppProtocol` defaults to `nil` and `ServicePort.AppProtocol` also defaults to `nil` you can always include it.

charts/redpanda/values.go
chrisseto: Optional is the default for jsonschema. To remain consistent let's not specify it in the tag.

charts/redpanda/values.yaml
```diff
@@ -931,6 +931,8 @@ listeners:
   admin:
     # -- The port for both internal and external connections to the Admin API.
     port: 9644
+    # -- Optional instrumentation hint - https://kubernetes.io/docs/concepts/services-networking/service/#application-protocol
+    # appProtocol: "http"
```
chrisseto, line 935: It appears there's a weird quirk of `helm-docs` where it doesn't really handle commented out lines. You'll notice that `README.md` now states that `"http"` is the default value which isn't true.

```suggestion
    # appProtocol:
```

chrisseto (commented):
Would you mind also adding a regression test by appending an example value to `charts/redpanda/testdata/template-cases.txtar`?

A minimal example is all that's needed. For the assertions `ASSERT-NO-ERROR` and `ASSERT-FIELD-EQUALS` should do it.

See https://github.com/redpanda-data/helm-charts/blob/41cc48195bb690db78739be79906c22c434fd932/charts/redpanda/testdata/template-cases.txtar#L165-L175

charts/redpanda/service_internal.go
lemonprogis: this is great, thanks for the feedback, changing now.

lemonprogis:
@chrisseto I'm unsure why those tests are failing, any help would be appreciated thanks!

chrisseto:
The chart testing failures are due to docker's rate limits. We generally re-run those until they succeed (or get a new runner with a new IP really)

The go tests are due to changes in the chart's output. `appProtocol` is now included as `null` in the output due to a quick in gotohelm, which is technically acceptable. 

To update the golden files run `nix develop -c go test ./charts/redpanda -short -update` and commit the changes. Rafal or I can push the changes in if you'd prefer.

lemonprogis:
Ok cool, thanks for the offer, but I just pushed the change!

RafalKorepta:
Due to unexpected quick release of Redpanda v24.2.2 I merged changes to `template-cases.golden.txtar`. Could you rebase and re-run  `nix develop -c go test ./charts/redpanda -short -update` and push changes?

RafalKorepta:
I couldn't push changes to your fork. Thank you for the contribution. It will be released soon.

Issue #1381, opened 2024-06-25
Redpanda fails to start and console failed to validate SASL config after 5.8.9 upgrade


zegerius:
### What happened?

Starting our development cluster Redpanda causes the following error moving from 5.8.8 to 5.8.9:

Redpanda doesn't become healthy:
```
  Warning  Unhealthy  25s (x64 over 9m55s)  kubelet            Readiness probe failed: command "/bin/sh -c set -x\nRESULT=$(rpk cluster health)\necho $RESULT\necho $RESULT | grep 'Healthy:.*true'\n" timed out
```

Which is odd, because when I exec the `rpk cluster health` command in the container:
```
redpanda@redpanda-0:/$ rpk cluster health
CLUSTER HEALTH OVERVIEW
=======================
Healthy:                          true
Unhealthy reasons:                []
Controller ID:                    0
All nodes:                        [0]
Nodes down:                       []
Leaderless partitions (0):        []
Under-replicated partitions (0):  []
```

That is potentially due to a faulty readiness check command:
```
redpanda@redpanda-0:/$ rpk cluster health
CLUSTER HEALTH OVERVIEW
=======================
Healthy:                          true
Unhealthy reasons:                []
Controller ID:                    0
All nodes:                        [0]
Nodes down:                       []
Leaderless partitions (0):        []
Under-replicated partitions (0):  []
redpanda@redpanda-0:/$ /bin/sh -c set -x\nRESULT=$(rpk cluster health)\necho $RESULT\necho $RESULT | grep 'Healthy:.*true'\n
redpanda@redpanda-0:/$ echo $?
1
```

Which could be simplified to:
```
redpanda@redpanda-0:/$ rpk cluster health | grep -q 'Healthy:.*true'
redpanda@redpanda-0:/$ echo $?
0
```

But I don't see any changes to the readiness check as compared to the previous version, I think?

I am not sure if that is related to `timeout` as k8s reports. Otherwise there is something else going on.

Additionally, the console doesn't start with an error I have never seen before. This seems to be related to the Go migration?
```json
{
	"level": "fatal",
	"msg": "failed to validate config",
	"error": "failed to validate Kafka config: failed to validate sasl config: given sasl mechanism '%!s(<nil>)' is invalid"
}
```

### What did you expect to happen?

Normal startup.

### How can we reproduce it (as minimally and precisely as possible)?. Please include values file.

<details>

```console
[...]

PR #1475, opened 2024-08-07
Fix mechanism check in superuser creation

RafalKorepta:
If superuser list is provided in values and mechanism is missing, the check for not equal to empty string does not match as in template engine mechanism is `nil`.

Fixes https://github.com/redpanda-data/helm-charts/issues/1381

### DIFF
--- a/charts/redpanda/chart_template_test.go
+++ b/charts/redpanda/chart_template_test.go
@@ -125,11 +125,48 @@ func TestTemplate(t *testing.T) {
 					require.NoError(t, err)
 					AssertStatefulsetAllVolumesAreUsed(t, out)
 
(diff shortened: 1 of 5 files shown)


charts/redpanda/chart_template_test.go
```diff
@@ -133,6 +137,39 @@ func TestTemplate(t *testing.T) {
 	}
 }
 
+func AssertSuperUsersAreValid(t *testing.T, manifests []byte) {
```
chrisseto, line 140: A static assertions might be better here? I don't think this would catch invalid formatting like `user:%s(nil)`. It does catch the existing issue though because of the switch.

chrisseto (approved):
Fix LGTM! Up to you about if you want to change the assertion or not.

PR #1477, opened 2024-08-07
redpanda: allow overriding (parts) of pod specs

chrisseto:
Previously it was not possible to individually set `annotations`, `labels`, or securityContexts on the Pods of the post install/upgrade Jobs. The inability to do so may prevent certain users from being able to run our chart within their environment. Namely, those using istio or those with strict Pod security requirements.

This commit extends the pattern of `statefulset.podTemplate` to apply to both the post install and post upgrade jobs. It also adds the ability to set pod and container security contexts via the pod template by implementing a good enough version of strategic merge patch that primarily relies on sprig's `merge`.

Additionally, this commit deprecates a handful of fields that are now better served via `podTemplate`.


chrisseto:
There's a couple failing tests that I need to clean up but I want to get some opinions before diving too much deeper into this.

I've been trying to avoid fully implementing SMP/JSON MP but I'm wondering if it would be better to go all in rather than a half baked approach.

@JakeSCahill , if you have any suggestions on how to better document this I'm all ears. I think I've been looking at it too long to properly explain it.

I also needed to check how the updated README is now rendering :)

charts/redpanda/values.go
```diff
@@ -539,51 +552,58 @@ func (s *Storage) StorageMinFreeBytes() int64 {
...
+	SecurityContext *corev1.PodSecurityContext `json:"securityContext,omitempty"`
 }
 
 type PodTemplate struct {
-	Labels      map[string]string `json:"labels" jsonschema:"required"`
-	Annotations map[string]string `json:"annotations" jsonschema:"required"`
-	Spec        PodSpec           `json:"spec" jsonschema:"required"`
+	Labels      map[string]string `json:"labels,omitempty" jsonschema:"required"`
+	Annotations map[string]string `json:"annotations,omitempty" jsonschema:"required"`
+	Spec        PodSpec           `json:"spec,omitempty" jsonschema:"required"`
```
RafalKorepta, line 606:
```suggestion
	Spec        *PodSpec           `json:"spec,omitempty" jsonschema:"required"`
```

charts/redpanda/statefulset.go
```diff
@@ -583,7 +562,7 @@ func statefulSetContainerRedpanda(dot *helmette.Dot) *corev1.Container {
 	container := &corev1.Container{
 		Name:  Name(dot),
 		Image: fmt.Sprintf(`%s:%s`, values.Image.Repository, Tag(dot)),
-		Env:   StatefulSetRedpandaEnv(dot),
+		Env:   statefulSetRedpandaEnv(),
```
RafalKorepta, line 565:
```suggestion
		Env:   StrategicMergePatch(values.Statefulset.PodTemplate.Spec.Containers.Env, statefulSetRedpandaEnv()),
```

charts/redpanda/helpers_test.go
```diff
@@ -0,0 +1,153 @@
...
+	"testing"
+
+	"github.com/redpanda-data/helm-charts/charts/redpanda"
+	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
+)
+
+func TestStategicMergePatch(t *testing.T) {
```
RafalKorepta, line 13: Could you add EnvVar example of appending overwrite on top of what original has, just to visualise this in one of those test cases?

RafalKorepta (approved):
I like this approach. In my opinion without change to values definition eventually helmette package (in shims) or helpers.go should implement generic strategic merge patch.

charts/redpanda/helpers_test.go
chrisseto: Done! And it caught a but 😅

PR #1479, opened 2024-08-07
Bump operator and redpanda versions

RafalKorepta:


charts/operator/Chart.yaml
```diff
@@ -6,11 +6,11 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 0.4.26
+version: 0.4.27
 
 # This is the default version of the operator being deployed.
 # ** NOTE for maintainers: please enssure the artifacthub image annotation is updated before merging
-appVersion: v2.1.28-24.2.1
+appVersion: v2.1.29-24.2.1
```
JakeSCahill, line 13: Should this be `v2.1.29-24.2.2`?

charts/operator/Chart.yaml
```diff
@@ -34,11 +34,11 @@ annotations:
       url: https://helm.sh/docs/intro/install/
   artifacthub.io/images: |
     - name: redpanda-operator
-      image: docker.redpanda.com/redpandadata/redpanda-operator:v2.1.28-24.2.1
+      image: docker.redpanda.com/redpandadata/redpanda-operator:v2.1.29-24.2.1
     - name: configurator
-      image: docker.redpanda.com/redpandadata/configurator:v2.1.28-24.2.1
+      image: docker.redpanda.com/redpandadata/configurator:v2.1.29-24.2.1
```
JakeSCahill, line 39: Seee https://github.com/redpanda-data/helm-charts/pull/1479/files#r1709170158

charts/operator/Chart.yaml
RafalKorepta: It should but Chris was quicker with the release. I could create a new tag and ask devprod to remove the old image

chrisseto (approved):
LGTM once the new operator image is ready. 

Thanks for filling out release notes! Maybe we should add a linter to help remind us to fill them out.

RafalKorepta:
It would be good to have linter, but I'm not sure where and when it should be executed.

Issue #1476, opened 2024-08-07
Redpanda console helm chart 0.7.27 missing image tag when deploying console


brokenjacobs:
### What happened?

Upgraded to 0.7.27 using semantic versioning in my CD and the chart fails to deploy and rolls back due to a missing image tag in the deployment of the console.

### What did you expect to happen?

A tag be specified for the chart

### How can we reproduce it (as minimally and precisely as possible)?. Please include values file.

<details>

```console
autoscaling:
  enabled: false
  maxReplicas: 100
  minReplicas: 1
  targetCPUUtilizationPercentage: 80
console:
  config:
    connect:
      clusters:
      - name: lc-mirrormaker2
        url: http://lc-mirrormaker2-mirrormaker2-api:8083
      connectTimeout: 15s
      enabled: true
      readTimeout: 60s
      requestTimeout: 6s
    kafka:
      brokers:
      - lc-standalone-kafka-bootstrap:9095
      clientId: redpanda-console
      protobuf:
        enabled: true
        schemaRegistry:
          enabled: true
      schemaRegistry:
        enabled: true
        tls:
          enabled: false
        urls:
        - http://apicurio-registry-kafkasql.kafka.svc.cluster.local/apis/ccompat/v6
      tls:
        caFilepath: /etc/kafka/certs/ca.crt
        certFilepath: /etc/kafka/user-cert/user.crt
        enabled: true
        keyFilepath: /etc/kafka/user-cert/user.key
    logger:
      level: info
    metricsNamespace: console
extraEnv: []
extraVolumeMounts:
- mountPath: /etc/kafka/user-cert
  name: kafka-user-certs
  readOnly: true
- mountPath: /etc/kafka/certs
  name: kafka-certs
  readOnly: true
extraVolumes:
- name: kafka-user-certs
  secret:
    secretName: redpanda-console-user
- name: kafka-certs
  secret:
    secretName: lc-standalone-cluster-ca-cert
ingress:
  annotations:
    traefik.ingress.kubernetes.io/router.middlewares: traefik-system-compress@kubernetescrd
    traefik.ingress.kubernetes.io/router.tls: "true"
  className: traefik-internal
  enabled: true
  hosts:
  - host: kafkaconsole.***
    paths:
    - path: /
      pathType: ImplementationSpecific
  tls:
  - hosts:
    - kafkaconsole.***
    secretName: lc-wildcard-tls
resources:
  limits:
    cpu: "0.5"
    memory: 256Mi
[...]

PR #1483, opened 2024-08-09
fix(console): container image tag empty

jgrasl:
Deref only checks for nil but the tag is provided as empty in the values.yml resulting in console: instead of console:tag

Fixes: [1476](https://github.com/redpanda-data/helm-charts/issues/1476)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -141,6 +141,13 @@
 
 ## Console Chart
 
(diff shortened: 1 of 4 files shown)


RafalKorepta (approved):

chrisseto:
Thanks for the fix @jgrasl! (We got the CLA assistant figured out)

PR #1484, opened 2024-08-09
Add port-forwarding dialer implementation

andrewstucki:
This adds a port-forwarding dialer implementation as discussed in https://github.com/redpanda-data/redpanda-operator/pull/181#discussion_r1707845922.

It allows us to simply swap in a `Dial`/`DialContext` implementation rather than going through an entire port-forwarding setup and targeting a local port in order for us to reach inside the Kubernetes network.

The thought is that this would be mainly for testing.


pkg/kube/dialer.go
```diff
@@ -0,0 +1,336 @@
...
+	dataStream, err := conn.CreateStream(headers)
+	if err != nil {
+		return nil, err
+	}
+
+	return wrapConn(network, address, dataStream, errorStream), nil
+}
+
+// DialContext is a simple wrapper around Dial.
+func (p *PodDialer) DialContext(_ context.Context, network string, address string) (net.Conn, error) {
```
chrisseto, line 110: Explicitly comment that `context.Context` is NOT respected here. There doesn't seem to be a super straightforward way to plumb context aware into spdy :/ that's going to come back and bite us but better to have it in code we own instead of hiding in between k8s code and an unrelated proxy connection.

pkg/kube/dialer.go
```diff
@@ -0,0 +1,336 @@
...
+	return p.Dial(network, address)
+}
+
+// parseDNS attempts to determine the intended pod to target, currently the
```
chrisseto, line 114: This is a beautiful comment! The symmetry of dialing within a k8s cluster is nice but I think it's too tailored towards StatefulSets, it introduces a needless information overhead, and the defaulting semantics could very easily get me in trouble.

Let's say I want to dial into a core DNS Pod so I can resolve the `kubernetes` service to find the cluster domain. I won't have a cluster domain as I'm trying to fetch it, the supported formats aren't valid hostnames within the cluster as core DNS isn't run under a headless service, and the namespace I'm dialing into (`kube-system`) is different from the default (Assuming that I'll be dialing into a different namespace later once I have the hostname).

I'd recommend only supporting `[pod-name].[namespace-name]`.

If we need to support the FQDN to make hostname verification work correctly, require the format to be `[pod-name].[service-name].[namespace-name].svc[.cluster-domain][.]` (cluster domain and trailing `.` would be optional). Ignore the validation of `service-name` and `cluster-domain` as hostname verification will handle that for you. This will remove any possible ambiguity between the two formats and ensure no defaulting takes place accidentally.

pkg/kube/dialer.go
```diff
@@ -0,0 +1,336 @@
...
+// If no namespace-name is supplied, the default namespace is assumed.
+// No validation is done to ensure that the pod is actually referenced by the
+// given service or that the DNS record exists in Kubernetes, instead this assumes
+// that things are set up properly in Kubernetes such that the FQDN passed here
+// matches Kubernetes' own DNS records and a pod within the cluster would be able
+// to resolve the same FQDN to the pod that we do.
+//
+// These assumptions allow us to use this dialer to issues requests *as if* we are
+// in the Kubernetes network even from outside of it (i.e. in tests that attempt to
+// connect to a pod at a given DNS address).
```
chrisseto, line 131: super nit: wouldn't the correct term be "given domain" or "given hostname"? I understand what you mean by DNS address but AFAIK it's not technically a correct term.

pkg/kube/dialer.go
```diff
@@ -0,0 +1,336 @@
...
+		Namespace(pod.Namespace).
+		Name(pod.Name).
+		SubResource("portforward")
+
+	dialer := spdy.NewDialer(upgrader, &http.Client{Transport: transport}, http.MethodPost, req.URL())
+	conn, protocol, err := dialer.Dial(portForwardProtocolV1Name)
+	if err != nil {
+		return nil, err
+	}
+	if protocol != portForwardProtocolV1Name {
```
chrisseto, line 199: super nit: separate if statements by a new line (We should just lint this)

pkg/kube/dialer.go
```diff
@@ -0,0 +1,336 @@
...
+		return nil, fmt.Errorf("unable to negotiate protocol: client supports %q, server returned %q", portForwardProtocolV1Name, protocol)
+	}
+
+	p.connections[pod] = conn
+	return conn, nil
+}
+
+type conn struct {
+	httpstream.Stream
```
chrisseto, line 208: nit: add a newline after embedded struct members to visually distinguish them. (Also something we should lint)

Is there a reason to embed the stream if we're overriding most of the methods it provides?

pkg/kube/dialer.go
```diff
@@ -0,0 +1,336 @@
...
+	errorStream httpstream.Stream
+	network     string
+	remote      string
+
+	errCh chan error
```
chrisseto, line 213: Is there a reason why this is separated from the other struct members?

pkg/kube/dialer.go
```diff
@@ -0,0 +1,336 @@
...
+
+func (c *conn) pollErrors() {
+	data, err := io.ReadAll(c.errorStream)
+	if err != nil {
+		c.writeError(err)
+		return
+	}
+
+	if len(data) != 0 {
+		c.writeError(errors.New(string(data)))
```
chrisseto, line 237: Let's add a bit of context here so it's possible to trackdown this error message if we ever encounter it. Alternatively, `errors.WithStack` would work just as well.

pkg/kube/dialer.go
```diff
@@ -0,0 +1,336 @@
...
+		return
+	}
+}
+
+func (c *conn) writeError(err error) {
+	select {
+	case c.errCh <- err:
```
chrisseto, line 244: Why is there a `default` here? `errCh` is buffered with 1 element and it appears to only ever be written to once. The default could lead to us unexpectedly dropping errors. Given that the `DialContext` doesn't support cancelation, we _really_ shouldn't be dropping errors even if this is test only code.

pkg/kube/dialer.go
```diff
@@ -0,0 +1,336 @@
...
+func (c *conn) checkError() error {
+	select {
+	case err := <-c.errCh:
+		return err
+	default:
+		return nil
+	}
+}
+
+var _ net.Conn = (*conn)(nil)
```
chrisseto, line 258: I'm used to seeing these assertions directly beneath the struct declaration. It feels a bit buried down here but it is kinda nice to see them grouped together. Which pattern have you seen more frequently?

pkg/kube/dialer.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+
+	go c.pollErrors()
+	return c
+}
+
+func (c *conn) pollErrors() {
+	data, err := io.ReadAll(c.errorStream)
+	if err != nil {
+		c.writeError(err)
```
chrisseto, line 232: If we encounter an error, should we go ahead and close the entire connection? It's unlikely that it will continue to work if an error is written, failing early could prevent test cases from hanging on a read/write to a wonky connection and it would vastly simplify the logic within Read/Write (IMO).

pkg/kube/dialer.go
```diff
@@ -0,0 +1,336 @@
...
+	if err := c.checkError(); err != nil {
+		return 0, err
+	}
+
+	n, err := c.Stream.Read(data)
+
+	// prioritize any sort of checks propagated on
+	// the error stream
+	if err := c.checkError(); err != nil {
+		return n, err
```
chrisseto, line 270: nit: `errors.Join` will let you return two errors.

pkg/kube/dialer.go
```diff
@@ -0,0 +1,336 @@
...
+	if err := c.checkError(); err != nil {
+		return 0, err
+	}
+
+	n, err := c.Stream.Write(b)
+
+	// prioritize any sort of checks propagated on
+	// the error stream
+	if err := c.checkError(); err != nil {
+		return n, err
```
chrisseto, line 285: nit: `errors.Join` will let you return two errors.

pkg/kube/dialer.go
```diff
@@ -0,0 +1,336 @@
...
+	return n, err
+}
+
+func (c *conn) Close() error {
+	closeErr := c.Reset()
+
+	// prioritize any sort of checks propagated on
+	// the error stream
+	if err := c.checkError(); err != nil {
+		return err
```
chrisseto, line 296: nit: `errors.Join` will let you return two errors.

pkg/kube/dialer.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+	return closeErr
+}
+
+func (c *conn) SetDeadline(t time.Time) error {
+	if conn, ok := c.Stream.(net.Conn); ok {
```
chrisseto, line 302: ooc, are there ever times that `c.Stream` would have a different implementation? As a reader of this code, I'd want to know if `c.Stream` is known to definitely or definitely not implement `net.Conn` but I now have more questions. It seems important to know one way or the other as context cancellation isn't plumbed through.

pkg/kube/dialer_test.go
```diff
@@ -0,0 +1,132 @@
...
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/tools/clientcmd"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func TestDialer(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
+	defer cancel()
+
+	container, err := k3s.Run(ctx, "rancher/k3s:v1.27.1-k3s1")
```
chrisseto, line 25: Let's chat about testcontainer's `k3s` vs `k3d` vs `kind` soon and come up with some guidelines. We now have 3 competing standards. Alternatively, if the k3s container starts up in less than 30s I may be sold.

pkg/kube/dialer.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+
+	// service - labels[1]
+	pod.Name = labels[0]
+	return
+}
+
+// mutex must be held while calling this
+func (p *PodDialer) connectionForPod(pod types.NamespacedName) (httpstream.Connection, error) {
+	if conn, ok := p.connections[pod]; ok {
```
chrisseto, line 170: I don't see connections ever getting cleaned up from `p.connections`. Rather than worrying about leaking connections and goroutines, I think we'll be better served by eating the overhead of additional connections and having each connection from `connectionForPod` manage it's own lifecycle by closing the parent connection whenever the connection itself is closed. WDYT?

pkg/kube/dialer.go
```diff
@@ -0,0 +1,336 @@
...
+		return conn, nil
+	}
+
+	transport, upgrader, err := spdy.RoundTripperFor(p.config)
+	if err != nil {
+		return nil, err
+	}
+
+	cfg := p.config
```
chrisseto, line 179: nit: It's not safe (I think) to copy rest configs in this manor. Prefer to use `DeepCopy` (or educate me if I'm mistaken I just checked if the method existed)

chrisseto (commented):
Very excited to get this merged in! I'm particularly picking on resource leakages and usability issues. I suspect this will quickly become a cornerstone of our testing endeavors and I'd like it to be as solid as we can get it.

pkg/kube/dialer.go
andrewstucki: I think I thought of a way where we can actually respect the Context by using a polling goroutine in the underlying connection implementation to just forcibly (and idempotently) Close the underlying stream. I'll push up the changes in a bit as I work through the rest of the comments here.

pkg/kube/dialer.go
andrewstucki: So, the idea is that an error in the underlying `errorStream` poller would cause the connection to terminate, so we really only care about the first error ever returned. That said, I'm cleaning up some of this to more directly indicate that the connection is no longer valid once we've received some sort of error and that everything should be trashed.

The reason for the buffered channel is to intentionally drop anything but the first write. You're right about the error poller only having written to the channel once initially, missed that 😬. I added a context checking goroutine which will also write here though, so I think I'll likely just keep this as is for now until you have a second pass review.

pkg/kube/dialer.go
andrewstucki: this got buried as I added more functions and I forgot to clean it up, moving it

pkg/kube/dialer.go
andrewstucki: So, would you prefer to switch to an unchecked cast? It's a bit cumbersome to track down, but the underlying k8s library eventually resolves to a [`*github.com/moby/spdystream.Stream`](https://github.com/moby/spdystream/blob/77eb080123d208697674a07b74ceaf94c98bee8b/stream.go#L323-L335) which implements `net.Conn`.

pkg/kube/dialer.go
andrewstucki: I'll see if I can do something a bit trickier, but with a bit less overhead. Going to just add some reference counting in the parent connection. When a stream is closed then it decrements its parent reference count, when it hits 0 it'll get pruned from the map.

pkg/kube/dialer.go
andrewstucki: I believe `*rest.Config` is fine to pass around, the `DeepCopy` that it has it actually inherits from embedding `*rest.TLSClientConfig`, so calling `DeepCopy` will return that rather than a `*rest.Config` - I believe you'd only have a problem if someone is modifying the config after you're passing it around due to synchronization issues.

pkg/kube/dialer.go
andrewstucki: it's a preference of mine to generally keep fields (even if private) separate from each other if one comes from some sort of initialization and the others are only used internally -- I'm generally even more inclined to do this with synchronization primitives -- just a habit I've developed over the years, but I can change it if you want.

andrewstucki:
@chrisseto - addressed some of the PR feedback and explained the changes in some comments to your comments. Let me know if you have any other questions. I left some of them unresolved too, so let me know if any of those are show-stoppers.

pkg/kube/dialer.go
andrewstucki: Leaving this unresolved because I'm not entirely sure whether or not we'd want to join the two errors together. Thoughts?

pkg/kube/dialer.go
chrisseto: Ah gotcha! I do something similar but with _slightly_ different "rules". No need to change. Just wanted to understand your reasoning. Thanks for explaining!

pkg/kube/dialer.go
chrisseto: Let's leave this unresolved as error handling in go is a fun topic. My rule of thumb here is "If either error could contain the information I might need for debugging, I want both of them"

pkg/kube/dialer.go
chrisseto: That or perhaps return an error? I think I'd lean towards a type assertion as I wouldn't want these method calls to suddenly start to do nothing silently.

pkg/kube/dialer.go
```diff
@@ -0,0 +1,442 @@
...
+	}
+}
+
+func (c *conn) checkError() error {
+	select {
+	case err := <-c.errCh:
+		return err
+	default:
+		if c.closed.Load() {
+			return net.ErrClosed
```
chrisseto, line 341: What was the reason for adding this check? It seems that omitting it would result in _roughly_ the same behavior, depending on the behavior of `.Read`/`.Write` on a closed stream.

chrisseto (approved):
LGTM!

Please squash and merge (mentioning because we haven't touched on that).

PR #1485, opened 2024-08-09
Expose redpanda global 'Dot' context publicly

andrewstucki:
Exposing a `Dot` method so that we can do roughly the following:

1. Initialize a `Dot` instance with some given `PartialValues`/release version.
2. Call into the individual object construction helpers (i.e. `ServiceInternal`) to generate a Kubernetes object with filled-in user-specified values + defaults
3. Use those rendered objects to do things like construct connection parameters for a cluster.


chrisseto (approved):
LGTM!

PR #1486, opened 2024-08-09
Add KafkaClient and AdminClient helpers to construct clients from a chart specification

andrewstucki:
This adds in a helper for constructing a client to a cluster based on a given chart configuration using our generator code.

This code re-implements the more ad-hoc code from the WIP PR https://github.com/redpanda-data/redpanda-operator/pull/181, allowing us to leverage more of the generator helpers and co-locating the code that actually knows how our resource names are constructed. This, combined with https://github.com/redpanda-data/helm-charts/pull/1484 and our existing helm client in this repo should allow me to re-implement the intent of the original PR (initializing clients from a Redpanda `ClusterRef`) in a much cleaner, more ergonomic way.


pkg/redpanda/client.go
```diff
@@ -0,0 +1,210 @@
...
+		return fmt.Errorf("error fetching client certificate default/%s: %w", clientCertName, err)
+	}
+	saslError := func(err error) error {
+		return fmt.Errorf("error fetching SASL authentication for %s/%s: %w", saslUsers.Namespace, saslUsers.Name, err)
+	}
+
+	if redpanda.TLSEnabled(dot) {
+		tlsConfig := &tls.Config{MinVersion: tls.VersionTLS12, ServerName: serverName}
+
+		serverCert, found := helmette.Lookup[corev1.Secret](dot, namespace, kafkaRootCertName)
```
chrisseto, line 134: I don't think we want to use this function outside of helm packages if we can avoid it. The lack of an explicit client specification, panics, and context support make it a poor choice.  We should probably add some redtape around it 😓 

The `kube` package's `Get` function has the same ergonomics albeit you'll need a `kube.Ctl` instance

pkg/redpanda/client.go
```diff
@@ -0,0 +1,210 @@
...
+		if err != nil {
+			return nil, serverTLSError(fmt.Errorf("unable to parse public key %w", err))
+		}
+		pool := x509.NewCertPool()
+		pool.AddCert(serverParsedCertificate)
+
+		tlsConfig.RootCAs = pool
+
+		if redpanda.ClientAuthRequired(dot) {
+			clientCert, found := helmette.Lookup[corev1.Secret](dot, "default", clientCertName)
```
chrisseto, line 155:
```suggestion
			clientCert, found := helmette.Lookup[corev1.Secret](dot, namespace, clientCertName)
```

pkg/redpanda/client.go
```diff
@@ -0,0 +1,210 @@
...
+	ErrSASLSecretNotFound          = errors.New("users secret not found")
+	ErrSASLSecretKeyNotFound       = errors.New("users secret key not found")
+	ErrSASLSecretSuperuserNotFound = errors.New("users secret has no users")
+
+	supportedSASLMechanisms = []string{
+		"PLAIN", "SCRAM-SHA-256", "SCRAM-SHA-512",
+	}
+)
+
+func firstUser(data []byte) (user string, password string, mechanism string) {
```
chrisseto, line 41: nit: prefer to place unexpected helpers at the bottom of the file and exported functions at the top in some degree of ordering.

pkg/redpanda/client.go
```diff
@@ -0,0 +1,210 @@
...
+			return nil, err
+		}
+		return tls.Client(conn, config), nil
+	}
+}
+
+// KafkaClient creates a client to talk to a Redpanda cluster based on its helm
+// configuration over its internal listeners.
+func KafkaClient(config kube.Config, release helmette.Release, partial redpanda.PartialValues, dialer DialContextFunc, opts ...kgo.Opt) (*kgo.Client, error) {
+	dot, err := redpanda.Dot(release, partial)
```
chrisseto, line 98: Should this function just accept a `Dot` instead of constructing it internally?

pkg/redpanda/client.go
andrewstucki: So, I'm not so sure about this namely because of what I pointed out in Slack (when you were on vacation, you may not have seen it). Found this bit in our chart code which seems to be wrong to me?

https://github.com/redpanda-data/helm-charts/blob/dd21905ce82532acf38680054639710985b3f4e1/charts/redpanda/certs.go#L114-L117

Notice the lack of a `Namespace` there.

pkg/redpanda/client.go
andrewstucki: Probably, it'd be a lot cleaner initialization to do so, I'll change it.

pkg/redpanda/client.go
andrewstucki: Just to keep the semantics fairly similar, I introduced a `SafeLookup` in the package that doesn't panic, the `Lookup` then follows more or less the behavior of a `Must*` function, it just calls `SafeLookup` and panics on any returned errors.

pkg/redpanda/client.go
chrisseto: Ah. I'm 90% certain that's a bug. We should be setting Namespace on all objects. I _think_ that `helm install` implicit sets it if not specified but I could be incorrect there. Do you have a link to the slack thread on hand?

andrewstucki: Sure, and I hadn't thought about the `helm install` bringing with it implicit namespaces (though if we're going to go about doing directly client-based installation, we should probably either pin the client to a particular namespace, or just explicitly set it on all objects).

Here's the [thread](https://redpandadata.slack.com/archives/C04K2N6S2A3/p1723222386293679) (though no one responded yet since y'all were OOO).

andrewstucki: So, looks like `helm install` does merge in namespaces implicitly, I'll open up another PR making this explicit here, though I'd definitely be interested in taking a look at the templating tests too to see how we can enforce explicit namespace setting in all of our namespaced resources.

chrisseto (approved):

PR #1492, opened 2024-08-19
release: redpanda 5.9.1 and console 0.7.29

chrisseto:
This commit releases the aforementioned versions of the redpanda and console charts.

The console chart was previously partially released. The Chart.yaml's version was not bumped which resulted in no release being created.


RafalKorepta (approved):
LGTM

PR #1498, opened 2024-08-23
operator: bump app version and release

chrisseto:
This commit updates the operator chart's `appVersion` to the fastest operator version `v2.2.0-24.2.2` and releases it.


andrewstucki (approved):
LGTM, though not familiar with the release process, so not sure what other steps there might be 😅

chrisseto:
For the helm charts, this is it. The rest is handled by a github action that runs chart-releaser.

PR #1499, opened 2024-08-24
gha: update to use oidc

andrewhsu:
jira: https://redpandadata.atlassian.net/browse/PESDLC-1731

use [oidc](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services) to retrieve secrets


chrisseto (approved):
LGTM!

PR #1285, opened 2024-05-17
operator: v1alpha2 mitigations

chrisseto:

PR #1509, opened 2024-09-03
Move operator to gotohelm

RafalKorepta:
To solve problem of missing security context this commit moves almost all resources to go. genschema is currently broken as I tried to solve particular issue with operator schema generation. Need to fix before merging this PR.


cmd/genschema/main.go
```diff
@@ -71,7 +81,7 @@ func main() {
 		},
 	}
 
-	schema := r.Reflect(&redpanda.Values{})
+	schema := r.Reflect(&operator.Values{})
```
chrisseto, line 84: Could we instead just parameterize this script? Running a switch on `os.Args[1]` should be pretty straight forward.

charts/operator/values.go
```diff
@@ -0,0 +1,154 @@
...
+					pair.Value,
+					{Type: "null"},
+				},
+			})
+		}
+		makeArrayNullableRecursive(pair.Value)
+	}
+}
+
+const (
```
chrisseto, line 90: nit: Move consts to the top of the file.

charts/operator/values.go
```diff
@@ -0,0 +1,154 @@
...
+
+type WrappedPodSpec corev1.PodSpec
+
+// +gotohelm:ignore=true
+func (WrappedPodSpec) JSONSchemaExtend(schema *jsonschema.Schema) {
+	makeArrayNullableRecursive(schema)
+}
+
+// +gotohelm:ignore=true
+func makeArrayNullableRecursive(schema *jsonschema.Schema) {
```
chrisseto, line 76: Would it make sense to just run this against the final generated schema in `genschema`?

charts/operator/values.go
```diff
@@ -0,0 +1,154 @@
...
+	Strategy                *appsv1.DeploymentStrategy    `json:"strategy,omitempty"`
+	Annotations             map[string]string             `json:"annotations,omitempty"`
+	PodAnnotations          map[string]string             `json:"podAnnotations"`
+	PodLabels               map[string]string             `json:"podLabels"`
+	AdditionalCmdFlags      []string                      `json:"additionalCmdFlags"`
+	CommonLabels            map[string]string             `json:"commonLabels"`
+	Monitoring              MonitoringConfig              `json:"monitoring"`
+	WebhookSecretName       string                        `json:"webhookSecretName"`
+	PodSpec                 *WrappedPodSpec               `json:"podSpec,omitempty"`
+	OperatorSecurityContext *corev1.SecurityContext       `json:"operatorSecurityContext,omitempty"`
```
chrisseto, line 47: Rather than hanging this value off the main struct, can we just copy the convention set forth in the redpanda helmchart were we perform a pseudo merge patch? Doing so will be a bit more ergonomic and more importantly consistent across our charts.

charts/operator/values.yaml
```diff
@@ -152,6 +159,26 @@ monitoring:
 # Sets the name of the Secret in which to store the self-signed TLS certificate for the Webhooks when `webhook.enabled` is `true`.
 webhookSecretName: webhook-server-cert
 
+# -- Sets almost all fields of operator Deployment PodTemplateSpec
```
chrisseto, line 162: Can we copy the documentation that's within the redpanda chart? IIRC that should include a few examples as well.

charts/operator/values.go
RafalKorepta: I removed `OperatorSecurityContext` and I took that value from `PodSpec`.

charts/operator/values.yaml
RafalKorepta: Agree

charts/operator/deployment.go
```diff
@@ -0,0 +1,271 @@
...
+	return dot.Chart.AppVersion
+}
+
+func isWebhookEnabled(dot *helmette.Dot) bool {
+	values := helmette.Unwrap[Values](dot.Values)
+
+	return values.Webhook.Enabled && values.Scope == Cluster
+}
+
+type PossibleEnvVar struct {
```
chrisseto, line 211: This appears unused?

charts/operator/helpers.go
```diff
@@ -0,0 +1,370 @@
...
+	}
+
+	if overrides.Spec.Affinity != nil {
+		original.Spec.Affinity = helmette.MergeTo[*corev1.Affinity](
+			overrides.Spec.Affinity,
+			helmette.Default(&corev1.Affinity{}, original.Spec.Affinity),
+		)
+	}
+
+	if overrides.Spec.TopologySpreadConstraints != nil && len(overrides.Spec.TopologySpreadConstraints) > 0 {
```
chrisseto, line 159: I don't think it was necessary to fully implement this when we just need to be able to override SecurityPolicy 😅 

Did you try just calling `merge` on the pod specs and then filling in fields that need special handling? I think I tried that for containers and it didn't work but I don't recall why....

charts/operator/rbac.go
```diff
@@ -0,0 +1,797 @@
...
+// +gotohelm:filename=_rbac.go.tpl
+package operator
+
+import (
+	"github.com/redpanda-data/helm-charts/pkg/gotohelm/helmette"
+	rbacv1 "k8s.io/api/rbac/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func ClusterRole(dot *helmette.Dot) []rbacv1.ClusterRole {
```
chrisseto, line 25: Note for the future (not this PR). We could add the ability to load files into gotohelm and just read these from disk so they don't get out of sync. Similar to what I did here: https://github.com/redpanda-data/helm-charts/pull/1285

charts/operator/serviceaccount.go
```diff
@@ -0,0 +1,59 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// Create the name of the service account to use
+func ServiceAccountName(dot *helmette.Dot) string {
+	values := helmette.Unwrap[Values](dot.Values)
+
+	if values.ServiceAccount.Create {
+		if values.ServiceAccount.Name != nil && *values.ServiceAccount.Name != "" {
```
chrisseto, line 30: nit: ptr.Deref :P

Edit: Feel free to ignore this. I want to write a linter and this will be a good test case.

chrisseto (approved):
No major blockers!

charts/operator/certificates.go
```diff
@@ -0,0 +1,85 @@
...
+				fmt.Sprintf("%s-webhook-service.%s.svc", Fullname(dot), dot.Release.Namespace),
+				fmt.Sprintf("%s-webhook-service.%s.svc.%s", Fullname(dot), dot.Release.Namespace, values.ClusterDomain),
+			},
+			IssuerRef: cmmeta.ObjectReference{
+				Kind: "Issuer",
+				Name: fmt.Sprintf("%s-selfsigned-issuer", helmette.Trunc(len("-selfsigned-issuer"), Fullname(dot))),
+			},
+			SecretName: values.WebhookSecretName,
+			PrivateKey: &certv1.CertificatePrivateKey{
+				// There is an issue with gotohelm when RotationPolicyNever is used.
```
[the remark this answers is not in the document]
RafalKorepta: I extended this comment with an explanation about what is failing.

charts/operator/chart_test.go
```diff
@@ -0,0 +1,249 @@
...
+		// Every 5 iterations, decrease nil chance to ensure that we're biased
+		// towards exploring most cases.
+		if i%5 == 0 && nilChance > .1 {
+			nilChance -= .1
+		}
+
+		var values PartialValues
+		fuzzer.NilChance(nilChance).Fuzz(&values)
+		// Special case as fuzzer does not assign correctly scope
+		s := OperatorScope(Namespace)
```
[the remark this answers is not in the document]
RafalKorepta: There is 50/51 split.
* [operator scope namespace](https://github.com/redpanda-data/helm-charts/pull/1509/files#diff-d7d89e287882d63ed9ce9a5e069524c6506f2e1e7669add7825429302cbe5a60R199)
* [operator scope cluster](https://github.com/redpanda-data/helm-charts/pull/1509/files#diff-d7d89e287882d63ed9ce9a5e069524c6506f2e1e7669add7825429302cbe5a60R165)

charts/operator/chart_test.go
```diff
@@ -0,0 +1,249 @@
...
+
+	archive := txtar.Format(&txtar.Archive{
+		Comment: []byte(fmt.Sprintf(`Generated by %s`, t.Name())),
+		Files:   files,
+	})
+
+	require.NoError(t, os.WriteFile("testdata/template-cases-generated.txtar", archive, 0o644))
+}
+
+func makeSureTagIsNotEmptyString(values PartialValues, fuzzer *fuzz.Fuzzer) {
```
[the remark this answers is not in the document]
RafalKorepta: It is exactly JSON schema that enforces me to re-generate any tag. That's why this function is created.

charts/operator/deployment.go
RafalKorepta: THX removed.

charts/operator/helpers.go
RafalKorepta: It was not reliable, that's why I copied from redpanda chart and filled all other fields.

charts/operator/serviceaccount.go
RafalKorepta: Done

charts/operator/chart_test.go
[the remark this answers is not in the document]
chrisseto: OH. Not entirely sure how I missed that. Rather than copying this loop it would be much more clear to just wrap it in something like:
```go
for _, scope := range []OperatorScope{Namespace, Cluster} {

if scope == Namespace {
// Namespace specific code here.
}
}
```

chrisseto (approved):

charts/operator/chart_test.go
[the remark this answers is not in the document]
RafalKorepta: It should handle that, but for me it was to much false positive. I didn't want to wait.

charts/operator/chart_test.go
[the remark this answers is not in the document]
RafalKorepta: Right I will change that.

PR #1510, opened 2024-09-03
redpanda: correct `truststore_file` usage

chrisseto:
Prior to this commit, RPK's admin and kafka client configuration would incorrect inherit the `truststore_file` setting of the internal listeners. This was mistakenly done when truststore support was first implemented due to unfamiliar naming conventions and blind grepping for instances of `truststore_file`.

Utilizing truststores, a CA for client certificate verification, would result in both jobs failing as said truststores were (correctly) not mounted to the jobs but (incorrectly) referenced in RPK's client configuration.

This commit updates the naming to `ca_file` [1] and correctly maps the value to the `ca.crt` of the server's certificate OR the server's certificate if no `ca.crt` is present.

[1] https://github.com/redpanda-data/redpanda/blob/817450a480f4f2cadf66de1adc301cfaf6ccde46/src/go/rpk/pkg/config/redpanda_yaml.go#L124-L125


chrisseto:
Looks like readiness probes are failing in CI. Strange that didn't show up in the go integration tests. 🤔

RafalKorepta (approved):

PR #1516, opened 2024-09-11
redpanda: respect `default_topic_replications` if explicitly set

### LINKED ISSUE #1501: `default_topic_replications` is not respected if provided
@clee and @ncole noticed that the chart will override `default_topic_replications` to something very similar to `.replicas` even if it's explicitly set.

Clearly we should not be doing this.

JIRA Link: [K8S-334](https://redpandadata.atlassian.net/browse/K8S-334)

chrisseto:
Prior to this commit, `default_topic_replications` could be unexpectedly modified by the chart based on some internal math. This behavior was opaque and frustrating to users attempting to finely tune their cluster configuration.

This commit removes the default value of `default_topic_replications` in `values.yaml` and instead sets it within the chart only if `default_topic_replications` is NOT provided. This preserves the existing behavior of setting `default_topic_replications` to 3 or 1 based on `statefulset.replicas`.

Fixes #1501 K8S-334

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,15 @@
 #### Fixed
 #### Removed
 
+### [5.9.3](https://github.com/redpanda-data/helm-charts/releases/tag/redpanda-5.9.3) - 2024-09-11
+#### Added
+* Add basic bootstrap user support (#1513)
+#### Changed
+#### Fixed
+* When specified, `truststore_file` is no longer propagated to client configurations.
+* If provided, `config.cluster.default_topic_replications` is now respected regardless of the value of `statefulset.replicas`.
+#### Removed
+
 ### [5.9.1](https://github.com/redpanda-data/helm-charts/releases/tag/redpanda-5.9.1) - 2024-8-19
 #### Added
 #### Changed
--- a/charts/redpanda/Chart.yaml
+++ b/charts/redpanda/Chart.yaml
@@ -23,7 +23,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 5.9.2
+version: 5.9.3
 
 # The app version is the default version of Redpanda to install.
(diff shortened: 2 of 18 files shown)


RafalKorepta:
Due to 
https://github.com/redpanda-data/helm-charts/blob/182bd51aeaaf2c53a86d9b5fd39a92be48ea5689/charts/redpanda/templates/tests/test-kafka-nodelete.yaml#L74
test case which is a copy of [produce-consume](https://github.com/redpanda-data/helm-charts/blob/b5dc3eff4cfc1be500cdde1da7bff47fc57be41a/charts/redpanda/templates/tests/test-kafka-produce-consume.yaml) that waits for replication factor being equal to 3 introduced in [PR that added calculation](https://github.com/redpanda-data/helm-charts/commit/f17b328db0497bf40f9e317aa8f2a0e20578ebd5#diff-40911a47c390205693aeafd3b14433a80885b43371729fd5461e0911d7f1374bR74).

To solve all CI failures in this PR the `.bootstrap.yaml` needs to have `default_topic_replications`. Should we explicitly set it in `values.yaml` or implicitly inject in the `.bootstrap.yaml` (in config map)?

chrisseto:
I'm just going to remove that test assertion as it doesn't seem valid anymore.

chrisseto:
Oh neat, there was a test that asserted that `default_topic_replications` was ignored. Should be fixed this time around.

RafalKorepta (approved):

chrisseto:
For posterity: I've ignored ~3 (1 but x3 with a build matrix) test failures from chart-testing due to chart testing, for some reason, using the previous install's `ci` folder which makes it impossible to fix within this PR.

The test in question was showcasing that `default_topic_replications` was being overwritten which is no longer the case with this PR: https://github.com/redpanda-data/helm-charts/pull/1516/files#diff-ca900734c6893a676c3be206969c7985aa6609c43b024085f701a04bb1648293L33

PR #1519, opened 2024-09-12
redpanda: remove default cluster configuration values

chrisseto:
This commit removes all cluster configuration values that are explicitly set in `values.yaml` to the default that redpanda would set in favor of linking to redpanda's docs.

Doing so will ease many of the checks being performed within the chart as the existence of a key in `config.cluster`, `config.tunable`, or `storage.tiered.config` are now enough to indicate that the value should be respected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@
 #### Changed
 #### Fixed
 #### Removed
+* All zero, empty, or default cluster configurations have been removed from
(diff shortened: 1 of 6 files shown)


RafalKorepta (approved):
LGTM

chrisseto:
> @chrisseto Is this considered a breaking change?

I don't think it is. There's been no change to the API and all the removed values were either ignored or are already set/implied by redpanda. Some files will have changed but there shouldn't be an observable difference AFAIK.

PR #1524, opened 2024-09-13
redpanda: don't set cluster config values in redpanda.yaml

### LINKED ISSUE #1345: Cluster level properties incorrectly being set at broker level
### What happened?

Setting (e.g.) `topic_memory_per_partition` results in this being added to the broker level `redpanda.yaml`

This is a cluster level property, so should be set with `rpk cluster config set` or `.bootstrap.yaml`

-

https://redpandadata.slack.com/archives/C01H6JRQX1S/p1717166492204159

_No response_

### Chart Version(s)

-

### Cloud provider

-

chrisseto:
Prior to this commit, cluster config values were being set in redpanda.yaml. This was being done as a historical artifact of redpanda <= 22.1 [1].

While doing so is largely benign it clutters the, already quite busy, redpanda.yaml and can result in unnecessary restarts when only changing cluster configurations.

This commit:
- Removes cluster configurations from redpanda.yaml
- Updates the config hash on the StatefulSet to only be computed from `redpanda.yaml`. This allows users to update cluster configurations from the chart without initiating a rolling restart.

Fixes #1345

[1]: https://www.redpanda.com/blog/raft-centralized-cluster-configuration-improvements


RafalKorepta:
The error in test case 99
```
helm install redpanda-fi5sxqqswg charts/redpanda --namespace redpanda-fi5sxqqswg --wait --values charts/redpanda/ci/99-none-existent-config-options-with-empty-values.yaml --timeout 600s --set=image.tag=
```
comes from s3 accessibility in new version of Redpanda. 
```
2024-09-13T21:15:50.5543062Z ERROR 2024-09-13 21:13:10,462 [shard 0:main] s3 - s3_client.cc:618 - Couldn't reach S3 storage with either path style or virtual_host style requests.
2024-09-13T21:15:50.5544601Z ERROR 2024-09-13 21:13:10,462 [shard 0:main] client_pool - client_pool.cc:79 - Self configuration of the cloud storage client failed. This indicates a misconfiguration of Redpanda. Aborting start-up ...
2024-09-13T21:15:50.5545826Z INFO  2024-09-13 21:13:10,462 [shard 0:main] main - application.cc:466 - Stopping...
```
https://github.com/redpanda-data/helm-charts/actions/runs/10855662859/job/30128727897?pr=1524#step:16:38828

BenPope (approved):
LGTM.

BenPope:
In values.go I see `TieredStorageConfig.CloudStorageCacheDirectory`, but wasn't sure where it's used. It's node config, along with `cloud_storage_inventory_hash_path_directory`. Most of the rest of TieredStorageConfig is cluster config, from what I can see.

chrisseto:
> I would drop tests with tiered storage. If we need that validation we could resurrect buildkite tests. Maybe not with chart testing.

I think a test case with minio might be the best bet as that'll have fewer moving parts and should provide the same degree of coverage. WDYT?

chrisseto:
> In values.go I see `TieredStorageConfig.CloudStorageCacheDirectory`, but wasn't sure where it's used. It's node config, along with `cloud_storage_inventory_hash_path_directory`. Most of the rest of TieredStorageConfig is cluster config, from what I can see.

Ah, that's interesting. A quick search didn't show anything in the docs. Are those settings publicly documented anywhere? The `tiered.config` struct is certainly going to be the most attractive to end users to set those settings given that it follows the naming of `cloud_storage_*`. It might be worth explicitly rejecting those from `tiered.config` via JSON schema.

RafalKorepta:
> I think a test case with minio might be the best bet as that'll have fewer moving parts and should provide the same degree of coverage. WDYT?

That would be great. But I would think it is additional work that might take you some time.

chrisseto:
> That would be great. But I would think it is additional work that might take you some time.

I was going to file an issue to do that later on and remove the current test 😅

RafalKorepta (approved):
LGTM

PR #1527, opened 2024-09-16
Add RBAC rules for operator so it can manage users

andrewstucki:
This adds the RBAC rules necessary for the operator to manage the new User resource we're adding. In order for this to pass tests I had to skip the `TestChartDifferences` test since that ensures that the same manifests render between operator releases, which is not the case here where we're adding some stuff to the operator manifests.


charts/operator/chart_test.go
```diff
@@ -234,6 +234,10 @@ const preTranspilerChartVersion = "0.4.28"
 // TestChartDifferences can be removed if in the next operator chart version values definition changes or any resource.
 // That test only validates clean transition to gotohelm definition of the operator helm chart.
 func TestChartDifferences(t *testing.T) {
```
chrisseto, line 236: I think we're safe to remove this test now as we've finished the transition to gotohelm.

Think it would be worth it to add a test that runs Kustomize against the `appVersion` of the operator in the chart and checks for divergences of RBAC?

chrisseto (approved):
Could you update the CHANGELOG before merging?

charts/operator/chart_test.go
RafalKorepta: Agree. It was just for me to validate I'm not breaking chart all over the places.

andrewstucki: 👍 another reason I'd _love_ to merge some of this into the same codebase 😅

chrisseto: Come to think of it, the operator chart might be a great first attempt at having the helm charts just follow another repository given that all the sources of truth at in the operator repo.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### [Unreleased](https://github.com/redpanda-data/helm-charts/releases/tag/redpanda-FILLMEIN) - YYYY-MM-DD
 #### Added
+* Add RBAC rules for the operator chart so it can manage users
```
RafalKorepta, line 7: I approved, but this should be moved to operator chart section. In the https://github.com/redpanda-data/helm-charts/pull/1530 I'm fixing that

PR #1532, opened 2024-09-17
Add KubeConfig option for genereting all resources

RafalKorepta:
With that change new reconciler can use Dot and Template function to generate all manifests.

Reference:

https://github.com/redpanda-data/redpanda-operator/pull/223#discussion_r1752665726


andrewstucki (approved):
I'm fine with this, in general we need to set the kubeconfig field directly anyway, so this makes it a little less ad-hoc to do so.

PR #1533, opened 2024-09-18
Add console chart templeting

RafalKorepta:
For operator reconciliation the Kubernetes resources generation will be
available as go function.

In this commit the Unwrap will now decode corev1.Volume correctly to not lose
inlined VolumeSource.

The helmette.Empty will recognise empty string correctly.

Stucts used to convert TLS configuration for Console are now copied and json
tags are used insdead of yaml, to correctly marshal and unmarshal Console
configuration.

Fix typo in the `ConsoleTLS` function name.

Adjust test case which asserts equality between helm template result and
go template function.

Allow genpartial load generated files as Redpanda values references
Console PartialValues.


charts/console/deployment.go
```diff
@@ -62,6 +62,9 @@ func Deployment(dot *helmette.Dot) *appsv1.Deployment {
 	if values.InitContainers.ExtraInitContainers != nil {
 		initContainers = helmette.UnmarshalYamlArray[corev1.Container](helmette.Tpl(*values.InitContainers.ExtraInitContainers, dot))
 	}
+	if initContainers == nil {
```
chrisseto, line 65: Why has this been added? Shouldn't we be seeing updates to console's test cases?

charts/redpanda/chart.go
```diff
@@ -109,7 +114,195 @@ func Template(release helmette.Release, values PartialValues, kubeConfig kube.Co
 		j++
 	}
 
-	return manifests[:j], nil
+	manifests = manifests[:j]
+
+	v := helmette.UnmarshalInto[Values](dot.Values)
```
chrisseto, line 119: This should be extracted into it's own function. I think we'd benefit greatly from making this code run through gotohelm so we can finally put the `console/configmap-and-deployment.yaml` to rest. We'll need to do the same for `connectors` any ways and I have no desire to maintain both this block and the helm file.

A hacky but possibly quick solution could be updating gotohelm to accept a list of packages as "subcharts" which don't get transpiled but function references get transpiled as if they were. ie:

```go
console.Deployment(...)
```

Would become (with the correct calling convention of course):

```
{{ "console.Deployment" ... }}
```

and we'd assume that `console.Deployment` is available due to subcharting.

charts/redpanda/chart_test.go
```diff
@@ -792,12 +803,19 @@ func TestGoHelmEquivalence(t *testing.T) {
 	helmObjs[stsIdx].(*appsv1.StatefulSet).Spec.Template.Spec.Containers[0].Resources, err = valuesutil.UnmarshalInto[corev1.ResourceRequirements](helmObjs[stsIdx].(*appsv1.StatefulSet).Spec.Template.Spec.Containers[0].Resources)
 	require.NoError(t, err)
 
+	// Drop annotations that gives different result as ConfigMap translated from go and helm template
+	// gives different result. The ConfigMap is first translated to yaml then calculated sha256sum.
+	goObjs[depIdx].(*appsv1.Deployment).Spec.Template.Annotations["checksum/config"] = ""
```
chrisseto, line 808: We didn't need to do this anywhere else. Why do we need to do it here? Is this a divergence between the gocode and helm code?

charts/redpanda/values.go
```diff
@@ -59,7 +59,7 @@ type Values struct {
 	AuditLogging     AuditLogging                  `json:"auditLogging"`
 	Enterprise       Enterprise                    `json:"enterprise"`
 	RackAwareness    RackAwareness                 `json:"rackAwareness"`
-	Console          Console                       `json:"console"`
+	Console          console.PartialValues         `json:"console"`
```
chrisseto, line 62: Isn't this technically `console.Values` as helm injections console's `values.yaml` here?

charts/redpanda/values.go
```diff
@@ -1497,8 +1498,16 @@ func (l *KafkaListeners) TrustStores(tls *TLS) []*TrustStore {
 	return tss
 }
 
-func (k *KafkaListeners) ConsolemTLS(tls *TLS) config.KafkaTLS {
```
chrisseto, line 1500: I never changed this because I thought it was "Console mTLS" 😬

charts/redpanda/values.go
```diff
@@ -1165,8 +1158,16 @@ type AdminListeners struct {
...
-	t := config.RedpandaAdminAPITLS{Enabled: l.TLS.IsEnabled(tls)}
+type RedpandaAdminAPITLS struct {
+	Enabled               bool   `json:"enabled"`
+	CaFilepath            string `json:"caFilepath"`
+	CertFilepath          string `json:"certFilepath"`
+	KeyFilepath           string `json:"keyFilepath"`
+	InsecureSkipTLSVerify bool   `json:"insecureSkipTlsVerify"`
+}
+
+func (l *AdminListeners) ConsoleTLS(tls *TLS) RedpandaAdminAPITLS {
```
chrisseto, line 1169: Please add comments here describing what is being returned.

cmd/genpartial/main.go
```diff
@@ -50,9 +50,8 @@ func main() {
 	}
 
 	pkgs := Must(packages.Load(&packages.Config{
-		Dir:        cwd,
-		Mode:       mode,
-		BuildFlags: []string{"-tags=generate"},
```
chrisseto, line 55: IIRC not excluding generated files really mucks with the build, especially if the generator produces invalid output. Or maybe it was that the original version would accidentally generate partial versions of the partials 🤔

pkg/gotohelm/helmette/shims.go
```diff
@@ -114,6 +115,8 @@ func Unwrap[T any](from Values) T {
 			switch reflect.New(to).Interface().(type) {
 			case *resource.Quantity:
 				return valuesutil.UnmarshalInto[*resource.Quantity](val)
+			case *corev1.Volume:
```
chrisseto, line 118: Oh dang, I bet we've got some other types that will bite us. Can you add a regression test for this?

charts/redpanda/values.go
RafalKorepta: It can be mutual TLS if `TLS.RequireClientAuth` is set to `true`

PR #1534, opened 2024-09-18
redpanda: replace bash jobs with go utilities

chrisseto:
This commit removes the bash scripts that used to power the post_upgrade and post_install_upgrade job with the recently implement `sync-cluster-config` sub command of the redpanda operator.

`bootstrap.yaml` will now be correctly populated with secret values from cluster inception. bootstrap will contain place holder values in the form of `$REDPANDA_PROPERTY_NAME` which are filled out by the envsubst init container.

The post_upgrade job has been removed as the post_install_upgrade job performed the majority of it's behavior sans setting cloud storage secrets which it now does thanks to the aforementioned bootstrap yaml envsubst'ing.


chrisseto:
I'm leaving this in draft for the moment as we'll need to perform a release of the operator to pick up `sync-cluster-config`. I also need to merge a PR that configures logging before said release as sync-cluster-config doesn't currently output any logs 😓

RafalKorepta (approved):
It looks good. There is new operator release https://github.com/redpanda-data/redpanda-operator/releases/tag/v2.2.3-24.2.4

mattfysh:
this is causing startup to hang: https://github.com/redpanda-data/helm-charts/issues/1564

PR #1546, opened 2024-10-02
Make gotohelm capable for generating subcharts

RafalKorepta:
#### https://github.com/redpanda-data/helm-charts/pull/1546/commits/a2a9c655c63571ace2e274b8831e407a21896ce0

To leverage gotothelm capability to transpile multiple package the Taskfile will change to call not only Redpanda helm chart, but it's dependencies too console and connectors.

All templated files will have package prefix in the file name. The `+gotohelm:namespace` single package comment could be used to overwrite the go package name.

This commit does not have code to use transpiled imports. That's why console and connectors translation template files are still in use.

For some reason operator chart had old auto generated files. This commit removes them.

#### https://github.com/redpanda-data/helm-charts/pull/1546/commits/6f5800ba3d12f0e55498ee9424ec12b4b059670b

Add mutability test case in transpiler unit test.


pkg/gotohelm/rewrite.go
```diff
@@ -67,26 +67,30 @@ func LoadPackages(cfg *packages.Config, patterns ...string) ([]*packages.Package
 			return nil, errors.Wrapf(errors.Join(errs...), "package %s", pkg.Name)
 		}
 
-		for _, parsed := range pkg.Syntax {
-			filename := pkg.Fset.File(parsed.Pos()).Name()
+		for _, fd := range pkgWithFilesTuple(pkg, deps) {
```
chrisseto, line 70: I don't understand the need for this modification. Does calling `LoadPackages(cfg, append(mainpkg, deps...)...)` not work here?

cmd/gotohelm/main.go
```diff
@@ -5,27 +5,42 @@ import (
...
-
+	flag.Var(&deps, "deps", "The list, separated by comma, of dependencies which will be transpiled")
 	flag.Parse()
 
 	cwd, _ := os.Getwd()
 
 	pkgs, err := gotohelm.LoadPackages(&packages.Config{
 		Dir: cwd,
-	}, flag.Args()...)
+	}, deps, flag.Args()...)
```
chrisseto, line 37: As noted in `LoadPackages`, couldn't you instead call `LoadPackages(..., append(flag.Args(), deps...)...)` here?

cmd/gotohelm/main.go
```diff
@@ -5,27 +5,42 @@ import (
...
+
+func (d *dependencies) Set(value string) error {
+	for _, dep := range strings.Split(value, ",") {
+		*d = append(*d, strings.TrimSpace(dep))
+	}
+	return nil
+}
+
 func main() {
+	var deps dependencies
```
chrisseto, line 28: No change needed but given that we accept variadic args here and that `LoadPackages` seems like it would play nice, you could make the invocation: `gotohelm <main package> [deps...]`

The transpile phase could then be:
```go
root = pkgs[0]
deps = pkgs[1:]

chart, err := gotohelm.Transpile(pkg, deps)
```

_In theory_ that might mean you could easily do something like:

```bash
# To be read as "Transpile the redpanda chart. Anything imported by the redpanda chart should be treated like a subchart/dependency.
gotohelm ./charts/redpanda ./charts/...
```

Though that will depend on the exact resolution of `packages.Load`, so play around with it a bit.

pkg/gotohelm/transpiler.go
```diff
@@ -114,6 +126,20 @@ func (t *Transpiler) Transpile() *Chart {
 	return &chart
 }
 
+func addPackagePrefixIfNotExist(f *ast.File, name string) {
```
chrisseto, line 129: Package prefixes, as a concept, already exist. They're called "namespaces" (`Transpiler.namespaces`). Let's re-use that concept here?

pkg/gotohelm/transpiler.go
```diff
@@ -97,13 +101,21 @@ type Transpiler struct {
...
 }
 
 func (t *Transpiler) Transpile() *Chart {
 	var chart Chart
-	for _, f := range t.Files {
-		if transpiled := t.transpileFile(f); transpiled != nil {
-			chart.Files = append(chart.Files, transpiled)
+
+	for _, p := range t.pkgToProcess {
+		t.Package = p
```
chrisseto, line 112: I'm a staunch believer that mutability is the root of evil. Rather than making the transpiler's internal state change depending on how it's called, move this logic "up" into the top level `Transpile` call. If this starts to get really hairy, it might make sense to make two different concepts, a `PackageTranspiler` and a `ChartTranspiler`.

We actually, have some of this already built already because of how the `bootstrap`/`shims` package works. Let's take that functionality and extend it to be more generic.

1. Modify `Chart` to have a `SubCharts` or `Dependencies` `[]*Chart` field. Have `Transpile` automatically include `shims` in that field (it's currently added in `Transpiler.Transpile` but we don't want that to be the case anymore).
2. Modify `Transpiler` to have a `subcharted []*packages.Package` field. (There may be a more efficient data structure. Up to you!). In `transpileCallExpr` extend the block that starts with `if callee.Pkg().Path() == t.Package.PkgPath` to instead check if a call is coming from a "subcharted" package and otherwise behave the same way.
3. Modify `Transpile` to roughly do:
```go
func Transpile(root *packages.Package, deps []*packages.Package) (*Chart, error) {
// We want root to be in deps so a chart calling it's own functions vs calling subcharted functions works the same way.
deps = append(deps, root)

chart := newTranspiler(root, deps)
for _, dep := range deps {
    chart.SubCharts = append(chart.SubCharts, newTranspiler(dep, deps))
}

// Always append the shims subchart (Moved up from the Transpiler.Transpile call)
chart.Subcharts = append(charts.Subcharts, shims)
}
```

And I think that's all it'll take to get this working as desired 🎉 we know that everything works already because the `shims` module (namespacing) works.

pkg/gotohelm/rewrite.go
RafalKorepta: Screenshot 2024-10-03 at 12 34 58
When you provide multiple patterns to package.Load call, then as a result you have list of packages. 

That can work, but the shim "package" would be duplicated. Maybe this is better to exclude "sneaky" shim append call and make it a proper pkg translation call.

cmd/gotohelm/main.go
RafalKorepta: I will do that, thanks for suggestion.

cmd/gotohelm/main.go
RafalKorepta: Agree

RafalKorepta:
@chrisseto I change approach. My initial desire was to load one package and deduce all information from it. When I failed, then the next thing that comes to my mind was to introduce --deps flag. With your thoughtful comment about transpiler mutability I changed whole approach and gotohelm is now orchestrator and transpiler, if instructed allows to call different packages if the namespace is filled correctly.

pkg/gotohelm/transpiler.go
```diff
@@ -129,6 +158,12 @@ func (t *Transpiler) transpileFile(f *ast.File) *File {
 		name = fileDirectives["filename"]
 	}
 
+	// Every file name is prepended with namespace
```
chrisseto, line 161: Why has this change been added? If this is to handle name conflicts when we get to transpiling subcharts, could we instead just make subdirectories for dependent charts? IMO this makes it a lot more difficult to sort through the generated files.

The following seems like a much more ergonomic layout to me. Is there a reason we can't do that?
```
templates/
   deployment.yaml
   _deployment.tpl
   console/
       deployment.yaml
       _deployment.tpl
```

pkg/gotohelm/transpiler.go
```diff
@@ -984,7 +1019,7 @@ func (t *Transpiler) transpileCallExpr(n *ast.CallExpr) Node {
 	// Call to function within the same package. A-Okay. It's transpiled. NB:
 	// This is intentionally after the builtins check to support our bootstrap
 	// package's builtin bindings.
-	if callee.Pkg().Path() == t.Package.PkgPath {
+	if t.namespaceFor(callee.Pkg()) != "" {
```
chrisseto, line 1022: Let's not combine the meaning of namespaces with anything else. This is going to be very difficult to trackdown in the future if we need to make changes as namespace changing doesn't have anything to do with subcharting or dependencies.

Instead, let's introduce a new field on `transpiler` to hold the set of all packages which may be called in this way. `dependencies` could be a good name?

cmd/gotohelm/main.go
```diff
@@ -24,13 +23,21 @@ func main() {
 		panic(err)
 	}
 
+	pkgs = append(pkgs, gotohelm.ShimsPkg)
```
chrisseto, line 26: This logic should like within the `Transpile` function. Why was it moved out? Seems like it would have worked without issue as it was 🤔

chrisseto (commented):
I like that we're not doing any mutations any more but what did you think about the rest of this [comment](https://github.com/redpanda-data/helm-charts/pull/1546#discussion_r1784678707)?

pkg/gotohelm/transpiler.go
RafalKorepta: agree I can change that

pkg/gotohelm/transpiler.go
RafalKorepta: When we have idependent packages loaded and our cli args only group do not provide dependencies flag then there is no particular graph direction (one depends on another package) When transpilation starts from root it is fine, but the gotohelm cli will go to another package and it will have the same set of packages included.

cmd/gotohelm/main.go
RafalKorepta: It can but the file will be generated per package. In Redpanda chart case you will find shim in 3 folders. I believe the definition of the function/template can not duplicate within one helm chart. That's why I treat bootstrap packae as a separated entitiy

RafalKorepta:
Agree with @chrisseto this approach is more complicated as Release and Values for subchart would need to be generated and we don't want to do that right now.

PR #1549, opened 2024-10-04
gotohelm: subcharting PoC

chrisseto:
This is not complete by any means. Just wanted to see where blockers where, if any. Was delighted to discover that `.Subcharts` did 90% of the hard work for us and very few changes to gotohelm would be required.


chrisseto:
closing in favor of https://github.com/redpanda-data/helm-charts/pull/1550

PR #1550, opened 2024-10-07
gotohelm subcharting support

chrisseto:
#### 606a8ee0cf6e0dda1242b7f8111dfe91d496d892 gotohelm: add support for subcharting

Prior to this commit, it was not possible to utilize helm's subcharting feature
within a gotohelm managed helm chart.

To add support this commit has modified gotohelm to accept another top level
input "dependencies", a list of go packages that may be referenced within a chart.

When encountered, functions belonging to packages in dependencies will be
transpiled as if they are present in the package being transpiled. In doing so,
gotohelm acts as a naive linker and assumes that the called function is
provided via some external mechanism.

The `gotohelm` binary has also had it's signature updated to:
`gotohelm <target package> [deps...]`

#### 9f7a70ceb4e16fa7b357922ae57d78064ceeb3a7 gotohelm: implement go `helm template` analog

Prior to this commit chart packages would need to implement their own analog
function for `helm template/install/upgrade`. This caused strange divergences,
especially in the case of subcharting, as there wasn't a single "entry-point"
that could be correctly used for the chart.

For example, charts would need to implement helpers to construct a
helmette.Dot, including the merging of values with their defaults. Such logic
can't be transpiled but should be the same for every chart.

This commit adds a `GoChart` construct to `gotohelm` which encapsulates the
various utilities required to execute a chart in go the same way it would be
executed within helm. `GoChart`.

Doing so allows for chart packages to implement and expose a single entrypoint
function with the signature `func(*helmette.Dot) []kube.Object`. This
entrypoint may then be rendered in helm via `_shim.render-manifest` and then in
go using `gotohelm.Load`.

#### a81dd17937824544b4a75aef38a0aea57458ab6a redpanda: utilize `gotohelm.GoChart`

This commit replaces the redpanda chart's "home brewed" `Template` function
with gotohelm's new `GoChart` functionality. It does not currently support
executing the console or connectors subcharts.

On the helm side, all the various `.yaml` files have been removed in favor of a
single entry-point.yaml that calls the `redpanda.render` function. An
unfortunate side effect of this is that the ordering of `helm template` has
been altered in a way that's impossible to emulate[1] resulting in a large diff
churn to the redpanda golden files[2].

[^1]: Helm sorts outputs based on their kind and file name. Previously the
console files and redpanda files were "interleaved" due to their naming. Now
that all redpanda manifests are emitted from a single file (entry-point.yaml)
such interleaving is not possible.

[^2]: Due to the similarity the manifests, the diff is quite messy. The only
real change is the ordering of the manifests as output by helm.


pkg/gotohelm/transpiler.go
```diff
@@ -1039,7 +1039,8 @@ func (t *Transpiler) transpileCallExpr(n *ast.CallExpr) Node {
 	case "github.com/redpanda-data/helm-charts/pkg/gotohelm/helmette.Merge":
 		dict := DictLiteral{}
 		return &BuiltInCall{FuncName: "merge", Arguments: append([]Node{&dict}, args...)}
-	case "github.com/redpanda-data/helm-charts/pkg/gotohelm/helmette.(Values).AsMap":
+	// Add note about this change.
```
RafalKorepta, line 1042: NIT: I will leave that for later. I would like to work on top of this change.

charts/redpanda/chart.go
```diff
@@ -90,38 +77,46 @@ func Template(release helmette.Release, values PartialValues, kubeConfig kube.Co
...
-		if reflect.ValueOf(manifests[i]).IsNil() {
-			continue
-		}
-		manifests[j] = manifests[i]
-		j++
+	// NB: gotohelm doesn't currently have a way to handle casting from
+	// []Instance -> []Interface as doing so generally requires some go
+	// helpers.
+	// Instead, it's easiest (though painful to read and write) to iterate over
+	// all functions that return slices and append them one at a time.
```
RafalKorepta, line 84: I fell like `asObj` would work in gotohelm transpiler. I liked how previous neat append list was constructed.

There is only one change to the `asObj` function as nitialization of array is not supported in our transpiler.
```go
func asObj[T kube.Object](manifests []T) []kube.Object {
	out := []kube.Object{}
	for i := range manifests {
		out = append(out, manifests[i])
	}
	return out
}
```

RafalKorepta (approved):
LGTM. No blockers

charts/redpanda/chart.go
chrisseto: Oh, right! Feel free to resurrect that if you'd like.

PR #1551, opened 2024-10-08
gotohelm: support helm 3.10.x / go < 1.19

### LINKED ISSUE #1531: Charts fail on Helm 3.10 around eq/ne nil
`### What happened?

In various places, trying to use the RP helm chart with Helm 3.10 (as downloaded from Helm) fails with an error along the lines of:

(Tweaking the chart gives a different first instance of this happening, but it's the same form of error)

```
Error: template: redpanda/templates/tests/test-schemaregistry-status.yaml:26:11: executing "redpanda/templates/tests/test-schemaregistry-status.yaml" at <include "full.labels" .>: error calling include: template: redpanda/templates/_helpers.tpl:43:11: executing "full.labels" at <include "redpanda.FullLabels" (dict "a" (list .))>: error calling include: template: redpanda/templates/_helpers.go.tpl:46:8: executing "redpanda.FullLabels" at <ne $values.commonLabels (coalesce nil)>: error calling ne: uncomparable type map[string]interface {}: map[]
```

Chart templates/works without issue

```
$ helm template redpanda redpanda/redpanda --version 5.8.8

Error: template: redpanda/templates/tests/test-schemaregistry-internal-tls-status.yaml:28:11: executing "redpanda/templates/tests/test-schemaregistry-internal-tls-status.yaml" at <include "full.labels" .>: error calling include: template: redpanda/templates/_helpers.tpl:43:11: executing "full.labels" at <include "redpanda.FullLabels" (dict "a" (list .))>: error calling include: template: redpanda/templates/_helpers.go.tpl:46:8: executing "redpanda.FullLabels" at <ne $values.commonLabels (coalesce nil)>: error calling ne: uncomparable type map[string]interface {}: map[]
```

As discussed, seems to relate to the version of Go used for that Helm release: https://github.com/redpanda-data/helm-charts/issues/1454

_No response_

### Chart Version(s)

5.8.8

### Cloud provider

nil

JIRA Link: [K8S-363](https://redpandadata.atlassian.net/browse/K8S-363)

[K8S-363]: https://redpandadata.atlassian.net/browse/K8S-363?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

chrisseto:
Prior to this commit gotohelm generated charts would fail to execute on any text/template engine from go < 1.19 due to it's implementation of `eq`.

We've encountered this many times and have found that the offender is often another piece of software that's vendoring helm in some way preventing users from being able to easily upgrade.

This commit adds a workaround into the gotohelm transpile to fallback to a different form of comparison when comparing a typed value to an untyped nil.

As there's not a meaningful way to test this in gotohelm itself, a minimal test case has been added to execute the redpanda chart with helm 3.10.3.

Additionally another issue with configmap-and-deployment.yaml and helm 3.10.3 was discovered and fixed by adding `Template.{BasePath,Name}` to the faux "dot" that's generated.

Fixes #1531, K8S-363


pkg/gotohelm/transpiler.go
```diff
@@ -699,6 +723,16 @@ func (t *Transpiler) transpileExpr(n ast.Expr) Node {
...
 
+			{"_", token.EQL.String(), "untyped nil"}: go119eq("eq", "lhs"),
+			{"_", token.NEQ.String(), "untyped nil"}: go119eq("ne", "lhs"),
+			{"untyped nil", token.EQL.String(), "_"}: go119eq("eq", "rhs"),
+			{"untyped nil", token.NEQ.String(), "_"}: go119eq("ne", "rhs"),
+
+			// Support for string + string!
+			{"string", token.ADD.String(), "string"}: func(a, b Node) Node {
+				return builtin("printf", NewLiteral("%s%s"), a, b)
+			},
```
RafalKorepta, line 734: It would be good to add one test for that.

RafalKorepta (approved):
LGTM

pkg/gotohelm/transpiler.go
chrisseto: Done!

RafalKorepta (approved):

PR #1554, opened 2024-10-09
Use new Chart Render functionality in console

RafalKorepta:
The `_shims.render-manifest` will call `console.render` only once to generate all console resources.

It's prelude for redpanda helm chart and de-flux controller.


charts/console/chart.go
```diff
@@ -0,0 +1,77 @@
...
+	// objects produced by the redpanda chart.
+	Scheme = runtime.NewScheme()
+
+	//go:embed Chart.yaml
+	chartYAML []byte
+
+	//go:embed values.yaml
+	defaultValuesYAML []byte
+
+	// ChartLabel is the go version of the redpanda helm chart.
```
chrisseto, line 40:
```suggestion
	// ChartLabel is the go version of the console helm chart.
```

chrisseto (approved):
LGTM!

PR #1557, opened 2024-10-10
Calculate recursively values from all chart dependencies

RafalKorepta:
To correctly render objects from root chart along with it's dependency all values needs to be calculated. From each dependency default values could be overwrited by higher level chart or even input values. The condition need to disable not only rendering but the values calculation too.

### DIFF
--- a/pkg/gotohelm/gochart.go
+++ b/pkg/gotohelm/gochart.go
@@ -51,7 +51,7 @@ func Load(chartYAML, defaultValuesYAML []byte, render RenderFunc, dependencies .
 }
 
 // LoadValues coheres the provided values into a [helmette.Values] and merges
-// it with the default values of this chart.
+// it with the default values of this chart. Dependencies are not loaded.
 func (c *GoChart) LoadValues(values any) (helmette.Values, error) {
 	valuesYaml, err := yaml.Marshal(values)
 	if err != nil {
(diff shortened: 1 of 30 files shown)


pkg/gotohelm/gochart.go
```diff
@@ -50,6 +50,25 @@ func Load(chartYAML, defaultValuesYAML []byte, render RenderFunc, dependencies .
...
+			return nil, errors.WithStack(errors.Wrapf(err, "failed to calculate dependency value of %q dependency", dep.metadata.Name))
+		}
+
+		depDefaultValues := map[string]any{}
+		err = yaml.Unmarshal(dep.defaultValues, &depDefaultValues)
+		if err != nil {
+			return nil, errors.WithStack(errors.Wrapf(err, "failed to unmarshal %q dependency", dep.metadata.Name))
+		}
+
+		merged[dep.metadata.Name] = helmette.Merge(depDefaultValues, depsValues)
```
chrisseto, line 67: Any reason to not use `helm.MergeYAMLValues` here? That'd be more consistent with helm itself. Any divergences in `merge` and whatever helm is doing internal could bite us.

pkg/gotohelm/gochart.go
```diff
@@ -58,7 +77,17 @@ func (c *GoChart) LoadValues(values any) (helmette.Values, error) {
 		return nil, errors.WithStack(err)
 	}
 
-	merged, err := helm.MergeYAMLValues("", c.defaultValues, valuesYaml)
+	merged, err := c.calculateDepsValues()
```
chrisseto, line 80: I feel that splitting this logic across two functions makes it more difficult to follow. Why not make LoadValues itself recursive?

chrisseto (approved):

pkg/gotohelm/gochart.go
RafalKorepta: No reason. I'm used to use `Merge` function to correctly transpile. This package will not be transpiled. I will change that call.

RafalKorepta: Maybe there is only one issue that I need to marshall both input parameters

pkg/gotohelm/gochart.go
RafalKorepta: I inlined that function.

pkg/gotohelm/gochart.go
RafalKorepta: After more thoughts it's fine. I used `helm.MergeYAMLValues`

pkg/gotohelm/gochart.go
```diff
@@ -174,6 +199,11 @@ func (c *GoChart) render(dot *helmette.Dot) ([]kube.Object, error) {
 			return nil, errors.Newf("missing dependency %q", dep.Name)
 		}
 
+		subdot.Values, err = subchart.LoadValues(dot.Values[dep.Name])
```
chrisseto, line 202: Do we need to be doing this if `LoadValues` is already recursively loading and merging values?

pkg/gotohelm/gochart.go
```diff
@@ -51,14 +51,37 @@ func Load(chartYAML, defaultValuesYAML []byte, render RenderFunc, dependencies .
...
+		if err != nil {
+			return nil, errors.WithStack(errors.Wrapf(err, "failed to calculate dependency value of %q dependency", dep.metadata.Name))
+		}
+
+		depsValuesByte, err := yaml.Marshal(depsValues)
+		if err != nil {
+			return nil, errors.WithStack(err)
+		}
+
+		merged[dep.metadata.Name], err = helm.MergeYAMLValues("", dep.defaultValues, depsValuesByte)
```
chrisseto, line 73: Now that this is a single function, I see why it was difficult to follow across two functions. This is already happening in the call to `dep.LoadValues(nil)`, is it not?

If we handled dependencies after the first call to `helm.MergeYAMLValues`, we could pass down `merged.Table(dep.Name)` and avoid having to call `yaml.Marshal` two more times.

Though we'll need to take into consideration `Condition` if we're performing a recursive load within this function.

Alternatively, (I think we discussed this but I don't recall why it wouldn't work) if `LoadValues` only loaded and merged the values for the current chart, we could add: `values[dep.Name] = subdot.Values` to `Dot` and achieve the same functionality.

pkg/gotohelm/gochart.go
RafalKorepta: You are right!
I don't know why I added this call.

chrisseto (approved):
I had some comments but decided to test out changes locally instead of asking questions and your tests answered my questions for me. Nice test cases!!

Helm's logic here is.... incredibly convoluted. Thanks for trekking through it.

chrisseto:
Oh looks like your test leaves the work tree dirty:

```
        pkg/gotohelm/testdata/subchart/dependency-excluded-by-default/charts/
        pkg/gotohelm/testdata/subchart/dependency-included-by-default/charts/
        pkg/gotohelm/testdata/subchart/root/charts/
        pkg/gotohelm/testdata/subchart/values-overwrite/charts/
```

Could you update `pkg/gotohelm/testdata/.gitignore` to ignore `*/charts/`?

PR #1558, opened 2024-10-10
Add console subchart in redpanda helm chart

RafalKorepta:
This PR is based on top of https://github.com/redpanda-data/helm-charts/pull/1557

#### https://github.com/redpanda-data/helm-charts/pull/1558/commits/caedc4a279b3808cb6324df91a247f990a51b6cc

Support full Console volumes

The Console partial values are used for pragmatic reasons as genpartial does
not convert console.Values to console.PartialValues inside Redpanda Values
partial generation. In both structs the type is the same. Unit tests that
marshals Redpanda partial values into file does not work as expected. The
console part of the marshalled values have multiple empty strings or nulls.
It is caused by the fact omitempty is not always set in Console.Values. In
normal template evaluation those nulls and empty strings would be replaced by
default values from values.yaml file. To unblock operator integration
Console.PartialValues is used, until genpartail is smart enough.

#### https://github.com/redpanda-data/helm-charts/pull/1558/commits/598dfe90d9458e89ececa3e3a3f1b05068ccaf61

Add Console as subchart in Redpanda chart

#### https://github.com/redpanda-data/helm-charts/pull/1558/commits/ac9795eb6a3d5d40c72e619f7ec00a0cd1f54822

Use certification path const in all places


charts/redpanda/chart_test.go
```diff
@@ -748,7 +749,16 @@ func TestGoHelmEquivalence(t *testing.T) {
 		Enabled: ptr.To(false),
 	}
 
-	values.Console = &redpanda.PartialConsole{Enabled: ptr.To(false)}
+	values.Console = &console.PartialValues{
+		Enabled: ptr.To(true),
+		Secret: &console.PartialSecretConfig{
```
chrisseto, line 754: Can you add a note about why the secret is set here? At first I thought it was just being copied but then I realized we need it to be consistent cross both helm and go invocations.

charts/redpanda/console.tpl.go
```diff
@@ -20,10 +20,192 @@ import (
...
+	"github.com/redpanda-data/helm-charts/charts/console"
 	"github.com/redpanda-data/helm-charts/pkg/gotohelm/helmette"
+	"github.com/redpanda-data/helm-charts/pkg/kube"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/utils/ptr"
 )
 
-func ConsoleConfig(dot *helmette.Dot) any {
+// To connect Console with Redpanda the values are adjusted.
+func renderConsole(dot *helmette.Dot) []kube.Object {
```
chrisseto, line 31: Can we beef up this comment and the function name? IMO this looks like it's just rendering the console subcahrt itself but it's doing a lot more.

What do you think about this scaffolding?
```go
// consoleChartIntegration plumbs redpanda connection information into the console subchart.
// It does this by ...
// It does this because ...
func consoleChartIntegration(dot *helmette.Dot) []kube.Object {
}
```

charts/redpanda/console.tpl.go
```diff
@@ -20,10 +20,192 @@ import (
...
+	v := helmette.UnmarshalInto[Values](dot.Values)
+
+	if !ptr.Deref(v.Console.Enabled, true) {
+		return nil
+	}
+
+	consoleDot := dot.Subcharts["console"]
+
+	consoleValue := helmette.UnmarshalInto[console.Values](consoleDot.Values)
```
chrisseto, line 40: Let's add a lot of comments to this function. There were a handful previously but we can definitely do better.

Noting that the `console.*` section of the redpanda chart disables the secret/configmap in order for the redpanda chart to render it with connection information will be a great start.

charts/redpanda/console.tpl.go
```diff
@@ -20,10 +20,192 @@ import (
...
 	"github.com/redpanda-data/helm-charts/pkg/gotohelm/helmette"
+	"github.com/redpanda-data/helm-charts/pkg/kube"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/utils/ptr"
 )
 
-func ConsoleConfig(dot *helmette.Dot) any {
+// To connect Console with Redpanda the values are adjusted.
+func renderConsole(dot *helmette.Dot) []kube.Object {
+	v := helmette.UnmarshalInto[Values](dot.Values)
```
chrisseto, line 32: Could we avoid single letter variable names, especially in a function this long?

charts/redpanda/console.tpl.go
```diff
@@ -20,10 +20,192 @@ import (
...
+		consoleValue.Console.Config = ConsoleConfig(dot)
+	}
+
+	if !ptr.Deref(v.Console.Deployment.Create, false) {
+		consoleValue.Deployment.Create = true
+		extraVolumes := []corev1.Volume{}
+		extraVolumeMounts := []corev1.VolumeMount{}
+		extraEnvVars := []corev1.EnvVar{}
+		if v.Auth.IsSASLEnabled() {
+			command := append([]string{},
```
chrisseto, line 59: nit:
```suggestion
			command := []string{
```

charts/redpanda/console.tpl.go
```diff
@@ -20,10 +20,192 @@ import (
...
+		}
+
+		if v.Listeners.Kafka.TLS.IsEnabled(&v.TLS) {
+			certName := v.Listeners.Kafka.TLS.Cert
+			cert := v.TLS.Certs.MustGet(certName)
+			secretName := fmt.Sprintf("%s-%s-cert", Fullname(dot), certName)
+			if cert.SecretRef != nil {
+				secretName = cert.SecretRef.Name
+			}
+			if cert.CAEnabled {
```
chrisseto, line 94: You can either solve this here or file a ticket handle it later: truststores (Defined on the listener instead of the cert) need to be handled here as they can also be used to set the CA.

charts/redpanda/console.tpl.go
```diff
@@ -20,10 +20,192 @@ import (
...
+		}
+
+		if v.Listeners.SchemaRegistry.TLS.IsEnabled(&v.TLS) {
+			certName := v.Listeners.SchemaRegistry.TLS.Cert
+			cert := v.TLS.Certs.MustGet(certName)
+			secretName := fmt.Sprintf("%s-%s-cert", Fullname(dot), certName)
+			if cert.SecretRef != nil {
+				secretName = cert.SecretRef.Name
+			}
+			if cert.CAEnabled {
```
chrisseto, line 124: Same note as above about truststores.

charts/redpanda/console.tpl.go
```diff
@@ -20,10 +20,192 @@ import (
...
+		}
+
+		if v.Listeners.Admin.TLS.IsEnabled(&v.TLS) {
+			certName := v.Listeners.Admin.TLS.Cert
+			cert := v.TLS.Certs.MustGet(certName)
+			secretName := fmt.Sprintf("%s-%s-cert", Fullname(dot), certName)
+			if cert.SecretRef != nil {
+				secretName = cert.SecretRef.Name
+			}
+			if cert.CAEnabled {
```
chrisseto, line 154: Same note as above about truststores.

charts/redpanda/values.go
```diff
@@ -1602,8 +1611,16 @@ func (l *SchemaRegistryListeners) TrustStores(tls *TLS) []*TrustStore {
 	return tss
 }
 
-func (sr *SchemaRegistryListeners) ConsoleTLS(tls *TLS) config.SchemaTLS {
-	t := config.SchemaTLS{Enabled: sr.TLS.IsEnabled(tls)}
+type SchemaTLS struct {
```
chrisseto, line 1614: Could we just use a single struct for this? It looks like all the keys are the same. Please add comments/notes about why this struct is defined as well.

cmd/genpartial/main.go
```diff
@@ -50,9 +50,8 @@ func main() {
 	}
 
 	pkgs := Must(packages.Load(&packages.Config{
-		Dir:        cwd,
-		Mode:       mode,
-		BuildFlags: []string{"-tags=generate"},
+		Dir:  cwd,
+		Mode: mode,
```
chrisseto, line 54: Is this change intentional?

RafalKorepta: It is intentional as Redpanda.Values have now console.PartialValues. As the PartialValues struct is in the file that have generated build tag, the load package needs that.

https://github.com/redpanda-data/helm-charts/blob/0ea509830613d936b213dae497e72356fa858485/charts/console/values_partial.gen.go#L1

charts/redpanda/values.go
RafalKorepta: I will move to single struct.

cmd/genpartial/main.go
chrisseto: Should we also remove the build directive from the generation as nothing will be using that now?

charts/redpanda/console.tpl.go
RafalKorepta: Ok, I will change that

charts/redpanda/console.tpl.go
RafalKorepta: thx

charts/redpanda/console.tpl.go
RafalKorepta: I think the Volumes and VolumeMounts have now much more clear functions.

charts/redpanda/console.tpl.go
RafalKorepta: I'm open for comments on the new function description. Will resolve this comment thread.

charts/redpanda/chart_test.go
RafalKorepta: In the first implementation secret was always generated. Now as License is added it makes sense to have JWT placeholder for stable tests.

Do you need a one line comment that states "For generating stable test result"?

charts/redpanda/values.go
```diff
@@ -31,6 +31,8 @@ const (
 	// PostUpgradeContainerName is the user facing name of the post-upgrade
 	// job's container.
 	PostUpgradeContainerName = "post-upgrade"
+
+	certificationMountPoint = "/etc/tls/certs"
```
chrisseto, line 35:
```suggestion
	certificateMountPoint = "/etc/tls/certs"
```

Please add a comment as well :)

chrisseto (approved):
LGTM, just the one variable name comment.

charts/redpanda/values.go
RafalKorepta: Done

PR #1562, opened 2024-10-14
gotohelm: don't transpile in `init()`

chrisseto:
Prior to this commit gotohelm could not safely be imported in other packages/binaries (namely the operator). This was due to gotohelm's init function calling `packages.Load` which delegates to `go list` which may not exist (e.g. in distroless) or be readily executable (e.g. with read-only file systems).

This commit moves the transpilation of the bootstrap package into a lazy operation which is only invoked by the transpiler. Doing so will allow other packages/binaries to consume charts in the form of `gotohelm.GoChart` without needing to worry about the transpiler.

### DIFF
--- a/pkg/gotohelm/bootstrap.go
+++ b/pkg/gotohelm/bootstrap.go
@@ -5,19 +5,9 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"sync"
 
 	"golang.org/x/tools/go/packages"
-)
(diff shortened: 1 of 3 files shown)


chrisseto:
Ah, I also removed a handful of unused flags from `packages.Load`. It _might_ make transpilation a touch faster.

RafalKorepta (approved):

PR #1569, opened 2024-10-18
operator: fix `--configurator-tag`

chrisseto:
Prior to this commit, `--configurator-tag` would render as `%s(nil)` due checking for emptiness on the wrong value.

This commit corrects the check to be on `values.Configurator.Tag`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -202,6 +202,7 @@
 #### Changed
 * The minimum Kubernetes version has been bumped to `1.25.0`
 #### Fixed
+* `--configurator-tag` now correctly falls back to `.appVersion`
 #### Removed
 
 ### [0.4.31](https://github.com/redpanda-data/helm-charts/releases/tag/operator-0.4.31) - 2024-10-7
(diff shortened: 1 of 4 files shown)


RafalKorepta (approved):
LGTM

charts/operator/testdata/template-cases.golden.txtar
```diff
@@ -14884,7 +14884,7 @@ spec:
         - --health-probe-bind-address=:8081
         - --metrics-bind-address=127.0.0.1:8080
         - --leader-elect
-        - --configurator-tag=v2.2.4-24.2.5
+        - --configurator-tag=Q8X86Dd4
```
andrewstucki, line 14887: are these, which look like generated values going to be problematic?

RafalKorepta: This exact value should not be a problem as many values are generated and are not validated (like tag should be semver'ish).

chrisseto: Nope! They're randomly generated strings that are part of the test. In these cases, the override was being specified to such an weird string but the buggy check was checking a value that was actually empty.

If you grep the codebase you'll find that value in `template-cases-generated.txtar`:
```yaml
configurator:
  pullPolicy: IfNotPresent
  repository: R4X
  tag: Q8X86Dd4
```

andrewstucki (approved):

PR #1574, opened 2024-10-22
Add ability to mount K8S token

### LINKED ISSUE #1525: automountServiceAccountToken missing for Redpanda + Operator + Connectors

Need the Redpanda + Operator + Connectors charts to support disabling automountServiceAccountToken.
Also need additional volumes and volumeMounts for the Operator chart.

### Why is this needed?

Azure Microsoft defender reported a High Severity security finding as below

"Kubernetes clusters should disable automounting API credentials" - Disable automounting API credentials to prevent a potentially compromised Pod resource to run API commands against Kubernetes clusters. For more information, see https://aka.ms/kubepolicydoc.

So in order to mitigate the above finding need to disable automounting of service account token by setting serviceAccount.automountServiceAccountToken = false .
This is not possible with the Redpanda, Operator and Connectors charts currently (only with the console chart), so opening this issue to request support for this.

I'm assuming the Operator (and perhaps others) requires access to Kubernetes API, so I also mention that the operator currently does not allow adding additional volumes and volumeMounts to manually mount the serviceAccount token.

Note - while it is best-practice to disable, the app might require the Kubernetes credentials, so in order for this to work you must also manually mount the service account credentials, example:

```
    volumes:
      - name: kube-api-access
        projected:
          defaultMode: 420
          sources:
          - serviceAccountToken:
              expirationSeconds: 3607
              path: token
          - configMap:
              items:
              - key: ca.crt
                path: ca.crt
              name: kube-root-ca.crt
          - downwardAPI:
              items:
              - fieldRef:
                  apiVersion: v1
                  fieldPath: metadata.namespace
                path: namespace
    volumeMounts:
      - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
        name: kube-api-access
        readOnly: true
```

RafalKorepta:
Redpanda values now will support strategic merge `volumes` and `volumeMounts` into whole Statefulset PodTemplate.

In operator chart the strategic merge is improved as `volumes` will be no longer overwrite, but merge with existing volumes.

Examples is added to show how to mount K8S token secret in Connectors and Console chart.

Fixes #1525


charts/connectors/testdata/template-cases.golden.txtar
```diff
@@ -10112,8 +10112,8 @@ spec:
         terminationMessagePath: /dev/termination-log
         terminationMessagePolicy: File
         volumeMounts:
-        - mountPath: /tmp
-          name: rp-connect-tmp
```
andrewstucki, line 10116: Is this behavior correct? I'd imagine if you don't specify a volume mount in the `volumeMounts` part of the values file you'd just follow the default behavior if the volume itself still exists (in this case overridden in the values.yaml)?

It seems like if you override the volume we no longer delegate to the default mounting behavior with this change.

charts/redpanda/helpers.go
```diff
@@ -464,6 +464,27 @@ func StrategicMergePatch(overrides PodTemplate, original corev1.PodTemplateSpec)
 		overrideContainers[string(container.Name)] = container
 	}
 
+	if overrides.Spec.Volumes != nil && len(overrides.Spec.Volumes) > 0 {
```
chrisseto, line 467: At this point, it's probably worth it to move this into a sharable form. 

You might be able to get away with implementing it in the bootstrap package and then re-exporting it from `helmette`?

chrisseto (commented):
Is there a reason to not change the defaults of our charts to explicitly mount the service account token? IMO that seems like a generally good change and will likely be more pleasant for end users to work with. Not that I'm opposed to further extending this functionality.

RafalKorepta:
> Is there a reason to not change the defaults of our charts to explicitly mount the service account token?

I'm not sure if that's good default. I would argue that most users of Operator and Redpanda chart (with side-car controllers) would want to use auto mount. Those functionality require K8S token, so doing that "manually" (default in values) is not what I would like to see. 

In this PR I'm trying to show examples for:
> I'm assuming the Operator (and perhaps others) requires access to Kubernetes API, so I also mention that the operator currently does not allow adding additional volumes and volumeMounts to manually mount the serviceAccount token.
>
> Note - while it is best-practice to disable, the app might require the Kubernetes credentials, so in order for this to work you must also manually mount the service account credentials, example:

In that case you would please Azure Policy, but still mount K8S token to the pod.

charts/operator/deployment.go
```diff
@@ -230,6 +249,60 @@ func operatorPodVolumes(dot *helmette.Dot) []corev1.Volume {
 			},
 		},
 	}
+
+	if values.ServiceAccount.Create && !ptr.Deref(values.ServiceAccount.AutomountServiceAccountToken, false) {
```
chrisseto, line 253: Do you know what would happen if `AutomountServiceAccountToken` was true and we added this mount?

Not pressing, just curious.

charts/operator/deployment.go
```diff
@@ -230,6 +249,60 @@ func operatorPodVolumes(dot *helmette.Dot) []corev1.Volume {
...
+		vol = append(vol, kubeTokenAPIVolume(ServiceAccountVolumeName))
+	}
+
+	return vol
+}
+
+// It's slightly changed variance of
```
chrisseto, line 260: nit:
```suggestion
// kubeTokenAPIVolume is a slightly changed variant of
```

Could you note what's changed or mention that divergences are commented in the body for clarity?

pkg/gotohelm/transpiler.go
```diff
@@ -99,6 +99,7 @@ func transpile(pkg *packages.Package, deps ...string) (_ *Chart, err error) {
 			"sort.Strings":               "sortAlpha",
 			"strings.ToLower":            "lower",
 			"strings.ToUpper":            "upper",
+			"strings.HasPrefix":          "hasPrefix",
```
chrisseto, line 102: Could you add a quick test for this just to showcase that the parmeters are in the right order? `lower` and `upper` only take one argument but most other string functions have the order swapped from the stdlib.

charts/operator/deployment.go
```diff
@@ -27,6 +28,24 @@ import (
 	"k8s.io/utils/ptr"
 )
 
+const (
+	// Injected bound service account token expiration which triggers monitoring of its time-bound feature.
+	// Reference
+	// https://github.com/kubernetes/kubernetes/blob/ae53151cb4e6fbba8bb78a2ef0b48a7c32a0a067/pkg/serviceaccount/claims.go#L38-L39
+	WarnOnlyBoundTokenExpirationSeconds = 60*60 + 7
```
chrisseto, line 35: nit: Give this a name that's meaningful to this package. `WarnOnlyBound` doesn't really make sense here IMO.

Also no need for this to be exported :)

chrisseto (commented):
I think we need one last bit, could you also set `automountServiceAccountToken` to false on all our PodSpec's as well? I don't know why Kubernetes thought there needed to be two fields to control this but there are 🤦

charts/operator/deployment.go
RafalKorepta: I didn't test that, but looking at the source code if volumes and volumesMounts follow naming scheme, then admission logic will not auto mount. 

https://github.com/kubernetes/kubernetes/blob/66e34012255abf1bbd0956a712817dad77c69c41/plugin/pkg/admission/serviceaccount/admission.go#L416-L482

charts/operator/deployment.go
RafalKorepta: Done

pkg/gotohelm/transpiler.go
RafalKorepta: You are right. I will add commit with tests.

charts/operator/deployment.go
RafalKorepta: Changed.

chrisseto (approved):
LGTM!

andrewstucki (approved):

PR #1575, opened 2024-10-22
feat: add redpanda connect chart

ooesili:
This PR refreshes [the old chart repo](https://github.com/redpanda-data/redpanda-connect-helm-chart) and moves the chart here.


mihaitodor (approved):
Thanks @ooesili 🏅 :shipit:

flake.nix
```diff
@@ -25,6 +25,9 @@
                 docker-tag-list = pkgs.callPackage ./.github/docker-tag-list.nix { };
                 helm-3-10-3 = pkgs.callPackage ./.github/helm.nix { };
                 setup-envtest = pkgs.callPackage ./.github/setup-envtest.nix { };
+                kubernetes-helm = prev.wrapHelm prev.kubernetes-helm {
+                  plugins = [ prev.kubernetes-helmPlugins.helm-unittest ];
```
chrisseto, line 29: TIL about helm unittest. We've got a bunch of custom tooling that does roughly the same thing 😅

charts/connect/templates/tests/test-connection.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: "{{ include "redpanda-connect.fullname" . }}-test-connection"
+  labels:
+    {{- include "redpanda-connect.labels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": test
+    "kube-linter.io/ignore-all": "Linter does not need to run on this test"
```
chrisseto, line 9: Should kube-linter be running at somepoint? I don't believe we have it configured in this repo but it would be good to.

charts/connect/templates/tests/test-connection.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: "{{ include "redpanda-connect.fullname" . }}-test-connection"
+  labels:
+    {{- include "redpanda-connect.labels" . | nindent 4 }}
+  annotations:
+    "helm.sh/hook": test
```
chrisseto, line 8: Could you clone the GHA setup for the other charts? Make a `.github/workflows/test_<chart-name>.yaml` and then connect it via `.github/workflows/pull_request.yaml`.

No need to setup chart testing or any of that, running `helm test` on a kind cluster for now would be fine IMO.

chrisseto (commented):
Mostly LGTM! Just want to make sure we get CI all setup ahead of time.

Other things to consider as well:
- Will chart-releaser just work or do we need to tweak the setup?
- How will users migrate from the standalone chart repo to this one?
- Do we need to update any docs? @JakeSCahill?

charts/connect/templates/tests/test-connection.yaml
ooesili: Copied this over from the stand alone repo. I'm not sure why this annotation is here, kube-lint wasn't running in CI there either.

I ran `kube-lint` on the repo just now and there are 52 errors, I'd rather not make fixing all of those a blocker for this PR. I think it would be best to add kube-lint in a follow-up PR.

chrisseto (commented):
Last two bits:

Copy this stanza and modify it to be for connect. It triggers GHAs to run the `test_connect.yaml` file whenever there's changes to the directories that you've listed.
https://github.com/redpanda-data/helm-charts/blob/ad1edfc56b700815609b9785e6ca39729d89855c/.github/workflows/pull_requests.yaml#L118-L122

You'll need to add connect to this block, which will require the check when there are changes to the connect chart:
https://github.com/redpanda-data/helm-charts/blob/ad1edfc56b700815609b9785e6ca39729d89855c/.github/workflows/pull_requests.yaml#L135-L142

charts/connect/Chart.yaml
```diff
@@ -0,0 +1,46 @@
...
+    url: https://github.com/orgs/redpanda-data/people
+type: application
+
+# The chart version and the app version are not the same and will not track
+# together. The chart version is a semver representation of changes to this
+# chart.
+version: 3.0.0
+
+# The app version is the default version of Redpanda to install.
+# ** NOTE for maintainers: please ensure the artifacthub image annotation is updated before merging
```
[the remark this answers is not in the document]
ooesili: Copied the test, seemed easy enough

charts/connect/main_test.go
```diff
@@ -0,0 +1,16 @@
+package main
```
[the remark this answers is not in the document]
ooesili: 🫡

charts/connect/main_test.go
```diff
@@ -0,0 +1,16 @@
...
+
+import (
+	"os"
+	"os/exec"
+	"testing"
+)
+
+func TestHelmUnitTest(t *testing.T) {
+	cmd := exec.Command("helm", "unittest", ".")
+	cmd.Stdout = os.Stdout
```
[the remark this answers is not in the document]
ooesili: 🫡

ooesili:
> Last two bits:
> 
> Copy this stanza and modify it to be for connect. It triggers GHAs to run the `test_connect.yaml` file whenever there's changes to the directories that you've listed.
> 
> https://github.com/redpanda-data/helm-charts/blob/ad1edfc56b700815609b9785e6ca39729d89855c/.github/workflows/pull_requests.yaml#L118-L122
> 
> You'll need to add connect to this block, which will require the check when there are changes to the connect chart:
> 
> https://github.com/redpanda-data/helm-charts/blob/ad1edfc56b700815609b9785e6ca39729d89855c/.github/workflows/pull_requests.yaml#L135-L142

Got it, found another part of the file that I think needed similar modifications too

charts/connect/main_test.go
```diff
@@ -0,0 +1,14 @@
...
+import (
+	"log"
+	"os/exec"
+	"testing"
+)
+
+func TestHelmUnitTest(t *testing.T) {
+	out, err := exec.Command("helm", "unittest", ".").CombinedOutput()
+	if err != nil {
+		log.Fatalf("failed to run helm unittest: %s", out)
```
chrisseto, line 12: `t.Fatalf` or `require.NoErrorf(err, "%s", out)`is preferable. `log.Fatalf` calls `os.Exit(1)` under the hood which also messes with `go test`

ooesili: Whoops I copied the code from [`chart_template_test.go`](https://github.com/redpanda-data/helm-charts/blob/160936fb09dad8dac241e7d69093bc0fc47e5335/charts/redpanda/chart_template_test.go#L39-L42) but I now see that it's a `TestMain()` function which behaves differently 🫣

chrisseto (approved):

PR #1580, opened 2024-10-28
Handle extraObjects like secrets

RogerBonati:
Provide extraObjects template to handle secrets

### DIFF
--- a/charts/kminion/values.yaml
+++ b/charts/kminion/values.yaml
@@ -205,6 +205,17 @@ deployment:
   # Add additional init containers, e. g. for oauth2-proxy
   initContainers: {}
 
+# Create dynamic manifests via values:
+extraObjects: []
+  # - apiVersion: v1
+  #   kind: Secret
+  #   metadata:
+  #     name: kafka-credentials-team-xy
+  #   type Opaque
+  #   data:
+  #     username: kafka_user_name
+  #     password: kafka_user_password
+
 kminion:
   # KMinion can be configured using environment variables and/or a YAML config. The yaml contents under config will
   # end up in a YAML file which will be mounted into the kminion container.


TheMeier:
Having such a mechanism is quite common in the helm world nowadays. Especially since it avoids creating custom charts around subcharts. Maybe you could re-consider

chrisseto:
If you have any references to other charts that do this, I'll take a look and re-evaluate the decision. However, given that there are myriad ways for end users to deploy a set of manifests without modifications to an upstream chart, I don't see a particularly compelling for us to maintain such a feature.

TheMeier:
@chrisseto 

Sure here is a few examples:

- [Grafana](https://github.com/grafana/helm-charts/blob/main/charts/grafana/templates/extra-manifests.yaml)
- [Prometheus](https://github.com/prometheus-community/helm-charts/blob/5f64862e5d434ad33d358d4dae3b3256d7281e5c/charts/prometheus/templates/extra-manifests.yaml) (Many more of the prometheus-community charts have this)
- [Cert-Manager](https://github.com/cert-manager/cert-manager/blob/master/deploy/charts/cert-manager/templates/extras-objects.yaml)
- [Traefik](https://github.com/traefik/traefik-helm-chart/blob/master/traefik/templates/extra-objects.yaml)
- [Kubernetes-Autoscaler](https://github.com/kubernetes/autoscaler/blob/master/charts/cluster-autoscaler/templates/extra-manifests.yaml)
- [All of bitnami charts via their template for all charts](https://github.com/bitnami/charts/blob/main/template/CHART_NAME/templates/extra-list.yaml)

There is also an issue for helm to make this a helm-feature: https://github.com/helm/helm/issues/12653

TheMeier:
@RafalKorepta @chrisseto any chance this is re-considered?

PR #1584, opened 2024-10-28
redpanda: export `Types`

chrisseto:
This commit adds a `Types` function to the redpanda chart that returns a slice of all possible `kube.Object/runtime.Object/client.Object` implementations that could be returned from various inputs of the redpanda chart.

This function will be utilized by the redpanda operator to aid in garbage collecting objects that have been removed across chart invocations. For example, toggling console on and off will result in a Deployment that needs to be removed.

### DIFF
--- a/charts/redpanda/chart.go
+++ b/charts/redpanda/chart.go
@@ -26,6 +26,13 @@ import (
 	"github.com/redpanda-data/helm-charts/pkg/gotohelm"
 	"github.com/redpanda-data/helm-charts/pkg/gotohelm/helmette"
 	"github.com/redpanda-data/helm-charts/pkg/kube"
+	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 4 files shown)


RafalKorepta (approved):

charts/redpanda/chart.go
```diff
@@ -45,6 +52,31 @@ var (
 	Chart = gotohelm.MustLoad(chartYAML, defaultValuesYAML, render, console.Chart, connectors.Chart)
 )
 
+// Types returns a slice containing the set of all [kube.Object] types that
+// could be returned by the redpanda chart.
+func Types() []kube.Object {
```
RafalKorepta, line 57: Looking at the output of the linter you might consider tag that would exclude this function from transpilation.

https://github.com/redpanda-data/helm-charts/actions/runs/11559675098/job/32174854859#step:5:337

PR #1586, opened 2024-10-30
console: consistently set `.Namespace`

chrisseto:
Prior to this commit objects produced from the console helm chart would inconsistently set `namespace`. This isn't particularly noticeable as `kubectl` and `helm` often set namespace transparently but it can cause issues when using other tooling.

This commit sets namespace on all objects it was set on, adds a test assertion that namespace is present across all objects, and removes usages of `extraInitContainers` from tests as they generated invalid manifests.


RafalKorepta (approved):

chrisseto:
Hmmmm, there's an interesting CI failure. GoHelmEquivalence fails because the console chart isn't updated in tests which isn't possible because we need to merge the change to get a release of console....

RafalKorepta:
Could we disable console in this test, release and then re-enable that test?

PR #1588, opened 2024-10-30
Add Schema CRD RBAC for operator chart

andrewstucki:
This adds RBAC rules for allowing the new `Schema` CRD to be managed by our operator.


RafalKorepta (approved):
Probably you will need to rerun `go test -update charts/operator`

PR #1590, opened 2024-10-31
feat(connect): add migration guide from old chart

ooesili:
This add instructions for users coming from the old (and soon to be archived) [redpanda-connect-helm-chart repo](https://github.com/redpanda-data/redpanda-connect-helm-chart).


chrisseto (approved):
LGTM!

PR #1594, opened 2024-11-06
Doc 686 Update descriptions of configs in the RP Connect values file

JakeSCahill:


charts/connect/README.md.gotmpl
```diff
@@ -8,77 +8,13 @@ description: Find the default values and descriptions of settings in the Redpand
 {{ define "chart.description" -}}
 Redpanda Connect is a high performance and resilient stream processor, able to connect various sources and sinks in a range of brokering patterns and perform hydration, enrichments, transformations and filters on payloads.
 
-This Helm Chart deploys a single Redpanda Connect instance in either streams mode or standalone.
```
JakeSCahill, line 11: This is all being moved to docs to avoid long-form guides in the reference.

charts/connect/README.md.gotmpl
```diff
@@ -8,77 +8,13 @@ description: Find the default values and descriptions of settings in the Redpand
...
-```
-
-Currently the streams mode `ConfigMap` should be applied **separately from and before installation of** the helm chart; support for deploying additional `ConfigMap`'s within the chart may be implemented later.
+This Helm chart deploys a Redpanda Connect instance in either streams mode or standalone.
 
-### Global Configuration
+This page describes the contents of the chart's [`values.yaml` file](https://github.com/redpanda-data/helm-charts/blob/main/charts/connect/values.yaml). Each of the settings is listed and described on this page, along with any default values.
 
-When deploying Redpanda Connect in streams mode, you may want to configure global tracing, logging and http configuration which is shared across all of your pipelines.
+For instructions on how to install and use the chart, including how to override and customize the chart's values, refer to the [deployment documentation](https://docs.redpanda.com/redpanda-connect/get-started/helm-chart/).
```
JakeSCahill, line 15: This page will go live when the docs PR is merged (link to PR TBD)

charts/connect/values.yaml
```diff
@@ -2,34 +2,54 @@
...
+  # -- Labels to add to Redpanda Connect Pods.
   podLabels: {}
-  # annotations -- Set annotations on the Redpanda Connect Deployment
+  # -- Set Annotations on the Redpanda Connect Deployment.
   annotations: {}
-  # Default 60. ref: https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/
+  # -- Time in seconds for Pod termination grace period.
   terminationGracePeriodSeconds: 60
-  # Rollout the deployment on ConfigMap changes
+  # -- Rollout the deployment on ConfigMap changes.
```
JakeSCahill, line 18: I tried testing this but not sure how it works. Could someone confirm what it's supposed to do?

david-yu:
Adding @ooesili as a reviewer.

charts/connect/values.yaml
RafalKorepta: If user would update (`helm upgrade`) connect configuration which would only change ConfigMap the Deployment and along with it Pods would not picked up (missing iNotify events probably in Redpanda Connect). That's why config map sha256 sum annotation set in Deployment would help automate config change rollout.
```
  template:
    metadata:
      annotations:
        {{- if .Values.deployment.rolloutConfigMap }}
        checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
        {{- end}}
```
https://github.com/redpanda-data/helm-charts/blob/0eab2ffff805416423b7ebec93cb2788fbbeecb5/charts/connect/templates/deployment.yaml#L20-L25

RafalKorepta: But as I see https://github.com/redpanda-data/helm-charts/pull/1594/files#diff-2d6335db92d8ff0df28b4fea97de9c234115fdca59c11fb7a09f7c86275d0539R320 watch option should be mentioned here :) just to cross reference such 'watch' capability.

PR #1595, opened 2024-11-11
operator: synchronize `{Cluster,}Roles` with operator manifests

chrisseto:
Prior to this commit the operator chart's declared permissions had grown out of sync with those declared by the operator itself. This commit re-synchronizes the permissions, adds in regression tests to prevent such drift from happening again, and releases the updated chart.


chrisseto:
Wait on https://github.com/redpanda-data/redpanda-operator/pull/301 to be merged and for a release to be cut.

charts/operator/testdata/kustomization.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  # TODO: Move these links back to main and then to the tag in appVersion.
+  - https://raw.githubusercontent.com/redpanda-data/redpanda-operator/chris/p/correct-rbac/operator/config/rbac/leader-election-role/role.yaml
```
andrewstucki, line 5: I'm assuming this will get updated after https://github.com/redpanda-data/redpanda-operator/pull/301 merges

charts/operator/rbac.go
```diff
@@ -220,6 +225,21 @@ func ClusterRole(dot *helmette.Dot) []rbacv1.ClusterRole {
 						APIGroups: []string{"cluster.redpanda.com"},
 						Resources: []string{"topics/status"},
 					},
+					{
+						Verbs:     []string{"get", "list", "patch", "update", "watch"},
+						APIGroups: []string{"cluster.redpanda.com"},
+						Resources: []string{"users"},
```
andrewstucki, line 231: oh gosh, I just now saw the differentiation here between
```go
values.Scope == Namespace
```
and 
```go
values.Scope == Cluster
```

if you're adding this here (under `Cluster`) we'll likely want to do `schemas` as well. Then again is the `Cluster` scope intended to be for the `v1` operator based on the [bootup logic for the operator](https://github.com/redpanda-data/redpanda-operator/blob/3a34a83aa5ac5c2cdc6b77cf125575873a047980/operator/cmd/run/run.go#L233-L236)?

Either way I think that we'll want to either drop this (if it is solely for the v1 controller) or add in `schemas` as well.

andrewstucki (changes requested):

charts/operator/rbac.go
chrisseto: Fixed. To be on the safe side, I'm leaving the V2 permissions in V1 as they won't have an adverse affect if they're not installed or used AFAIK. Ideally, we'll eventually remove all the gating.

RafalKorepta:
It's already merged in https://github.com/redpanda-data/helm-charts/pull/1600

PR #1597, opened 2024-11-13
Fix TLS params for clients

andrewstucki:
Previously TLS configuration for clients using the `pkg/redpanda` subpackage (our operator CRDs when using a `clusterRef`), didn't take into account all of the various configurations for custom TLS configuration per-listener. This fixes that, ensuring that TLS is only configured when it's enabled on the listener itself (or globally and not explicitly disabled), and properly configuring the Kubernetes secrets from which to pull certificate values, whether issued through cert-manager or specified via an explicit `SecretRef`.

### DIFF
--- a/pkg/redpanda/client_test.go
+++ b/pkg/redpanda/client_test.go
@@ -3,7 +3,13 @@ package redpanda
 import (
 	"testing"
 
+	"github.com/redpanda-data/helm-charts/charts/redpanda"
+	"github.com/redpanda-data/helm-charts/pkg/gotohelm/helmette"
+	"github.com/redpanda-data/helm-charts/pkg/kube"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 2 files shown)


pkg/redpanda/client_test.go
```diff
@@ -3,7 +3,13 @@ package redpanda
 import (
 	"testing"
 
+	"github.com/redpanda-data/helm-charts/charts/redpanda"
+	"github.com/redpanda-data/helm-charts/pkg/gotohelm/helmette"
+	"github.com/redpanda-data/helm-charts/pkg/kube"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	v1 "k8s.io/api/core/v1"
```
chrisseto, line 11: nit: `corev1` (Really I just need to setup the linter here too)

chrisseto (approved):

PR #1600, opened 2024-11-15
Bump operator newest v2.3.0-24.3.1 version

RafalKorepta:
Based on https://github.com/redpanda-data/helm-charts/pull/1595

This PR adds to the Github filters `value.yaml` file as it should be validated when anything has changed. There can be argument that any change to the `values.yaml` should change `chart.yaml` too, but in this is not the case 100% of times.

The operator `chart.yaml` definition now will point to the latest operator release `v2.3.0-24.3.1`.

P.S. The release was removed as there was panic found in the configurator. It's already fixed with https://github.com/redpanda-data/redpanda-operator/pull/311. Now the release is blocked by connectors go based translation. Which will be solved in the next helm-chart PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,14 @@
 #### Fixed
 #### Removed
 
+
+### [5.9.12](https://github.com/redpanda-data/helm-charts/releases/tag/redpanda-5.9.12) - 2024-11-22
+#### Added
+#### Changed
+* Chart version to update operator side-car container tag
+#### Fixed
+#### Removed
+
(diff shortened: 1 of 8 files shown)


chrisseto:
I think we're waiting on https://github.com/redpanda-data/helm-charts/pull/1595 to bump the operator release?

chrisseto:
Oh and we need to wait for the docker tag to get fixed as well, right?

RafalKorepta:
You are right!

birdayz (approved):

PR #1602, opened 2024-11-17
feat: increase connect version and add new config parameter

dersvenhesse:
Updates the Connect chart to version 3.0.1 with the most recent Connect version and an added configuration parameter for telemetry submitting.


dersvenhesse:
Unfortunately, I cannot see why the linter is failing.

ooesili (approved):

PR #1601, opened 2024-11-15
Redpanda supports decreasing core count in v24.3

JakeSCahill:
Part of doc change: https://github.com/redpanda-data/docs/pull/850

PR #1609, opened 2024-11-22
Update descriptions of --memory and --reserve-memory

JakeSCahill:
Fixes https://github.com/redpanda-data/documentation-private/issues/2776

To have single Redpanda helm chart release the https://github.com/redpanda-data/helm-charts/pull/1601 change is appended to this PR.

### DIFF
--- a/charts/redpanda/values.yaml
+++ b/charts/redpanda/values.yaml
@@ -365,8 +365,8 @@ resources:
     # For details, see this [blog](https://redpanda.com/blog/tpc-buffers).
     # For this reason, Redpanda should only be given full cores.
     #
-    # Note: You can increase cores, but decreasing cores is not currently supported.
-    # See the [GitHub issue](https://github.com/redpanda-data/redpanda/issues/350).
+    # Note: You can increase cores, but decreasing cores is supported only from
+    # 24.3 Redpanda version.
     #
     # This setting is equivalent to `--smp`, `resources.requests.cpu`, and `resources.limits.cpu`.
     # For production, use `4` or greater.
--- a/charts/redpanda/values.yaml
+++ b/charts/redpanda/values.yaml
@@ -396,7 +396,7 @@ resources:
     # It is recommended to have at least 2Gi of memory per core for the Redpanda binary.
     # This memory is taken from the total memory given to each container.
     # The Helm chart allocates 80% of the container's memory to Redpanda, leaving the rest for
-    # the Seastar subsystem (reserveMemory) and other container processes.
+    # other container processes.
     # So at least 2.5Gi per core is recommended in order to ensure Redpanda has a full 2Gi.
     #
     # These values affect `--memory` and `--reserve-memory` flags passed to Redpanda and the memory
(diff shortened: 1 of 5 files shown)


charts/redpanda/values.yaml
```diff
@@ -421,18 +421,17 @@ resources:
...
-    # The amount of memory to allocate to a container is determined by the sum of three values:
-    # 1. Redpanda (at least 2Gi per core, ~80% of the container's total memory)
-    # 2. Seastar subsystem (200Mi * 0.2% of the container's total memory, 200Mi < x < 1Gi)
-    # 3. Other container processes (whatever small amount remains)
+    # If you are setting these values manually, follow these guidelines carefully. Incorrect settings can lead to performance degradation, instability, or even data loss. The total memory allocated to a container is determined as the sum of the following two areas:
+    #
+    #- Redpanda (including Seastar): Defined by the `--memory` parameter. Includes the memory used by the Redpanda process and the reserved memory allocated for Seastar. A minimum of 2Gi per core is required, and this value typically accounts for ~80% of the container’s total memory. For production, allocate at least 8Gi.
+    #
+    # - Operating system (OS): Defined by the `--reserve-memory` parameter. Represents the memory available for the operating system and other processes within the container.
```
chrisseto, line 434: As luck would have it, I was just poking around at these values myself. As I understand it, these values are passed directly through to seastar. Meaning that they should have the same meaning as they do in scylladb.

According to [their docs](https://opensource.docs.scylladb.com/stable/kb/memory-usage), these flags should be considered mutually exclusive:
> To limit the memory usage you can start scylla with --memory parameter. Alternatively, you can specify the amount of memory ScyllaDB should leave to the OS with --reserve-memory parameter. Keep in mind that the amount of memory left to the operating system needs to suffice external scylla modules.


See also this issue and comment from our very own Travis Downs: https://github.com/scylladb/seastar/issues/375#issuecomment-1081104600


Assuming that seastar correctly detects the memory limits imposed by the cgroups configured by Kubernetes, it seems like we should be discouraging the use of these flags when running in Kubernetes.

JakeSCahill: Added a warning to discourage their use.

RafalKorepta (approved):

JakeSCahill:
LGTM- thanks for the suggestion!

charts/redpanda/values.yaml
travisdowns: I'm not sure the description is correct. Redpanda does not really "reserve" memory per se. It only allocates memory for its heap, specified by `--memory`. The when `--memory` is used the `--reserved` value is basically just a check that the `--memory` amount is not too big of a fraction of total available memory: in that case it will produce an error and fail to start.

If you are _not_ using `--memory` then `--reserved` is much more important: it basically sets the heap size based on the total available memory, minus the amount passed as `--reserved`. Note that if you don't pass reserved, seastar is calculating a default reserved amount anyway (the greater of 1.5 GB or 7% of memory).

I think the --reserved option is most important when you do not have somethign like a helm chart to do the exact `--memory` calculation for you, in that case it's a way of letting seastar auto-detect how much memory to use. When the helm chart it mostly just gets in the way since we are already doing a calculation to set --memory based on total memory.

PR #1612, opened 2024-11-29
Fix job annotation overwrite

RafalKorepta:
#### https://github.com/redpanda-data/helm-charts/commit/aac3d49a833b7d366fc1ebb43929ef8b51d6423a

Fixed ability to overwrite annotation and labels in Job metadata

#### https://github.com/redpanda-data/helm-charts/commit/b483c4b1d1faeae752cac6b6faac87f2c36f7a2e

Remove none existent post-upgrade-job values
Reference
https://github.com/redpanda-data/helm-charts/releases/tag/redpanda-5.9.6
https://github.com/redpanda-data/helm-charts/pull/1534
https://github.com/redpanda-data/helm-charts/blob/05d3182eb9337bf77e3a00ea20efa8cf4d227545/CHANGELOG.md?plain=1#L85-L92


CHANGELOG.md
```diff
@@ -8,6 +8,14 @@
 #### Fixed
 #### Removed
 
+### [5.9.15](https://github.com/redpanda-data/helm-charts/releases/tag/redpanda-5.9.15) - 2024-11-29
+#### Added
+#### Changed
+#### Fixed
+* ability to overwrite annotation and labels in Job metadata
+#### Removed
+* none existent post-upgrade-job values of the none existent resource (removed in 5.9.6)
```
andrewstucki, line 17:
```suggestion
* non-existent post-upgrade-job values of the non-existent resource (removed in 5.9.6)
```

charts/redpanda/post_install_upgrade_job.go
```diff
@@ -97,18 +97,18 @@ func PostInstallUpgradeJob(dot *helmette.Dot) *batchv1.Job {
 			Name:      fmt.Sprintf("%s-configuration", Fullname(dot)),
 			Namespace: dot.Release.Namespace,
 			Labels: helmette.Merge(
-				FullLabels(dot),
 				helmette.Default(map[string]string{}, values.PostInstallJob.Labels),
+				FullLabels(dot),
 			),
 			Annotations: helmette.Merge(
+				helmette.Default(map[string]string{}, values.PostInstallJob.Annotations),
```
andrewstucki, line 104: I'm guessing that we have this change since we don't want others to be able to override the helm-specific annotations?

Is there a perceived use-case where we want users to be able to override these?

charts/redpanda/post_install_upgrade_job.go
```diff
@@ -97,18 +97,18 @@ func PostInstallUpgradeJob(dot *helmette.Dot) *batchv1.Job {
 			Name:      fmt.Sprintf("%s-configuration", Fullname(dot)),
 			Namespace: dot.Release.Namespace,
 			Labels: helmette.Merge(
-				FullLabels(dot),
 				helmette.Default(map[string]string{}, values.PostInstallJob.Labels),
+				FullLabels(dot),
```
andrewstucki, line 101: For this change if `PostInstallJob.Labels` is considered an **override** then shouldn't we allow for **overriding** `FullLabels(dot)`?

CHANGELOG.md
```diff
@@ -8,6 +8,14 @@
 #### Fixed
 #### Removed
 
+### [5.9.15](https://github.com/redpanda-data/helm-charts/releases/tag/redpanda-5.9.15) - 2024-11-29
+#### Added
+#### Changed
+#### Fixed
+* ability to overwrite annotation and labels in Job metadata
```
andrewstucki, line 15: It seems the current implementation isn't really for overriding, but adding additional labels/annotations that we normally wouldn't set? Basically: https://github.com/redpanda-data/helm-charts/blob/05d3182eb9337bf77e3a00ea20efa8cf4d227545/charts/redpanda/values.yaml#L555-L558

If so, then I think the implementation is fine, but this should be something like:

```suggestion
* ability to set additional annotation and labels in Job metadata
```

andrewstucki (approved):
Not sure the entire context of this, but made some comments and suggestions about wording for at least the changelog if the desire is to **add** annotations/labels v. **overriding** them.

charts/redpanda/post_install_upgrade_job.go
RafalKorepta: > I'm guessing that we have this change since we don't want others to be able to override the helm-specific annotations?

It's other way around. If someone would like to override (not recommended) with this change they will be able to change annotation. Merge from mergo works other way around. 

> Is there a perceived use-case where we want users to be able to override these?

As in the https://redpandacommunity.slack.com/archives/C01AJDUT88N/p1732833410021669 community member would like to delete Job after they succeed.

charts/redpanda/post_install_upgrade_job.go
RafalKorepta: This change does exactly that. It allows for overriding. I will add test case to assert that generated label by the `FullLabel` function will be changed if I provide in the same key different label value.

CHANGELOG.md
RafalKorepta: This change tries to change the `set additional` to `overwrite` even the default one. I will update the `values.yaml` comments.

chrisseto (approved):

PR #1613, opened 2024-12-03
fix(connect): empty lines and otel example

dotdc:
> [!NOTE]
> This PR backports the changes from https://github.com/redpanda-data/redpanda-connect-helm-chart/pull/100

Changes:
- Fixed empty lines after labels when `.Values.commonLabels` is empty (you can compare with `helm template .`)
- Fixed an error when tracer is configured with `openTelemetry`, should be `open_telemetry_collector` ([source](https://github.com/redpanda-data/connect/blob/main/docs/modules/components/pages/tracers/open_telemetry_collector.adoc#L2))

### DIFF
--- a/charts/connect/Chart.yaml
+++ b/charts/connect/Chart.yaml
@@ -24,7 +24,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 3.0.1
+version: 3.0.2
 
 # The app version is the default version of Redpanda to install.
 # ** NOTE for maintainers: please ensure the artifacthub image annotation is updated before merging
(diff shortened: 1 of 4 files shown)


ooesili (approved):

charts/connect/Chart.yaml
```diff
@@ -24,7 +24,7 @@ type: application
 # The chart version and the app version are not the same and will not track
 # together. The chart version is a semver representation of changes to this
 # chart.
-version: 3.0.1
+version: 3.0.2
```
david-yu, line 27: @ooesili Should we remove so we don't release upon merge?

ooesili: Yup, good call

PR #1616, opened 2024-12-04
Handle Service Account auto mount with Deployment Spec token auto mount

RafalKorepta:
The Deployment Spec token auto mount takes precendese over Service Account token
auto mount. Now both token auto mount properties should be set to true in order
to not mount volume.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -263,6 +263,13 @@
 #### Fixed
 #### Removed
 
+### [0.4.35](https://github.com/redpanda-data/helm-charts/releases/tag/operator-0.4.35) - 2024-12-04
+#### Added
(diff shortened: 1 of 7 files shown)


chrisseto:
A bit confused about what's being fixed and why it's being fixed this way. Could we not ignore these options and instead always include the service account token mounts / volumes?

RafalKorepta:
> A bit confused about what's being fixed and why it's being fixed this way. Could we not ignore these options and instead always include the service account token mounts / volumes?

You are right. I will just mount service account token, but only when service account is created. If user would like to disable Service Account, then PodTemplate would need to be used to mount service account token with service account name.

chrisseto (approved):

PR #1620, opened 2024-12-09
Switch redpanda-operator container image instead of configurator

RafalKorepta:
Configurator container image can be replaced with redpanda-operator as binary within redpanda-operator can achieve multiple purposes.


chrisseto:
Do we need to updating anything in the operator ahead of doing this? I'm specifically thinking that either command or args might need to be updated.

 It may also be good to update the defaults in the operator so there's no dangling references to `configurator`

RafalKorepta:
> Do we need to updating anything in the operator ahead of doing this? I'm specifically thinking that either command or args might need to be updated.

Right, https://github.com/redpanda-data/redpanda-operator/pull/345 is necessary.

PR #1621, opened 2024-12-16
remove gotohelm in favor of operator gotohelm

chrisseto:
Prior to this commit `gotohelm` was duplicated in the `redpanda-operator` repo.

This commit removes `pkg/gotohelm` and instead depends on `pkg/gotohelm` from the redpanda-operator repository. Due to the use of `replace` directives in the go.mod of `redpanda-operator/gotohelm`, it's not currently possible to `go install` it. As a temporary workaround, the `main.go` for gotohelm has been preserved in this repository.

As gotohelm now requires go >= 1.23, this commit also syncs the `flake.lock` for this repository with the redpanda-operator.


chrisseto:
Sorry for the size of this PR. The bulk of it is changes to the `gotohelm` import path and minor modifications to the charts due to updates in `gotohelm`.

I will call out that there are 2 changes I don't completely understand.
- There are changes to `values.schema.json` that imply and updated `corev1.Pod` struct.
- Similarly, there are some arcane looking changes to the operators test cases. I suspect these are also due to changes to the `corev1.Pod` struct.

chrisseto:
Ah, I missed this in the diff the first time around. `k8s.io/*` has been bumped to `v0.30.3` to match the operator. So the above changes are indeed expected.

RafalKorepta (approved):

PR #1622, opened 2024-12-16
redpanda: simplify setting resources

### LINKED ISSUE #1494: CPU requests aren't set on redpanda statefulset unless min memory is specified.
### What happened?

We expected to have CPU requests set for all pods, however we are not setting a minimum amount of memory in the helm chart. This results in a statefulset with CPU limits but no requests.

We expected the statefulset to have both requests/limits for at least CPU since it was specified in the values file.

Deploy the helm chart without a minimum memory limit specified, and observe the statefulset lacking a CPU requests configuration.

We took a peak at the code and it seems like a small bug here: https://github.com/redpanda-data/helm-charts/blob/main/charts/redpanda/statefulset.go#L762-L772 where the requests are only set if the minimum memory value is specified.

Redpanda

### Chart Version(s)

<details>

5.8.12

</details>

### Cloud provider

<details>
GCP
</details>

JIRA Link: [K8S-325](https://redpandadata.atlassian.net/browse/K8S-325)

chrisseto:
#### 1fff60a5f3a21bb77f6d705edd7ae6a9f11d002a redpanda: refactor resource flag computations

This commit has no user facing changes. It refactors the various functions that
affect and define the redpanda container's resources and the resource related
CLI flags (`--smp`, `--memory`, `--reserve-memory`, and `--overprovisioned`)
into methods on the `RedpandaResource` struct.

#### 9a266e5ebc234e0cd32b2ea4b54cf459d6c1287d redpanda: simplify setting redpanda resources

Prior to this commit it was not possible to "mix and match" the resource
requests and limits of the redpanda container. The redpanda chart also required
users to grok a new model of resources unique to the chart which made the chart
less user friendly.

This commit introduces a more user friendly and backwards compatible way of
directly controlling the requests and limits of the redpanda container. Rather
than directly exposing the `--smp`, `--memory`, `--reserve-memory`, and
`--overprovisioned` flags to the end user, this method will infer them from the
provided values with updated best practices.

Fixes #1494
K8S-325
K8S-434


charts/redpanda/README.md
```diff
@@ -621,7 +621,42 @@ Enable for features that need extra privileges. If you use the Redpanda Operator
...
+For example: `resources.requests.cpu` may be set without setting
+`resources.limits.cpu` to avoid the potential of CPU throttling.
+
+Redpanda's resource related CLI flags will then be calculated as follows:
+* `--smp floor(resources.{requests,limits}.cpu)`
+* `--memory resources.{requests,limits}.memory * 90%`
+* `--reserve-memory 0`
+* `--overprovisioned resources.{requests,limits}.cpu < 1000m`
+
+If neither a request nor a limit is for cpu or memory provided, the
```
JakeSCahill, line 647:
```suggestion
If neither a request nor a limit is provided for cpu or memory, the
```

charts/redpanda/README.md
```diff
@@ -621,7 +621,42 @@ Enable for features that need extra privileges. If you use the Redpanda Operator
...
+corresponding flag will be omitted. i.e. Setting `resources.limits` and
```
JakeSCahill, line 648:
```suggestion
corresponding flag will be omitted. As a result, setting `resources.limits` and
```

charts/redpanda/values.go
```diff
@@ -293,7 +295,25 @@ type Monitoring struct {
...
+// * "Historic" - When Limits and Requests are not set.
+//
+// Implied mode (recommended) will respect the Limits and Requests as given and
+// calculate redpanda's CLI flags based off what's provided. Should additional
+// tuning be required, flags may be set directly through their appropriate
+// values.yaml option.
+//
+// Historic mode is the default which introduced a bespoke modeling of both
+// redpanda's CLI flags and container resources. See it's substructs for
+// further explanation.
```
JakeSCahill, line 312:
```suggestion
// This calculation supports two modes:
// * Explicit mode (recommended):  Activated when `Limits` and `Requests` are set.
//   In this mode, the CLI flags are calculated directly based on the provided 
//   `Limits` and `Requests`. This mode ensures predictable resource allocation
//   and is recommended for production environments. If additional tuning is required, 
//   the CLI flags can be manually overridden using their respective values in the Helm chart.
//
// * Legacy mode (default): Used when `Limits` and `Requests` are not set.
//   In this mode, the container resources and CLI flags are calculated using
//   built-in default logic, where 80% of the container's memory is allocated to Redpanda 
//   and the rest is reserved for system overhead. Legacy mode is intended for backward 
//   compatibility and less controlled environments.
//
// Explicit mode offers better control and aligns with Kubernetes best practices.
// Legacy mode is a fallback for users who have not defined `Limits` and `Requests`.
```

JakeSCahill: 'Explicit mode' is more descriptive as it emphasizes that users explicitly set requests and limits for control.

Legacy mode implies that it's backward 
compatible while also signaling that users should switch to explicit mode.

charts/redpanda/values.go
```diff
@@ -360,14 +380,246 @@ type RedpandaResources struct {
...
+		if !ok {
+			memReq, ok = (*rr.Limits)[corev1.ResourceMemory]
+		}
+
+		// If neither requests nor limits are set, don't set --memory.
+		if !ok {
+			return nil
+		}
+
+		// Here we perform out memory reservation. Historically, the intention
```
JakeSCahill, line 531:
```suggestion
		// Here we perform our memory reservation. Historically, the intention
```

charts/redpanda/values.go
```diff
@@ -360,14 +380,246 @@ type RedpandaResources struct {
...
+		return 0
+	}
+
+	// This optional `redpanda` object allows you to specify the memory size for both the Redpanda
+	// process and the underlying reserved memory used by Seastar.
+	//
+	// The amount of memory to allocate to a container is determined by the sum of three values:
+	// 1. Redpanda (at least 2Gi per core, ~80% of the container's total memory)
+	// 2. Seastar subsystem (200Mi * 0.2% of the container's total memory, 200Mi < x < 1Gi)
+	// 3. Other container processes (whatever small amount remains)
```
JakeSCahill, line 603: Is this still correct?

charts/redpanda/values.go
```diff
@@ -360,14 +380,246 @@ type RedpandaResources struct {
...
+	memory := int64(0)
+	containerMemory := rr.containerMemory()
+
+	// This optional `redpanda` object allows you to specify the memory size for both the Redpanda
+	// process and the underlying reserved memory used by Seastar.
+	//
+	// The amount of memory to allocate to a container is determined by the sum of three values:
+	// 1. Redpanda (at least 2Gi per core, ~80% of the container's total memory)
+	// 2. Seastar subsystem (200Mi * 0.2% of the container's total memory, 200Mi < x < 1Gi)
+	// 3. Other container processes (whatever small amount remains)
```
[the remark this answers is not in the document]
chrisseto: Updated ever so slightly. The legacy mode still uses the 80% calculation so that part is left as is.

chrisseto:
So all tests are failing due to rpk barking about `--lock-memory` already being set. It _seems_ like there may be a bug floating around in RPK. I'll probably fallback to `enable_memory_locking` instead of specifying it as a flag.

charts/redpanda/README.md
```diff
@@ -621,7 +621,42 @@ Enable for features that need extra privileges. If you use the Redpanda Operator
...
+Resources may be specified by either setting `resources.cpu` and
+`resources.memory` (the default) or by setting `resources.requests` and
+`resources.limits`.
+
+For details on `resources.cpu` and `resources.memory`, see their respective
+documentation below.
+
+When `resources.limits` and `resources.requests` are set, the redpanda
+container's resources will be set to exactly the provided values. This allows
+users to granularly control limits and requests to best suite their use case.
```
travisdowns, line 637:
```suggestion
users to granularly control limits and requests to best suit their use case.
```

charts/redpanda/README.md
```diff
@@ -621,7 +621,42 @@ Enable for features that need extra privileges. If you use the Redpanda Operator
...
+For example: `resources.requests.cpu` may be set without setting
+`resources.limits.cpu` to avoid the potential of CPU throttling.
+
+Redpanda's resource related CLI flags will then be calculated as follows:
+* `--smp floor(resources.{requests,limits}.cpu)`
```
travisdowns, line 642: what does `{requests,limits}` mean here? We can only set one `--smp` flag and since we refer to two (possibly different) values (request, limit) it isn't clear to me how we set it.

travisdowns: I guess we need a special case for the case where the limit is < 1000m? Because we don't set --smp=0 then.

chrisseto: > what does {requests,limits} mean here?

Ah, sorry I didn't have a great syntax for this. It's a riff on [bash's curly brace expansion](https://www.gnu.org/software/bash/manual/html_node/Brace-Expansion.html) and meant to be "`resources.requests.cpu` or `resources.limits.cpu` if `resources.requests.cpu` is not specified". 

Any recommendations on a clear yet terse syntax? cc @JakeSCahill

Do you think `max(1, floor(resources.requests.cpu or resources.limits.cpu))` is clear enough?

> I guess we need a special case for the case where the limit is < 1000m? Because we don't set --smp=0 then.

Good call out! This is handled correctly in code but wasn't documented. I'll update to `max(1, floor(...))` once we have a good value for the `...`.

travisdowns: > and meant to be "resources.requests.cpu or resources.limits.cpu if resources.requests.cpu is not specified".

Ah, I see: but even if you expand it bash-style I think there is a "coalesce" or "first_non_null" type call missing. 

Overall I'm pretty dubious on the value of setting limits and requests differently for Redpanda, since `--smp` mostly means it can't use more than the given amount of CPU anyway, but I suppose we need to support this scenario in any case.

chrisseto: > Ah, I see: but even if you expand it bash-style I think there is a "coalesce" or "first_non_null" type call missing.

Correct. Think the proposed `or` is enough? `coalesce` would be my second pick but that's because I'm very fond of SQL so I'm not sure if it's generally understandable.

> Overall I'm pretty dubious on the value of setting limits and requests differently for Redpanda, since --smp mostly means it can't use more than the given amount of CPU anyway, but I suppose we need to support this scenario in any case.

It's certainly a bit tricky but we definitely need to support any combination of requests and limits. There's been a few asks for the ability to remove CPU limits, [this blog](https://home.robusta.dev/blog/stop-using-cpu-limits) (or the image in it) gives a pretty good explanation as to why IMO. 

We opt for `requests` first as Kubernetes' scheduler behaves similarly but, as documented, if we're getting anything wrong, the user can also manually specify the flag.

chrisseto:
Ugh. So found the roadblock here. `rpk` doesn't permit `--lock-memory` or `--overprovisioned` to be specified via `additional_start_flags` due to a bug.

We'll need to implement some argument parsing to be able to filter both values out and insert them into the correct dedicated spots before this can get merged.

charts/redpanda/README.md
travisdowns: > Correct. Think the proposed `or` is enough? `coalesce` would be my second pick but that's because I'm very fond of SQL so I'm not sure if it's generally understandable.

Both are pretty unclear to me. `or` is really leaning into the python-style "return value instead of boolean" which definitely isn't familiar to everyone.

Maybe just invent your own first_set thing and then define it before the list since you'll use it several times?

JakeSCahill: Maybe we can present it like this:

`--smp = max(1, floor(first_defined(resources.requests.cpu, resources.limits.cpu)))`

Where:

`first_defined(a, b)` means use `a` if it is defined, otherwise, use `b`.

Also, given this conversation and the new `max(1, ...)` syntax, we should update this paragraph:

> If neither a request nor a limit is provided for cpu or memory, the
corresponding flag will be omitted. As a result, setting `resources.limits`
and `resources.requests` to `{}` will result in redpanda being run without
`--smp` or `--memory`. (This is not recommended).

travisdowns:
> Ugh. So found the roadblock here. `rpk` doesn't permit `--lock-memory` or `--overprovisioned` to be specified via `additional_start_flags` due to a bug.

Yes, command line args which don't have a parameter like --foo get passed as --foo=bar right?

travisdowns:
I raised this a while ago, IIRC there was some workaround.

andrewstucki:
gross workaround/fix, but if what @travisdowns is talking about is the bug you're running into @chrisseto, seems like adding `--lock-memory` to https://github.com/redpanda-data/redpanda/blob/dev/src/go/rpk/pkg/redpanda/launcher.go#L79 might do the trick?

charts/redpanda/README.md
JakeSCahill: Oh of course. Nevermind 🙂

chrisseto:
None of the above from what I can tell. Try this:

```bash
docker run -it --entrypoint bash docker.redpanda.com/redpandadata/redpanda:v23.3.1
redpanda@bf5a0eb9a68e:/$ rpk start # Create the default /etc/redpanda/redpanda.yaml
redpanda@bf5a0eb9a68e:/$  ^-C # stop redpanda
redpanda@bf5a0eb9a68e:/$ vi /etc/redpanda/redpanda.yaml # Add additional_start_flags: [--lock-memory] to the rpk stanza

redpanda@bf5a0eb9a68e:/$ rpk start
Command "start" is deprecated, use "rpk redpanda start" instead
WARNING: This is a setup for development purposes only; in this mode your clusters may run unrealistically fast and data can be corrupted any time your computer shuts down uncleanly.
Error: configuration conflict. Flag '--lock-memory' is also present in 'rpk.additional_start_flags' in configuration file '/etc/redpanda/redpanda.yaml'. Please remove it and pass '--lock-memory' directly to `rpk start`
Usage:
  rpk start [flags]

Flags:
      --advertise-kafka-addr strings        A comma-separated list of Kafka addresses to advertise (scheme://host:port|name)
      --advertise-pandaproxy-addr strings   A comma-separated list of Pandaproxy addresses to advertise (scheme://host:port|name)
      --advertise-rpc-addr string           The advertised RPC address (host:port)
      --check                               When set to false will disable system checking before starting redpanda (default true)
  -h, --help                                Help for start
      --install-dir string                  Directory where redpanda has been installed
      --kafka-addr strings                  A comma-separated list of Kafka listener addresses to bind to (scheme://host:port|name)
      --mode string                         Mode sets well-known configuration properties for development or test environments; use --mode help for more info
      --pandaproxy-addr strings             A comma-separated list of Pandaproxy listener addresses to bind to (scheme://host:port|name)
      --rpc-addr string                     The RPC address to bind to (host:port)
      --schema-registry-addr strings        A comma-separated list of Schema Registry listener addresses to bind to (scheme://host:port|name)
  -s, --seeds strings                       A comma-separated list of seed nodes to connect to (scheme://host:port|name)
      --timeout duration                    The maximum time to wait for the checks and tune processes to complete (e.g. 300ms, 1.5s, 2h45m) (default 10s)
      --tune                                When present will enable tuning before starting redpanda
      --well-known-io string                The cloud vendor and VM type, in the format <vendor>:<vm type>:<storage type>

Global Flags:
      --config string            Redpanda or rpk config file; default search paths are ~/.config/rpk/rpk.yaml, $PWD, and /etc/redpanda/redpanda.yaml
  -X, --config-opt stringArray   Override rpk configuration settings; '-X help' for detail or '-X list' for terser detail
      --profile string           rpk profile to use
  -v, --verbose                  Enable verbose logging
```

My guess is that it's due to RPK's config typing `enable_memory_locking` as `bool` instead of `*bool`. I'll file an issue once this PR is finished up. We'll have to work around it because every running instance is affected so there's no much benefit to fixing it right now.

chrisseto:
Closing as this was merged into the operator repo and then synced back to this one.

PR #1625, opened 2024-12-17
Change tiered storage cache setting to `none`

### LINKED ISSUE #1176: Migrate default for `storage.tiered.mountType` to `none` to better match Redpanda defaults

When we can afford to make the change, we should switch the default value of `storage.tiered.mountType` to `none` from `emptyDir`.

https://github.com/redpanda-data/helm-charts/blob/12dcbd011583a337408e8e8249cd12bab4218485/charts/redpanda/values.yaml#L439-L445

### Why is this needed?

Redpanda's default behavior with Tiered Storage is to use the existing data directory for the tiered storage cache. Having a default of `emptyDir` deviates from that behavior and can result in accidentally using storage that's not optimized for our disk io needs. Since Tiered Storage uses disk to cache data from the object store, a slow disk can impact read latencies from tiered storage.

JIRA Link: [K8S-147](https://redpandadata.atlassian.net/browse/K8S-147)

[K8S-147]: https://redpandadata.atlassian.net/browse/K8S-147?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

RafalKorepta:
Set default value for tiered storage cache to `none` which will defer tiered
storage cache path to Redpanda process. Previously the `emptyDir` volume could
fill Kubernetes VM quickly and tip over such K8S Node. Then user would need to
restart Pod to clean the content of the `emptyDir` volume.

[K8S-147]

[K8S-147]: https://redpandadata.atlassian.net/browse/K8S-147

Fixes #1176

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,13 @@
 #### Fixed
 #### Removed
 
+### [5.9.17](https://github.com/redpanda-data/helm-charts/releases/tag/redpanda-5.9.17) - 2024-12-17
+#### Added
+#### Changed
+* Default for tiered storage cache to `none` which will defer tiered storage cache path to Redpanda process.
+#### Fixed
+#### Removed
+
 ### [5.9.16](https://github.com/redpanda-data/helm-charts/releases/tag/redpanda-5.9.16) - 2024-12-09
 #### Added
 #### Changed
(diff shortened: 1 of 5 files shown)


andrewstucki (approved):
I'm generally fine with this, can we double check/verify manually though what this looks like when we move from `emptyDir` to `none` in a release though? Just wanting to make sure nothing wonky happens/we aren't likely to break folks by switching this default on an existing installation.

chrisseto (approved):

PR #1626, opened 2024-12-18
Swap to emptyDir for connectors integration password

andrewstucki:
It looks like when this code was introduced swapping from the raw template to gotohelm, the mount type for the directory used in writing user passwords to was accidentally modified. There are currently no golden file tests for this so it wasn't caught. Here is the original mount type, which as you can see is `emptyDir`:

https://github.com/redpanda-data/helm-charts/blob/36064ff40ea8c1a8033a9034d720b03352251c52/charts/redpanda/templates/connectors/connectors.yaml#L71-L78


chrisseto (approved):

andrewstucki:
Just FYI, this only affects installations with SASL + connectors enabled, so I added a test case for that. Here's the diff from reverting the change after generating the golden file with the fixed test just to show it validates the change properly:

```
➜  helm-charts git:(fix-connectors-mount) ✗ go test ./charts/redpanda -short
--- FAIL: TestTemplate (7.62s)
    --- FAIL: TestTemplate/enable-connectors-sasl (0.08s)
        chart_template_test.go:144:
            	Error Trace:	/Users/andrewstucki/code/helm-charts/pkg/testutil/testutil.go:131
            	            				/Users/andrewstucki/code/helm-charts/pkg/testutil/testutil.go:209
            	            				/Users/andrewstucki/code/helm-charts/charts/redpanda/chart_template_test.go:144
            	Error:
            	            	spec.template.spec.volumes.redpanda-user-password  (apps/v1/Deployment/redpanda-connectors)
            	            	  - one map entry removed:     + one map entry added:
            	            	    emptyDir: {}                 secret:
            	            	                                   secretName: redpanda-users

            	Test:       	TestTemplate/enable-connectors-sasl
```

PR #1634, opened 2025-01-08
Create synchronization script

RafalKorepta:
The source of truth is now redpanda-operator repostiory where gotohelm,
genpartial and genschema cli are located. Those tools will generate inside
redpanda-operator repository the following charts:
* connectors
* console
* operator
* redpanda

Reference
https://github.com/redpanda-data/redpanda-operator/pull/339
https://github.com/redpanda-data/redpanda-operator/pull/354
https://github.com/redpanda-data/redpanda-operator/pull/365

---

Depends on https://github.com/redpanda-data/redpanda-operator/pull/372

[K8S-472]

[K8S-472]: https://redpandadata.atlassian.net/browse/K8S-472


Taskfile.yaml
```diff
@@ -201,3 +161,9 @@ tasks:
   kind-delete:
     cmds:
       - kind delete cluster --name {{.KIND_CLUSTERNAME}}
+
+  sync:charts:
```
chrisseto, line 165: Rather than have this target a relative directory could we perform this sync to a git SHA so there's no change of unexpected changes making their way in?

Git cloning into a subdirectory or downloading the tar.gz of the repo at the given SHA are the first options that come to mind.

RafalKorepta: Ok I will test the tar.gz route

chrisseto (approved):

PR #1651, opened 2025-03-10
Update Chart.yaml - Update description 

david-yu:
Changed from "Real-time engine" to "AI Data Platfom" for modern apps in the Description.


david-yu:
Will need to re-open on the Operator side.

PR #1658, opened 2025-04-04
sync console chart v3.0+k8s1-beta.1

chrisseto:


chrisseto:
I'm going to do some bad things in the name of expedience to make sure chart-release / GHA works as expected.

PR #1671, opened 2025-04-25
console: update console version to v3.0.1

weeco:


RafalKorepta:
We manage console chart form repdanda-operator repo. I will move your change to https://github.com/redpanda-data/redpanda-operator/pull/763

PR #1698, opened 2025-07-04
Add annotations for the Redpanda Connect Service

artemklevtsov:
Hi,

We use annotation-based scraping for the Prometheus so we need to set annotations for the service.

Regards.


mmatczuk (approved):
LGTM

PR #1709, opened 2025-09-30
Conditional 'and' for replicas

mediowen:
This allows for other tools (like KEDA) to manage HPA without GitOps tools like ArgoCD fighting scaling decisions.

While technically this changes default behaviour, `replicas` is set to 1 out of the box if omitted, so no practical change.


mmatczuk (approved):
Please add more meaningful commit message

mediowen:
Incidentally:
Not sure this is the right place to mention, but I received some marketing email from Redpanda Data this morning addressed to 'mediowen'. I assume this was either collected when I signed the CLA or has otherwise been scraped from my GitHub commit email. I double checked the CLA and couldn't readily find anything that consents to marketing communication, but perhaps I'm missing something.

Either way, this method of filling mailing lists doesn't feel great to me - if you've got any route to check up on that I'd appreciate it.

PR #1731, opened 2026-03-23
connect: bumped version to 4.84.1

dersvenhesse:
Using the current Connect version within the Chart and updated its version to 3.2.0.


david-yu:
Thanks for the contribution, I've bumped the RPCN binary here: https://github.com/redpanda-data/helm-charts/pull/1731 so will close.

